PR #4, opened 2022-12-01
feat(server): Support Galactica

OlivierDehaene:


server/text_generation/models/causal_lm.py
```diff
@@ -156,31 +156,29 @@ def concatenate(cls, batches: List["CausalLMBatch"]) -> "CausalLMBatch":
                 past_keys = past_keys.view(batch.size, -1, *past_keys.shape[-2:])
                 past_values = past_values.view(batch.size, -1, *past_values.shape[-2:])
 
-                _, num_heads, head_dim, padded_sequence_length = past_keys.shape
+                _, num_heads, padded_sequence_length, head_dim = past_values.shape
```
Narsil, line 159: IS it so that this line is more consistent ? 
I'm not sure why this is needed for galactica.

Narsil (approved):
Nice ! LGTM

server/text_generation/models/causal_lm.py
OlivierDehaene: I thought keys were always of shape `[batch_size, num_heads, head_dim, seq_length]` for all models but actually it's values that are always of shape `[batch_size, num_heads, seq_length, head_dim]`.

I'm just rectifying my mistake.

PR #5, opened 2022-12-02
fix: Some small fixes

njhill:
Thanks for sharing this really nice project!

I found some small things that may be of interest:
- Avoid theoretical hang in batcher loop
- Avoid a couple of clones in the router generate method
- Keep attention mask tensors as integers

Also a small suggestion: I think if [this](https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation/models/causal_lm.py#L361) line was changed, the `Model.num_heads` attribute would no longer be needed:
```diff
+ t.view(batch.size, -1, *t.shape[-2:])[next_batch_keep_indices]
- t.view(-1, self.num_heads, *t.shape[-2:])[next_batch_keep_indices]
```

### DIFF
--- a/router/src/batcher.rs
+++ b/router/src/batcher.rs
@@ -70,7 +70,7 @@ impl Batcher {
 
         // Notify the background task that we have a new entry in the database that needs
         // to be batched
-        self.shared.batching_task.notify_waiters();
+        self.shared.batching_task.notify_one();
 
         // Await on the response from the background task
         // We can safely unwrap as the background task will never drop the sender
@@ -104,10 +104,9 @@ async fn batching_task(
         // Get the next batch from the DB
         // This batch might be smaller than the maximum batch size if there are not enough requests
         // waiting in the DB
-        let mut waiting_tokens = 0;
         while let Some((request_ids, batch)) = db.next_batch(None, max_batch_size) {
             let mut cached_batch = wrap_future(client.generate(batch), request_ids, &db).await;
-            waiting_tokens += 1;
+            let mut waiting_tokens = 1;
 
             // We loop until we do not receive any cached batch from the inference server (== until
             // all requests have met their stopping criteria)
@@ -131,11 +130,11 @@ async fn batching_task(
                     if let Some((new_request_ids, new_batch)) =
                         db.next_batch(min_size, max_batch_size)
                     {
-                        // Reset waiting counter
-                        waiting_tokens = 0;
                         // Generate one token for this new batch to have the attention past in cache
                         let new_cached_batch =
                             wrap_future(client.generate(new_batch), new_request_ids, &db).await;
+                        // Reset waiting counter
+                        waiting_tokens = 1;
                         // Extend current batch with the new batch
                         if let Some(new_cached_batch) = new_cached_batch {
                             request_ids.extend(new_cached_batch.requests.iter().map(|req| req.id));
--- a/router/src/batcher.rs
+++ b/router/src/batcher.rs
@@ -105,7 +105,7 @@ async fn batching_task(
         // This batch might be smaller than the maximum batch size if there are not enough requests
         // waiting in the DB
         let mut waiting_tokens = 0;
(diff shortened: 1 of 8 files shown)


router/src/batcher.rs
```diff
@@ -105,7 +105,7 @@ async fn batching_task(
         // This batch might be smaller than the maximum batch size if there are not enough requests
         // waiting in the DB
         let mut waiting_tokens = 0;
-        if let Some((request_ids, batch)) = db.next_batch(None, max_batch_size) {
```
OlivierDehaene, line 108: I don't think we should change this line as it could lead to a wrong value for `waiting_tokens`.

If I understood correctly, the potential issue arises if a request comes while we are finishing a batch. The `notify_waiters` call happens while we are finishing a generation and the `batching_task` gets stuck till a new request happens and it can finally cross the `notified.await`.

Is my understanding correct? If so, what do you think of just changing [this line](https://github.com/huggingface/text-generation-inference/blob/main/router/src/batcher.rs#L73)?

```diff
- self.shared.batching_task.notify_waiters();
+ self.shared.batching_task.notify_one();
```

OlivierDehaene (commented):
Thank you for your interest in this project and for this PR!

I agree with your suggestion. We don't need `num_heads` anymore it's a relic of the past.

router/src/batcher.rs
njhill: > I don't think we should change this line as it could lead to a wrong value for `waiting_tokens`.

I think this could be resolved by moving the `moving_tokens` assignment after the `next_batch` call?

> Is my understanding correct?

Yes, that's correct.

> If so, what do you think of just changing [this line](https://github.com/huggingface/text-generation-inference/blob/main/router/src/batcher.rs#L73)?
> 
> ```diff
> - self.shared.batching_task.notify_waiters();
> + self.shared.batching_task.notify_one();
> ```

I think you're right that this is also necessary (I hadn't realized the differing semantics of `notify_waiited` and `notify_one`), but I don't think it's sufficient.

Consider a case where a burst of requests (> max batch size) arrives while a batch is in progress. The current batch all finish at the same time and so `None` is returned from `generate_with_cache` and we return to the start of the loop.

At this point it would be stuck at `notify_waiters`, but if this is changed to `notify_one` then we would still have a permit and so proceed to fill a new batch. However if this next max-sized batch also finished together without further requests arriving, the loop would now be stuck at `notify_one` despite requests remaining in the db queue.

njhill: @OlivierDehaene I pushed a couple more commits addressing the above, let me know what you think. Also noticed an incorrect error message.

OlivierDehaene: Yes it makes perfect sense! 

Then I think we should keep `notify_waiters`as is as I think it is more explicit. Our intent is to notify the batching task that we have new requests but it is its responsability to empty the db with this new looping logic.

Great catch btw! It's good to have a new pair of eyes on this code :)

OlivierDehaene (approved):
Looks great! 
Nit pick on `notify_waiters`.

OlivierDehaene (approved):

router/src/batcher.rs
njhill: @OlivierDehaene thanks, and thanks again for sharing it, I really like how clean and well-structured it is. And it got me to finally learn rust :)

router/src/batcher.rs
```diff
@@ -70,7 +70,7 @@ impl Batcher {
 
         // Notify the background task that we have a new entry in the database that needs
         // to be batched
-        self.shared.batching_task.notify_waiters();
+        self.shared.batching_task.notify_one();
```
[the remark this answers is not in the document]
njhill: @OlivierDehaene WDYT? :)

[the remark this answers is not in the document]
OlivierDehaene: You are right!
I added this change in #6.

PR #13, opened 2022-12-20
Easiest fix.

### LINKED ISSUE #12: Causal LM modifies the input when returning text
One of the issues if that Causal LM returns the entire `input + generated_text`. And `input` is actually `tokenizer.decoder(tokenizer.encode(input))`. The issue comes when that process is lossy, and causes weird behaviours such as this https://huggingface.co/bigscience/bloom/discussions/153#6397907b71eb2455d898e0a4 

We can instead either:
 - actually use the input instead of going through a potentially lossy mechanism.
 - change the API to always return "added" text. **BREAKING**

Narsil:
Fixes #12 in the easiest way I could think of.

### DIFF
--- a/server/text_generation/models/causal_lm.py
+++ b/server/text_generation/models/causal_lm.py
@@ -354,7 +354,8 @@ class CausalLM(Model):
             if stop:
                 # Decode all tokens
                 output_text = self.tokenizer.decode(
-                    all_input_ids.squeeze(-1), skip_special_tokens=True
+                    all_input_ids.squeeze(-1), skip_special_tokens=True,
+                    cleanup_tokenization_spaces=False
                 )
                 # Slice with input_length to remove padding
                 token_ids = all_input_ids[-new_input_length:]


thomasw21:
I think this just patches a current issue, typiecally the real issue is you use `tok.decode(tok.encode(input))` instead of just using input. See: https://github.com/huggingface/text-generation-inference/issues/12#issue-1504439206
It would work on BLOOM since we have lossless tokenizer.

OlivierDehaene:
I guess we can merge this as a short term solution. 

> Surely we can find a way to return input + tok.decode(generate_tokens) instead tok.decode(tok.encode(input) + generated_tokens.

This should be the next step. 

> Also the return_full_text=False suggestion you made in https://github.com/huggingface/transformers/pull/20846#issuecomment-1359707496 makes a lot more sense to me, but you'll need to change the web as well to support that.

I think it might be better to make the output schema clearer. Right now it looks like this:

```
{ 
  'generated_text': <input> + <generated> 
}
```
for CausalLMs

and 

```
{ 
  'generated_text': <generated> 
}
```
for Seq2SeqLMs

In my opinion it should be:


```
{ 
  'generated_text': <generated> 
  'prompt': <input>
}
```
or something similar.

But then it would return something different than other generative models in api-inference which is not good.

PR #15, opened 2022-12-28
Simplify some logic in db.rs, validation.rs

njhill:
I noticed some opportunity to collapse some of the logic, in case you are interested.

### DIFF
--- a/router/src/validation.rs
+++ b/router/src/validation.rs
@@ -6,7 +6,7 @@ use thiserror::Error;
 use tokenizers::tokenizer::Tokenizer;
 use tokio::sync::{mpsc, oneshot};
 
-const MAX_MAX_NEW_TOKENS: u32 = 512;
+const MAX_MAX_NEW_TOKENS: usize = 512;
 const MAX_STOP_SEQUENCES: usize = 4;
 
 /// Validation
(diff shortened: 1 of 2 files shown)


OlivierDehaene (commented):
Thanks

router/src/validation.rs
```diff
@@ -4,12 +4,11 @@ use axum::http::StatusCode;
 use axum::Json;
 use thiserror::Error;
 use tokenizers::tokenizer::Tokenizer;
-use tokenizers::{
-    DecoderWrapper, ModelWrapper, NormalizerWrapper, PostProcessorWrapper, PreTokenizerWrapper,
-    TokenizerImpl,
-};
 use tokio::sync::{mpsc, oneshot};
 
+const MAX_MAX_NEW_TOKENS: u32 = 512;
```
OlivierDehaene, line 9: Why u32?

njhill: @OlivierDehaene because it's compared with `request.parameters.max_new_tokens` which is u32. Could change it to `usize` and move where the cast is done, not sure it makes much difference?

OlivierDehaene: I was just wondering because you use `usize` in `MaxNewTokens(usize)`. I guess you could use `u32` everywhere.

njhill: @OlivierDehaene I've now changed it to `usize`, but I just realized what you mean w.r.t. changing the enum parameter to `u32`. I guess I thought it looked weird for those enum parameter types to be inconsistent but it's probably better, will update again.

OlivierDehaene (approved):
Thanks

PR #16, opened 2022-12-28
fix: Check for device type correctly when determining initial padding

njhill:
AFAIK there is no torch device type called "gpu".


OlivierDehaene (approved):
oops

PR #17, opened 2022-12-28
Improve logging from python shards

njhill:
I found these changes necessary to be able to do reasonable problem diagnosis.

- Ensure output from the shards makes its way to the launcher's stdout
- Log full exceptions raised by server rpc functions
- Fix stderr logging in `integration_tests.rs`


OlivierDehaene:
I was working on #19 before going in vacation. If that's ok with you, I will add you as a co-author on #19 and close #17.

njhill:
@OlivierDehaene of course, thanks! I added a few comments to #19.

PR #20, opened 2023-01-03
feat(router): Support return_full_text

OlivierDehaene:
@Narsil,

This is a bit annoying as return_full_text default value is not the same between CausalLMs (return_full_text=True) and Seq2SeqLMs (return_full_text=False).

What do you think should be the optimal solution here?

### DIFF
--- a/router/src/batcher.rs
+++ b/router/src/batcher.rs
@@ -188,8 +188,14 @@ fn send_generated(finished: Vec<GeneratedText>, db: &Db) {
             .remove(&output.request.unwrap().id)
             .expect("ID not found in db. This is a bug.");
 
+        let mut output_text = output.output_text;
+        if entry.request.parameters.return_full_text {
+            // Concat prompt and output text
+            output_text = entry.request.inputs + &output_text;
+        }
+
         let response = InferResponse {
-            output_text: output.output_text,
+            output_text,
             generated_tokens: output.generated_tokens,
             token_ids: output.token_ids,
             tokens: output.tokens,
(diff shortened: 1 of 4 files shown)


njhill:
My vote would be for the server to abstract this detail and have the default the same irrespective of the type of model, so that the external behaviour is equivalent.

Narsil:
> This is a bit annoying as return_full_text default value is not the same between CausalLMs (return_full_text=True) and Seq2SeqLMs (return_full_text=False).

What do you mean ? The default in `transformers` is always `True`. It just happens that for seq2seq the initial text is usually `[BOS]` so it gets discarded anyway during decoding. So `return_full_text=False` really only has an impact for decoder only. 

(To be perfectly honest the param is not implemented for `text2text-generation` since the two pipelines are actually separate in `transformers`)

njhill:
Given how this server abstracts over the two kinds of models, might it make sense to prepend the encoder input when `return_full_text` is specified as `True`? Perhaps the parameter could be named `include_input` or something to avoid confusion with the `transformers` parameter?

Narsil:
I think alignment with `transformers`  is desirable.
I think for encoder-decoder, it doesn't make sense to include the prompt beforehand. Because the prompt is encoded, and is only used as guidance for the decoding part. (So there can be 1 or more encoding, which can even be text or different modality).

The real previous text is the `decoder_start_ids`  if you want (usually `bos` but sometimes more complex like `[bos, <language_en>, <start_of_translation>]`. But enabling users to specify the start with maybe some readable tokens can be useful too.

PR #22, opened 2023-01-08
Mention that BLOOMZ was deactivated

vadi2:


OlivierDehaene:
Hey @vadi2, thanks for the MR.

However BLOOMZ isn't deactivated in the sense that it has everything available on the hub to run it whereas we never pushed the safetensors weights to make galactica available.

vadi2:
It does not seem to be possible to run bloomz via huggingface's web interface. Mind explaining what do you mean exactly?

OlivierDehaene:
Well while this tool was used for the BLOOMZ api-widget (which is indeed turned off at the moment), you can still use text-generation-inference to setup your own BLOOMZ API.

PR #24, opened 2023-01-17
fix: More miscellaneous minor fixes and simplification

njhill:
- Fix some type hints, in particular base tokenizer class
- Make use of `tensor.new_zero/empty` methods
- Simplify env var string parsing in launcher


OlivierDehaene (approved):
sure

PR #25, opened 2023-01-18
Proposal: Use bounded queue instead of database

njhill:
Entries are moved into a hashmap owned by the batching loop at time of request batching.

- Less code, less locking
- Fewer synchronization primitives - replaces mutex, arc, notifier, semaphore


OlivierDehaene:
Hello

Indeed this code requires less sync primitives. However, if I understood correctly, this is mainly because you share state via message passing instead of guarding the state with a Mutex. I don't think this comes from using a queue instead of a binary tree, right?

I don't think this has a performance benefit but you can prove me wrong on this. 

While I agree this implementation has less code, I think it is also harder to understand.

To be honest, I don't really understand what you want to achieve with this PR. Can you provide some background on where you are coming from? 
I don't see big drawbacks from the current implementation that you are solving with this refactor and the current implementation has been rock stable for us since we started using it in prod.

njhill:
Thanks @OlivierDehaene, yes this is really just a suggestion, please feel very free to reject it!

I will try to provide some more of the rationale:

At a high level what is needed is a blocking queue... this is kind of a standard MPSC event loop pattern where the incoming request threads are delivering tasks for a single-thread worker / event loop to consume. So an MPSC channel designed specifically for this seems the most natural thing to use. Once they are dequeued by the batcher no, more synchronization is needed, it only requires the ability to retrieve them to send the response when ready, so a simple non-shared hashmap works well.

IMHO this is actually bit easier to understand since the separation of concerns is clearer, but I know it's always easier to say that about code you wrote yourself!

The current locked BTree is kind of like a custom implementation of a blocking queue (and also doing the job of keeping track of in-progress requests), but the fact it uses a mutex means that it won't be as fast under heavy contention as the lock-free channel. Requests aren't only potentially contenting with each other but also the batcher task both when it's retrieving the batches and when it's sending the responses. This is exacerbated further when adding response streaming capability where all of the in-progress entries are retrieved on every batcher iteration (I know not applicable in your current impl).

All that said, it's likely the performance difference isn't noticeable unless the system is under heavy concurrent load, and perhaps not even then given that the main bottleneck of the system as a whole is the time taken by each forward pass of the model.

OlivierDehaene:
>  Requests aren't only potentially contenting with each other but also the batcher task when it's retrieving the batches

This cannot be solved even with a message queue though. At the end of the day, the messages will be ordered and your request for a "new batch" might still be behind a lot of "request add" messages. Exactly like a lock contention.

You can have priority queues however but since the bottleneck is clearly not here I don't see the point.

> This is exacerbated further when adding response streaming capability where all of the in-progress entries are retrieved on every batcher iteration.

This can be solved easily in the current implem by poping `next_batch`. It wasn't done at first because of legacy reasons that are not in the current implem.

router/src/batcher.rs
```diff
@@ -1,77 +1,77 @@
...
-            shared.clone(),
+            Queue::new(receiver),
         ));
 
-        Self { db, shared }
+        Self { sender }
+    }
+
+    /// Reserve a slot in the queue for sending a request
+    pub(crate) fn reserve_slot(&self) -> Result<RequestSender<'_>, TrySendError<()>> {
```
OlivierDehaene, line 47: You could make this private and part of infer for a clearer API.

router/src/batcher.rs
```diff
@@ -1,77 +1,77 @@
...
+        self.sender.try_reserve().map(|permit| RequestSender { permit })
     }
+}
+
+pub(crate) struct RequestSender<'a> {
```
OlivierDehaene, line 52: If reserve_slot is used inside infer, I don't think you need this struct.

router/src/queue.rs
```diff
@@ -0,0 +1,142 @@
...
+    /// Get the next batch without blocking
+    /// Corresponding entries are added to the entries map
+    pub(crate) fn try_next_batch(
+        &mut self,
+        min_size: usize,
+        max_size: usize,
+        entries: &mut IntMap<u64, Entry>,
+    ) -> Option<Batch> {
+        // Fill the buffer up to max_size, without waiting
+        while self.buffer.len() < max_size {
```
OlivierDehaene, line 76: I don't understand this part.
If the purpose of this new implem is to optimize the router, why move the enqueue logic on the hot path?

OlivierDehaene (commented):
If you are interested. 

I don't think I will merge this PR. However I think you highlighted some interesting drawbacks and I will make a few modifications to the code.

njhill:
> At the end of the day, the messages will be ordered and your request for a "new batch" might still be behind a lot of "request add" messages. Exactly like a lock contention.

I don't quite follow what you mean here, we might be talking past each other a bit. The channel is lockless, so there is virtually no contention between multiple incoming requests being enqueued at the back of the queue. The batcher is pulling entries from the _front_ of the queue - and similarly these enqueues/dequeues don't interfering with each other from a synchronization pov. The two can happen concurrently.

> You can have priority queues however but since the bottleneck is clearly not here I don't see the point.

I also don't understand what purpose/benefit a priority queue would provide here.


The BTree + mutex is essentially a custom implementation of a blocking concurrent queue.

router/src/batcher.rs
njhill: This is what I did originally. The reason for separating the reservation was so that it can happen before the request preprocessing steps (validation etc.), i.e. in place of the existing semaphore check. That way you aren't doing unnecessary stuff (tokenization etc.) for the rate-limited requests.

router/src/batcher.rs
njhill: Yes that's right, and you don't need reserve at all then, can just call `send` directly on the channel. But the reason for doing it is explained in above comment.

router/src/queue.rs
njhill: I don't understand the question - this loop is calling `try_recv()` to _dequeue_ entries from the channel. They are put into an intermediate buffer only so that we don't have to return them if there are fewer than `min_size`.

njhill: @OlivierDehaene oh I just realized what a possible source of the confusion is ... I probably should have been clearer that the MPSC channel _is_ the queue.. the VecDequeue is just a kind of staging buffer on the receive side.

OlivierDehaene: I know. 
I'm saying if the point is to make things the most async possible, fill this buffer in an async task and either return the buffer directly if it's big enough or None. I don't understand why you fill it when calling `next_batch`. To me this is nullyfing the benefits.
I will write a PR with this logic monday morning to illustrate what I mean.

njhill: I guess you could have the batcher loop try to fill the buffer while waiting for the shards to return (up to a max of max_batch_size elements), i.e. wait on both the queue and the response future together with `select!`. My intuition is that the time to transfer from the queue via the buffer is negligible and so this wouldn't buy you much, but it could be worth an experiment.

It might also be nice to arrange thing such that the batcher loop is pinned to a dedicated thread... I think this is possible with tokio.

OlivierDehaene:
Ok I now see what you mean.

The code is using a BTreeMap mainly to provide the flexibility to range on more complicated keys in the future but for now I admit that it is overkill. 

However, I'm still not convinced that its impact on performance is big enough to warrant a rewrite. 
I will add distributed tracing in the next days/weeks so I will have prod numbers soon enough.

OlivierDehaene:
Closed in favor of #52.
I will add you as co-author.

PR #27, opened 2023-01-20
feat(router): Remove second lock from batcher hot path

OlivierDehaene:
@njhill

### DIFF
--- a/router/src/db.rs
+++ b/router/src/db.rs
@@ -2,7 +2,7 @@ use crate::InferResponse;
 /// This code is massively inspired by Tokio mini-redis
 use crate::{GenerateParameters, GenerateRequest};
 use parking_lot::Mutex;
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 use std::sync::Arc;
 use text_generation_client::{
     Batch, ClientError, NextTokenChooserParameters, Request, StoppingCriteriaParameters,
@@ -112,18 +112,12 @@ impl Db {
         state.entries.insert(id, entry);
     }
 
-    /// Remove an entry from the database if it exists
-    pub(crate) fn remove(&self, id: &u64) -> Option<Entry> {
-        let mut state = self.shared.state.lock();
-        state.entries.remove(id)
-    }
-
     // Get the next batch
     pub(crate) fn next_batch(
         &self,
         min_size: Option<usize>,
         max_size: usize,
-    ) -> Option<(Vec<u64>, Batch)> {
+    ) -> Option<(HashMap<u64, Entry>, Batch)> {
         // Acquire lock
         let mut state = self.shared.state.lock();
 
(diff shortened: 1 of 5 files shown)


router/src/db.rs
```diff
@@ -135,13 +129,19 @@ impl Db {
                     return None;
                 }
             }
+            // Batch size
+            let size = requests.len();
+
+            let mut entries = HashMap::with_capacity(size);
```
njhill, line 135: `IntMap` from the [nohash_hasher](https://docs.rs/nohash-hasher/latest/nohash_hasher/) crate is much faster for integer keys. The default HashMap hashing is kind of slow.

router/src/db.rs
```diff
@@ -112,18 +112,12 @@ impl Db {
...
-        state.entries.remove(id)
-    }
-
     // Get the next batch
     pub(crate) fn next_batch(
         &self,
         min_size: Option<usize>,
         max_size: usize,
-    ) -> Option<(Vec<u64>, Batch)> {
+    ) -> Option<(HashMap<u64, Entry>, Batch)> {
```
njhill, line 120: I guess this ends up slightly cleaner code-wise than adding to an existing hashmap as is done my PR, but it means a bit more "happy path" work is done when extending a batch ... i.e. populating a new map and then transferring its contents to another one rather than just inserting the new entries once.

router/src/server.rs
```diff
@@ -30,16 +27,6 @@ async fn health(state: Extension<ServerState>) -> Result<(), (StatusCode, Json<E
...
-    // Limit concurrent requests by acquiring a permit from the semaphore
-    let _permit = state.limit_concurrent_requests.try_acquire().map_err(|_| {
-        (
-            StatusCode::TOO_MANY_REQUESTS,
-            Json(ErrorResponse {
-                error: "Model is overloaded".to_string(),
-            }),
-        )
-    })?;
-
```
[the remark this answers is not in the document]
OlivierDehaene: Yes I know I will move the validation logic to the batcher too.

router/src/db.rs
OlivierDehaene: How much faster are we talking about here?

njhill: Everything is relative :) but it's a drop-in replacement. Here is a [reference](https://nnethercote.github.io/perf-book/hashing.html).

router/src/db.rs
njhill: Here's a [ref](https://nnethercote.github.io/perf-book/heap-allocations.html#reusing-collections) for this one.

router/src/server.rs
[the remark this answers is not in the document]
OlivierDehaene: For now I moved it back. I will need to move it to the batcher at some point.

router/src/db.rs
OlivierDehaene: done

router/src/db.rs
OlivierDehaene: I preffer the api like this. I ll keep this in my mind though if we need to squeeze performance later.

PR #42, opened 2023-01-31
fix(server): fix seeding on gpu

OlivierDehaene:


server/text_generation/utils.py
```diff
@@ -89,6 +97,7 @@ def from_pb(cls, pb: generate_pb2.NextTokenChooserParameters) -> "NextTokenChoos
             top_p=pb.top_p,
             do_sample=pb.do_sample,
             seed=seed,
+            device=str(device),
```
njhill, line 100: No need to convert to string?

PR #44, opened 2023-01-31
fix(server): fix seeding with multiple shards

OlivierDehaene:


njhill:
@OlivierDehaene curious, did you observe incorrect behaviour in the sharded case? It seems to work fine for me without needing to set explicitly when a seed isn't provided.. I guess since the operations are still synchronized across the shards for the default generator (with initial set with `manual_seed`).

njhill:
@OlivierDehaene ok thanks, interesting... I've not observed that.

PR #47, opened 2023-01-31
feat(server): support repetition penalty

OlivierDehaene:


server/text_generation/utils.py
```diff
@@ -68,15 +70,20 @@ def __init__(
         if top_p is not None and top_p < 1.0:
             warpers.append(TopPLogitsWarper(top_p=top_p))
             sampling = True
+        if repetition_penalty is not None and repetition_penalty != 1.0:
+            warpers.append(RepetitionPenaltyLogitsProcessor(penalty=repetition_penalty))
+            sampling = True
```
njhill, line 75: I think this could apply equally to greedy ... perhaps remove this line?

PR #48, opened 2023-02-01
feat(server): allow gpt-neox models with odd vocab sizes to be sharded

OlivierDehaene:


server/text_generation/utils.py
```diff
@@ -91,7 +91,7 @@ def from_pb(
             top_p=pb.top_p,
             do_sample=pb.do_sample,
             seed=pb.seed,
-            device=str(device),
+            device=device,
```
njhill, line 94: nit but there's also still `str` type in the `Sampling` constructor

PR #73, opened 2023-02-17
feat(router): add cors allow origin options

OlivierDehaene:


julien-c (approved):

router/src/server.rs
```diff
@@ -391,6 +393,13 @@ pub async fn run(
         .install_recorder()
         .expect("failed to install metrics recorder");
 
+    // CORS layer
+    let allow_origin = allow_origin.unwrap_or(AllowOrigin::any());
+    let cors_layer = CorsLayer::new()
+        .allow_methods([Method::GET, Method::POST])
+        .allow_headers([http::header::CONTENT_TYPE])
```
julien-c, line 400: i think this header (as well as a few other ones) are always accepted by browers anyways

https://developer.mozilla.org/en-US/docs/Glossary/CORS-safelisted_request_header

PR #130, opened 2023-03-19
added llama

yk:
note: needs an up-to-date version of transformers, i.e. an update to the server makefile download


njhill:
@yk apart from the sentencepiece lib I don't think the other additions should be needed. You can just point the existing code (with updated transformers) at the llama model and it should load it fine.

dongs0104:
The router uses `huggingface/tokenizers(rust)`, which cannot use the bytefallback feature in sentencepiece BytePairEncoding used by LlamaTokenizer.

This is currently being implemented by huggingface/tokenizers.

ollmer:
The expected PR in tokenizers: https://github.com/huggingface/tokenizers/pull/1183

Narsil:
This PR contains the converter: https://github.com/huggingface/transformers/pull/22264

The tokenizer should be good here: https://huggingface.co/hf-internal-testing/tiny-random-llama

However we are still wrapping things up in `transformers`. The large test suite will have to be run (since making llama generating a lot more work than required)

yk:
> @yk Did you get this pr running? I tried running it with the latest transformers and tokenizers but missing the parallel models in the forked transformers repo prevented me from running it.

yea I've forked my own transformers to have both. I'll close this now since it's incoming upstream

PR #132, opened 2023-03-23
add cookies to Client constructors and requests

dconathan:
I have a use case where we need to pass cookies (for auth reasons) to an internally hosted server.

Note: I couldn't get the client tests to pass - do you need to have an HF token?

```python
FAILED tests/test_client.py::test_generate - text_generation.errors.BadRequestError: Authorization header is correct, but the token seems invalid
```

### DIFF
--- a/clients/python/text_generation/client.py
+++ b/clients/python/text_generation/client.py
@@ -36,7 +36,11 @@ class Client:
     """
 
     def __init__(
-        self, base_url: str, headers: Optional[Dict[str, str]] = None, timeout: int = 10
+        self,
+        base_url: str,
+        headers: Optional[Dict[str, str]] = None,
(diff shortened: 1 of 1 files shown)


OlivierDehaene:
Thanks for your PR!

I think the tests are failing because the InferenceAPIClient/InferenceAPIAsyncClient constructors are now broken since you changed the signatures of the underlying clients.

Can you update them?

dconathan:
Ah didn't see that.  However - tests still fail (they fail for me on `main` before my changes...)

OlivierDehaene:
Thanks for your work!!

OlivierDehaene:
Version v0.4.0 was pushed to pypi with your update.

PR #141, opened 2023-03-26
feat(server): Add mypy-protobuf

njhill:
Generates .pyi files for protobuf stubs which provide strong typing information. Very helpful for IDE auto-completion, etc.


OlivierDehaene (approved):
That s cool!
I only knew about https://github.com/danielgtaylor/python-betterproto

PR #144, opened 2023-03-28
feat(server): add flash attention llama

OlivierDehaene:
Draft as we are waiting for the transformers team to validate the fast tokenizer


router/src/infer.rs
```diff
@@ -87,10 +87,14 @@ impl Infer {
...
-        let valid_request = self.validation.validate(request).await?;
+        let valid_request = self.validation.validate(request).await.map_err(|err| {
+            metrics::increment_counter!("tgi_request_failure", "err" => "validation");
+            tracing::error!("{err}");
+            err
+        })?;
 
         // MPSC channel to communicate with the background batching task
-        let (response_tx, response_rx) = mpsc::unbounded_channel();
+        let (response_tx, response_rx) = flume::unbounded();
```
njhill, line 97: @OlivierDehaene presumably this is because you found flume to be faster than tokio mpsc?

server/text_generation_server/models/model.py
```diff
@@ -33,11 +24,26 @@ def batch_type(self) -> Type[B]:
...
+        self, all_input_ids: List[int], offset: Optional[int] = None
+    ) -> Tuple[str, Optional[int]]:
         """Hack to hopefully support generate_stream for the maximum number of tokenizers"""
-        # append token to special decode token and decode both
-        result = self.tokenizer.decode(
-            [self.special_decode_token_id, token_id], skip_special_tokens=False
+
+        # Decode all token minus last one and all tokens
+        results = self.tokenizer.batch_decode(
+            [all_input_ids[:-1], all_input_ids],
```
njhill, line 34: For large outputs, might this affect performance (repeatedly decoding all tokens)?

router/src/infer.rs
OlivierDehaene: Not really. 
I added it because it simplifies the validation workers as flume is multi-producer multi-consumer and then I thought it was a bit weird to have flume for this and not for the other mpsc channels in the repo.

server/text_generation_server/models/model.py
OlivierDehaene: Yes.
This branch should not be merged in this state.

njhill: @OlivierDehaene what do you think about the idea of encapsulating this logic into an incremental decoder as I was proposing [here](https://github.com/huggingface/text-generation-inference/blob/65ff6a73b3d5a3bb0cafa3d667b0d077bfcfc3f5/router/src/decoder.rs#L284-L291) (in rust but I guess similar could be done in python). Which can handle all of the tokenizer-specific aspects and unicode boundaries etc. Just feed it the token ids and take back strings to be concatenated. This impl has been working very well for us.

njhill: By the way this also handles the subtle difference between the decoder and encoder-decoder cases, where for decoder-only it's preferable to just start concatenating to the pre-tokenized input text rather than roundtripping it.

PR #145, opened 2023-03-28
feat(server): reduce mlp and attn in one op for flash neox

OlivierDehaene:


OlivierDehaene:
> Do you expect that identical input should produce identical (or extremely close) results irrespective of batch membership? 

It's always tricky as some noise is added when you have bigger batches but they should be close, yes. The fused layer norm seems to also add a lot of noise. You can try using nn.LayerNorm instead and it might solve the issue entirely. I need to check if it's truly worth it to use this kernel.

We have a lot of discussions internaly around this topic at the moment but we don't have a solution for now... (cc: @gante)

Can you share some info on the prompt and the resulting generations?

gante:
Based on my (limited) empirical observation in `transformers`, changing the numerical representation has a more significant impact than having a slightly incorrect position embedding -- lower resolution representations are quite sensible to the presence of numerical masking (padding). 

Also, based on working on the TF side of things, the order of FP operations can also change the results.

TBH, as long as the quality of the output doesn't decrease, I wouldn't worry much about it! :D

OlivierDehaene:
Lol @njhill you were rigth: https://github.com/huggingface/text-generation-inference/pull/150.

njhill:
Thanks @OlivierDehaene. I've been playing with the latest fixed version but it still doesn't seem to be completely stable. The same input as a member of different size batches gives different results.

However, I found that the order of inputs within the batch doesn't affect the output, and the contents/size of the other inputs generally doesn't either (strangely, reducing another input in the batch to < 8 tokens then did cause the result of the first input to change).

So I have a feeling something is still wrong, hopefully these clues might help. This doesn't seem like an FP precision thing to me since it seems stable for certain kinds of changes.

gante:
> reducing another input in the batch to < 8 tokens then did cause the result of the first input to change

@njhill You can also reproduce this behavior with batched generation in `transformers`, especially with non-FP32 computations. The more masked tokens (padding) you have, the higher the odds of this happening. 

(I'd be happy to provide a reproducer if you're curious :) )

OlivierDehaene:
@njhill,
Would you have a snippet I could run to reproduce the issue? I am unable to trigger it on my side.

njhill:
Thanks @gante. If you're referring to the existing GPT-NeoX impl in `transformers` then yes I'm aware of the behaviour but I think it is a still-to-be-fixed bug right? I did start to look into fixing it, I will try to get back to that soon.

I do also understand that in general lower precision can cause some padding instability, but in my experience the differences are usually minor/subtle rather than completely different output.

I agree with you that the differences don't really matter, but for research/reproducibility purposes, people submitting greedy requests expect to get identical results.

@OlivierDehaene I am still experimenting and want to make sure I can repro with your exact code/image rather than my fork. As soon as I've confirmed that I'll share an example.

njhill:
@OlivierDehaene after more experimentation I found that @gante is right, this does indeed seem to be numerical precision related. Specifically, the variation seems to be quite significant with bf16 (even more so for the flash attention impl), but results are much more stable with fp16 (and I assume fp32 too though I didn't try that).

In general, fp16 seems to be a better default for inferencing, even if the model was trained with bf16. @OlivierDehaene WDYT about changing that in the code? It's not used on V100 etc anyhow.

njhill:
@OlivierDehaene so as I said with float16 things seem to be very stable, but I did find one example which gives different results:

In greedy mode, max 10 tokens

The single input `The capital city of England is` produces the output ` London. The capital city of Scotland is Edinburgh.`

When combined in a batch of size 2 along with `Here is an explanation of FlashAttention:`, it produces the output ` London.\n\nThe capital city of England is`

PR #149, opened 2023-03-30
feat(benchmark): tui based benchmarking tool

OlivierDehaene:
image

cc: @njhill


njhill:
Thanks @OlivierDehaene, looks great! I haven't had a chance to look at it closely yet but will soon.

OlivierDehaene:
Thanks!
I find very useful to compare modeling code or how the model behave with different number of shards. It is also useful to see what is the expected latency in worst case scenarios (maximum expected context size, maximum decode tokens) and pick the correct max batch size bellow a latency threshold. 
However it is not perfect as the batch never contains any padding which make some models look better but it's a first version.

PR #159, opened 2023-04-07
fea(dockerfile): better layer caching

OlivierDehaene:


njhill:
@OlivierDehaene by the way, flash-attention builds way faster if you `pip install ninja` first.

OlivierDehaene:
> @OlivierDehaene by the way, flash-attention builds way faster if you `pip install ninja` first.

This line installs ninja. But for github actions it doesn't matter much as the runners are not very powerfull. https://github.com/huggingface/text-generation-inference/pull/159/files#diff-dd2c0eb6ea5cfc6c4bd4eac30934e2d5746747af48fef6da689e85b752f39557R85

njhill:
Ah, sorry, somehow I missed that. I know what you mean about the runners, but on my CI system even though it still takes a long time (~45min) it's still almost half the time it took without ninja.

PR #162, opened 2023-04-09
fix(rust-client): use join_all instead of select_all to hopefully fix nccl issues

OlivierDehaene:


OlivierDehaene:
@njhill,
Make sure you get this change in your fork as it seems to finally fix the elusive NCCL errors we have been seing on our side.

njhill:
Thanks @OlivierDehaene! Actually I'd already made a similar change a while back which could explain why we haven't encountered similar errors.

With the changes to move the stopping evaluation into the router, it's no longer the case that every call to the shards involves a forward pass (synchronized via torch.distributed) - for example pruning completed batches. So it became important to ensure that the calls to all shards completed and weren't cancelled prematurely. You can see this done partially in my PR [here](https://github.com/huggingface/text-generation-inference/pull/138/files#diff-6f20b73b45e0fd61e1ba6d6b4c5e8371ffd08df6d26a0414d83bad825a133e7eR86). In our fork I actually have a slightly different version of this aimed at mitigating any performance implications of having to wait for all of the responses. I can open another PR with that in case you're interested.

OlivierDehaene:
> I can open another PR with that in case you're interested.

Sure

PR #193, opened 2023-04-18
feat(server): have FlashGPTNeoXModel support HF accelerate

njhill:
It seems to work fine and loads 4-10x faster for me depending on the storage/page cache (non-sharded 20B parameter model).

However, when loaded this way, inference appears to be 10-15% slower for some reason.

Mainly opening this for discussion and to see if there are any ideas about why it would slow down inferencing.


OlivierDehaene:
With the new weight loading, this should not be needed anymore.

PR #194, opened 2023-04-18
feat(server): avoid manipulating position_ids for non-applicable models

njhill:
Currently, position_ids are always maintained/updated in the CausalLM case but this is unnecessary for models like BLOOM which don't use them.


Narsil:
Closing as stale.

Thanks for the contribution Nick, happy to take some back now that we're back on Apache ! 

Cheers.

PR #213, opened 2023-04-20
feat(server): flash attention past key value optimizations

njhill:
This changes `batch.past_key_values` to be a single tensor rather than list for batch size == 1 case.

There are separate optimizations for bs == 1 and bs > 1:
- Single request: Preallocate past kv tensor, no copies during decoding
- Multi request: Avoid an additional allocation/copy of all the data every iteration (done by `torch.nn.functional.pad`)

I haven't tested this ported version but in my fork it appears to speed things up (like 10-15%).

### DIFF
--- a/server/text_generation_server/models/flash_causal_lm.py
+++ b/server/text_generation_server/models/flash_causal_lm.py
@@ -407,17 +407,17 @@ class FlashCausalLM(Model):
             past_key_values,
         )
 
-        # Initialize past padding tensor
-        if self.past_pad is None:
-            self.past_pad = present.new_zeros(present.shape[0], 1, *present.shape[2:])
-
         # Initialize past_key_values in prefill
         if batch.past_key_values is None:
+            # Initialize past padding tensor
+            if self.past_pad is None:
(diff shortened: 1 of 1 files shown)


server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -390,9 +407,20 @@ def generate_token(
...
+
         # Initialize past_key_values in prefill
         if batch.past_key_values is None:
-            batch.past_key_values = [None] * len(batch)
+            batch.past_pad = self.past_pad
+            if len(batch) == 1:
+                # Preallocate tensor for bs = 1 case
+                batch.past_key_values = torch.nn.functional.pad(
+                    present, (0, 0, 0, 0, 0, 0, 0, batch.requests[0].stopping_parameters.max_new_tokens)
+                )
```
njhill, line 421: We can avoid an additional allocation/copy here by tweaking the `forward` method of the various `flash_x_modeling` impls to optionally allocate a larger initial tensor.

OlivierDehaene:
Thanks a lot!

njhill:
@OlivierDehaene curious whether this also shows improvements in your benchmarks?

PR #214, opened 2023-04-21
feat(server): reduce memory requirement

njhill:
Currently the memory requirement is much higher than it needs to be due the way that past key values are pruned and concatenated.

This PR changes `CausalLM` to do these operations layer by layer so that the corresponding tensors can be freed incrementally. It should roughly halve the memory requirement and so allow much larger batch sizes and/or sequence lengths.

Also changed the pruning to remove any unused padding at the front the batch (for example when the longest input sequence in the batch gets pruned).

These changes are also kind of required for the non-flash-attention impl of #210 (which I can push soon), to ensure that assumptions around mem usage based on point-in-time batch contents remain valid.

@OlivierDehaene if you approve of this I can add the corresponding `Seq2SeqLM` changes.

### DIFF
--- a/server/tests/models/test_bloom.py
+++ b/server/tests/models/test_bloom.py
@@ -175,12 +175,14 @@ def test_causal_lm_generate_token_completion_multi(
         generations[1].generated_text.generated_tokens
         == default_multi_requests_bloom_batch.stopping_criterias[1].max_new_tokens
     )
+    # Copy stopping_criterias before filtering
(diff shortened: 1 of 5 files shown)


OlivierDehaene:
@njhill, tests are failing but I like the idea.

njhill:
@OlivierDehaene thanks, let me fix the tests soon.

njhill:
@OlivierDehaene tests fixed and I've also added the same for seq2seq_lm.

OlivierDehaene (approved):
Cheers!

PR #220, opened 2023-04-21
fix(server): fix flash batch filtering

OlivierDehaene:


njhill:
Oops! I will get build of your main fork working in my env so that I can test these properly in future...

OlivierDehaene:
No worries!
I will try to do a sprint next week to go for a stable v1.0 with a way better test coverage. This should have been caught by a CI.

njhill:
@OlivierDehaene I actually have a more comprehensive set of tests, but in our internal version we expose a gRPC server rather than webserver, and the request/response options/format is a bit different.

It's similar to the existing rust-based `integration_tests` but in python (which seemed more flexible for orchestrating these kinds of tests). I have yaml files for each kind of model with a whole load of different tests cases exercising various permutations of the input options, covering error cases etc. and generally when adding functionality just need to add one or two entries to these.

The python fixtures starts the launcher, load/shuffle all the test cases, and then runs them in parallel. Each one actually translates to two requests in this set that are running - streaming and non-streaming. Running so many of them in parallel also exercises the batching including pruning/concatenation etc.

yaml is preferred over json imo because it's easier to structure/read and you can include comments, etc.

Example of a single test case from the yaml files:
```yaml
# Include logprobs
- name: Include logprobs
  request:
    params:
      method: GREEDY
      stopping:
        maxNewTokens: 4
      response:
        generatedTokens: true
        tokenLogprobs: true
    requests:
      - {"text": "A very long story:\n"}
  response:
    responses:
      - generatedTokenCount: 4
        inputTokenCount: 6
        stopReason: MAX_TOKENS
        text: "The first time I"
        tokens:
          - logprob: -2.284745
            text: The
          - logprob: -2.995144
            text: "\u0120first"
          - logprob: -2.3437245
            text: "\u0120time"
          - logprob: -0.6051528
            text: "\u0120I"
```

PR #226, opened 2023-04-24
feat(router): use number of tokens in batch as input for dynamic batching

OlivierDehaene:
@njhill, 
I was thinking about your PR this morning and came up with this implementation. 

@Narsil,
what do you think?

### DIFF
--- a/router/src/infer.rs
+++ b/router/src/infer.rs
@@ -265,12 +265,13 @@ async fn batching_task(
                 let batch_max_tokens = batch.max_tokens;
                 let mut batches = vec![batch];
                 metrics::gauge!("tgi_batch_current_size", batch_size as f64);
+                metrics::gauge!("tgi_batch_current_max_tokens", batch_max_tokens as f64);
 
                 let min_size = match waiting_tokens {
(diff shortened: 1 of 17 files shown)


router/src/infer.rs
```diff
@@ -263,48 +262,50 @@ async fn batching_task(
...
-                            entries.extend(new_entries);
-                            batches.push(new_cached_batch);
-                        }
+                let min_size = match waiting_tokens {
+                    // If we didn't onboard any new requests since >= max_waiting_tokens, we try
+                    // to add a new batch even though its size might be small
+                    _ if waiting_tokens >= max_waiting_tokens => None,
+                    // Minimum size criteria
+                    _ => Some((batch_size as f32 * waiting_served_ratio).floor() as usize),
+                };
```
Narsil, line 275:
```suggestion
                let min_size = if waiting_tokens >= max_waiting_tokens{
                    None
                }else{
                    // Minimum size criteria
                    Some((batch_size as f32 * waiting_served_ratio).floor() as usize)
                };
```
Simplification maybe ?

Narsil (dismissed):
LGTM

PR #244, opened 2023-04-26
feat(router): new healthcheck that skips the queue

Narsil:


njhill:
This was also on my todo list :)

We definitely want it to avoid the queue somehow, but I think it's still useful for it to do a full forward pass to catch a wider class of possible errors. It's saved me more than once before when upgrading to a newer version (Kubernetes won't switch over until the new one passes the health check).

What I had in mind was to keep track of the time of the last successfully completed request and if this was within the last e.g. 20 seconds then return healthy, or otherwise if the queue is empty then do a dummy request like it does now. wdyt?

router/src/lib.rs
```diff
@@ -19,6 +20,20 @@ pub struct HubModelInfo {
     pub pipeline_tag: Option<String>,
 }
 
+#[derive(Clone, Debug, Serialize, ToSchema)]
+pub struct HealthResponse {}
```
OlivierDehaene, line 24: I don't think this is used.

router/src/lib.rs
```diff
@@ -19,6 +20,20 @@ pub struct HubModelInfo {
...
+
+#[derive(Clone, Debug)]
+pub struct Health {
+    pub client: ShardedClient,
+}
+
+impl Health {
```
OlivierDehaene, line 31: I guess you don't really need the impl since client is pub

router/src/server.rs
```diff
@@ -82,9 +82,43 @@ async fn get_model_info(info: Extension<Info>) -> Json<Info> {
...
 
+#[utoipa::path(
+    get,
+    tag = "Text Generation Inference",
+    path = "/health",
+    request_body = HealthRequest,
+    responses(
+        (status = 200, description = "Everything is working fine"),
+        (status = 500, description = "Text generation inference is down", body = ErrorResponse,
+            example = json ! ({"error": "unhealthy"})),
```
OlivierDehaene, line 93: missing error_type but I guess it's missing everywhere.

router/src/server.rs
```diff
@@ -82,9 +82,43 @@ async fn get_model_info(info: Extension<Info>) -> Json<Info> {
...
+    mut health: Extension<Health>,
+) -> Result<Json<()>, (StatusCode, Json<ErrorResponse>)> {
+    // TODO: while this is the best health check we can do, it is a bit on the heavy side and might
+    //       be a bit too slow for a health check.
+    //       What we should do instead is check if the gRPC channels are still healthy.
+
+    // Send a small inference request
+    health.client.health().await.map_err(|_| {
+        (
+            StatusCode::INTERNAL_SERVER_ERROR,
```
OlivierDehaene, line 108: Maybe 503?

router/src/server.rs
```diff
@@ -82,9 +82,43 @@ async fn get_model_info(info: Extension<Info>) -> Json<Info> {
...
+        (status = 500, description = "Text generation inference is down", body = ErrorResponse,
+            example = json ! ({"error": "unhealthy"})),
+    )
+)]
+#[instrument]
 /// Health check method
+async fn health(
+    mut health: Extension<Health>,
+) -> Result<Json<()>, (StatusCode, Json<ErrorResponse>)> {
+    // TODO: while this is the best health check we can do, it is a bit on the heavy side and might
```
OlivierDehaene, line 101: This can be removed.

server/text_generation_server/server.py
```diff
@@ -29,6 +29,11 @@ def __init__(self, model: Model, cache: Cache, server_urls: List[str]):
     async def Info(self, request, context):
         return self.model.info
 
+    async def Health(self, request, context):
+        if self.model.device.type == "cuda":
+            torch.zeros((2, 2)).to(device=f"cuda:{os.environ['RANK']}")
```
OlivierDehaene, line 34: You can .cuda() directly since the default device is set already to the correct rank.

OlivierDehaene:
Hey @njhill!

One option would be to have an atomic bool that tracks GenerationErrors and once we get one we switch the logic to checking the healthyness of the shards with client.prefill with max_new_tokens = 1 to avoid caching. 
If it doesn't return an error we can switch back to the basic gRPC health check and discard it as a one of. 
If not we continue with the client.prefill check until we get killed by the livenessProbe.

njhill:
@OlivierDehaene something like that sounds good. But what what about the readiness case before there have been any "real" requests, we still want to make sure that a prefill completes successfully before starting to return success to the health probe.

router/src/server.rs
Narsil: Yes.

server/text_generation_server/server.py
Narsil: Done.

router/src/server.rs
Narsil: I added it.

router/src/lib.rs
Narsil: I started to do impl as I prefer to pubbing members in general (since modifying attributes on the fly is a rare requirement and it's enabled with pub attributes).

I went the full pub route for homogeneity's sake.

OlivierDehaene (approved):

PR #248, opened 2023-04-27
feat(server): add watermarking tests

ehsanmok:
Thank you for providing such a high quality library. This PR contains some improvements that I've found while skimming through the code.

### DIFF
--- a/router/src/health.rs
+++ b/router/src/health.rs
@@ -4,8 +4,9 @@ use text_generation_client::{
     Batch, NextTokenChooserParameters, Request, ShardedClient, StoppingCriteriaParameters,
 };
 
+// Note: Request ids and batch ids cannot collide.
 const LIVENESS_ID: u64 = u64::MAX;
-const BATCH_ID: u64 = u64::MAX - 1;
+const BATCH_ID: u64 = u64::MAX;
 
 #[derive(Clone, Debug)]
 pub(crate) struct Health {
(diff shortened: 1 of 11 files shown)


OlivierDehaene (commented):
Thanks!

router/src/health.rs
```diff
@@ -4,6 +4,9 @@ use text_generation_client::{
     Batch, NextTokenChooserParameters, Request, ShardedClient, StoppingCriteriaParameters,
 };
 
+const LIVENESS_ID: u64 = u64::MAX;
+const BATCH_ID: u64 = u64::MAX - 1;
```
OlivierDehaene, line 8: why -1?

server/pyproject.toml
```diff
@@ -35,6 +35,9 @@ bnb = ["bitsandbytes"]
 grpcio-tools = "^1.51.1"
 pytest = "^7.3.0"
 
+[tool.pytest.ini_options]
```
OlivierDehaene, line 38: We don't have asyncio tests in the server.

router/src/health.rs
ehsanmok: Avoiding collision of special ids?

server/pyproject.toml
ehsanmok: Ah! good catch

PR #251, opened 2023-04-27
fix(server): Small tidy of code from recent changes

njhill:
remaining_decode_tokens was calculated twice in Seq2SeqLMBatch.filter()


OlivierDehaene (approved):
Thanks

PR #257, opened 2023-04-28
doc(launcher): add more docs to the `launcher` itself and link in the README

Narsil:
This should show on

```
text-generation-launcher --help
```

I wrote the docstring a bit informally trying to describe things as I would describe a colleague instead of purely formally.
Might definitely need some refinement.

I like documenting here as it self documents the cli AND from the readme (since we refer to it).
Launcher has all the options from both the shards and the webserver making it a good unique source of all options.


OlivierDehaene (approved):
Super cool thanks!

PR #258, opened 2023-04-28
Adding docs on how dynamic batching works.

Narsil:
This PR starts the minimal possible amount of explanation I could think
of. It tries to explain how dynamic batching occurs, the interactions
with past key values and ignores the padding problem.

Maybe some drawings could help too but I kept it to text for now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,7 +42,8 @@ to power LLMs api-inference widgets.
 - Serve the most popular Large Language Models with a simple launcher
 - Tensor Parallelism for faster inference on multiple GPUs
 - Token streaming using Server-Sent Events (SSE)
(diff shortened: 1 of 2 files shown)


njhill:
@Narsil I have been referring to what's done here as "continuous batching" to differentiate from other kinds of dynamic batching.

Typical auto-batching that is done could also be considered dynamic (since batches are filled by the server) but the unique thing here is that we are removing/adding requests to the batch while the generation is in progress. So in theory you could have a continuous batch that never ends...

I have some drawings too that I can share.

OlivierDehaene:
@njhill, 
I really like the term "continuous batching" and I think it fits perfectly with the concept!

Narsil:
I renamed the concept. It is indeed more precise than just dynamic !

OlivierDehaene (approved):
Nice :)

PR #259, opened 2023-04-28
Specify response schema for compat_generate

feynmanliang:
We are generating an Elixir client from the OAPI spec.

Since there is no response schema reference for `/compat_generate`, the generated clients [currently output a `nil` for its response type](https://github.com/fmops/oapi_hf/blob/a7a7b55086a093a3d4306c2f30d90aae773bb233/lib/text_generation_inference/text_generation_inference.ex#L24).

This PR fixes this by adding the `GenerateResponse` schema to it.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -33,7 +33,14 @@
         },
         "responses": {
           "200": {
-            "description": "See /generate or /generate_stream"
+            "description": "Generated Text",
+            "content": {
+              "application/json": {
+                "schema": {
+                  "$ref": "#/components/schemas/GenerateResponse"
+                }
+              }
+            }
           },
           "422": {
             "description": "Input validation error",


OlivierDehaene (commented):
The reason this route does not have a reponse schema is because the schema changes dynamically with the `stream` boolean parameter. 
I don't think this can be expressed with OpenAPI but I might be wrong here.

Either way, if you are building a client for tgi, you should not use this route and use `/generate` and `/generate_stream` instead, except if you deploy tgi in a environment where you can't ingress on more that one route.

feynmanliang:
I see, that makes sense thank you!

I was trying to use this against https://huggingface.co/inference-api, which I believe implements `compat_generate`.

feynmanliang:
There is https://swagger.io/docs/specification/data-models/oneof-anyof-allof-not/#oneof for handling this coproduct / discriminated union return type, but I'm not sure how well supported it is by generators. I'll give it a shot and report back.

OlivierDehaene:
Nice I wasn't aware that this existed.
The current OpenAPI json is automatically generated in the router by [utoipa](https://docs.rs/utoipa/latest/utoipa/). If you can find a way to add it [here](https://github.com/huggingface/text-generation-inference/tree/main/router/src#L40) that would be great.

gsaivinay:
I've opened as PR adding this. Please take a look at #292

feynmanliang:
Nice, thank you @gsaivinay !

PR #261, opened 2023-04-29
Minor Dockerfile update

akashsonowal:
syntax correction update: pytorch==$PYTORCH_VERSION to "pytorch==$PYTORCH_VERSION"


OlivierDehaene (commented):
Is there any issue with the current dockerfile?
All cached layers will be invalidated by this change so if it's for style/lint only I will advocate against it.

PR #266, opened 2023-05-02
feat(server): support hf endpoint weight layout

OlivierDehaene:


server/text_generation_server/models/bloom.py
```diff
@@ -223,6 +223,15 @@ def linear(input, weight, bias):
...
 
+        uninitialized_parameters = []
+        for n, p in model.named_parameters():
+            if p.data.device == torch.device("meta"):
+                uninitialized_parameters.append(n)
+        if uninitialized_parameters:
+            raise RuntimeError(
+                f"found uninitialized parameters in model: {uninitialized_parameters}"
+            )
+
```
Narsil, line 234: I think this works, maybe we could have a common function ? :)

Narsil (approved):
LGTM

PR #282, opened 2023-05-04
feat: add snapshot testing

OlivierDehaene:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


njhill:
@OlivierDehaene I'm pretty sure it should be possible with `syrupy` but it's much nicer to have the snapshots serialized as yaml. This is mentioned for example here https://pypi.org/project/pytest-snapshot/.

IMO though it might be better to "roll your own" so that you can have multiple entries in a yaml file which each encapsulate an entire test case including inputs and expected outputs. That way it's easy to define a bunch with different permutations of parameters etc. The snapshot functionality is pretty easy to implement.

Then as features are added, testcase entries can be added to the yaml files without having to write separate tests in code.

To handle logprobs, you can just preprocess the dict to be compared by traversing it and replacing any floats with rounded values. I think this is preferable to excluding them altogether.

Narsil:
> To handle logprobs, you can just preprocess the dict to be compared by traversing it and replacing any floats with rounded values. I think this is preferable to excluding them altogether.

100% !

> IMO though it might be better to "roll your own"

Maybe at some point, fixing whatever tiring  points may come up. IMHO, let's ship some integration first, and fix whatever comes up. YAML for custom DSL *might* solve some problem, but maybe we can implement only when we actually have an issue with the current solution, wdyt ?

njhill:
@Narsil apologies, missed this before somehow. 
> IMHO, let's ship some integration first

SGTM!

> YAML for custom DSL might solve some problem, but maybe we can implement only when we actually have an issue with the current solution, wdyt ?

Calling it a DSL is probably a bit of a stretch. I was only suggesting to store pairs of input and expected output json of requests to be tested, and that having it persisted in YAML form is much more readable/maintainable than raw json (and can be commented, etc).

PR #285, opened 2023-05-05
feat(server): optim flash causal lm decode_token

OlivierDehaene:


server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -452,42 +469,73 @@ def generate_token(
...
-                batch.past_key_values = present
-            else:
-                # Add padding after each sequence
-                # This will have the correct shape after the final past_key_values concatenation before the model
-                # forward
-                batch.past_key_values = [None, self.past_pad] * len(batch)
+            )
+
+            # It is actually faster to do a whole other for loop here as the copy from present to past is fairly slow
+            # and will run asynchronously while we do the next for loop
```
njhill, line 516: Interesting :)

I guess torch.cat isn't smart enough for this?

OlivierDehaene: Well this is what torch.cat is doing at batch size 256. Kinda scary.

image

OlivierDehaene:
This is completely orthogonal to the warper vectorization. Just nice to have a small speedup just for the price of a small refacto.

server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -579,49 +708,41 @@ def generate_token(
...
                 # Remove generated token to only have prefill and add nan for first prompt token
-                prefill_logprobs = [float("nan")] + logprobs.gather(
-                    1, all_input_ids_tensor[1:input_length].unsqueeze(1)
-                ).squeeze(1)[:-1].tolist()
+                request_prefill_logprobs = [float("nan")] + prefill_logprobs[
+                    start_index : end_index - 1
+                ]
                 prefill_token_ids = all_input_ids[:-1]
                 prefill_texts = self.tokenizer.batch_decode(
                     prefill_token_ids,
```
jlamypoirier, line 716: `.cpu()` here for a big speedup

OlivierDehaene: What?

jlamypoirier: Ah, that's different from `CausalLM` where `prefill_token_ids` is a tensor.

PR #302, opened 2023-05-09
feat(server): use cuda graph in logits warping

OlivierDehaene:


njhill:
@OlivierDehaene this looks cool, do you know how it compares performance-wise to the vectorized impl in @jlamypoirier's PR? Could the two be combined?

OlivierDehaene:
The performance in prod are not so great unfortunately.
I will investigate further tomorrow.
The two can be merged for sure

PR #328, opened 2023-05-15
feat(ci): custom gpu runners

OlivierDehaene:


.github/workflows/build.yaml
```diff
@@ -85,7 +85,7 @@ jobs:
         with:
           cosign-release: 'v1.13.1'
       - name: Tailscale
-        uses: tailscale/github-action@v1
+        uses: tailscale/github-action@7bd8039bf25c23c4ab1b8d6e2cc2da2280601966
```
Narsil, line 88: Really ?:)

OlivierDehaene: Yes they didn't release a new version and @v1 has a bug =(

PR #329, opened 2023-05-15
Single place for TP layers + Dropout Layer Norm + FastLinear

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


server/text_generation_server/models/t5.py
```diff
@@ -10,24 +10,18 @@
     AutoModelForSeq2SeqLM,
     AutoConfig,
 )
-from transformers.models.t5.parallel_layers import (
```
OlivierDehaene, line 13: These are not the same.

Narsil: Oh really ?? How come, what's the difference ?

Those are in your fork right ? Isn't there a way to use the same ones ?

Narsil: https://github.com/OlivierDehaene/transformers/blob/text_generation_inference/src/transformers/models/t5/parallel_layers.py

The difference isn't glaring.

Narsil: Ok, maybe I need to finish the work then. I intended to remove all of these, to get a single loading point, where we don' t have to do post weight passes or after the fact transpositions.

Just load everything directly correctly. But it's starting to become quite a large PR (the full one).

OlivierDehaene: I'm fine with large prs ^^

Narsil: I'll still edit that one so that it can pass on its own if you're ok with it.
We still need to move `t5` to custom modeling in order to be on `transformers@main` right ?

OlivierDehaene: yes!

edward-io:
Should the `parallel_layers` code be committed on the transformers repo first before this PR is committed?

Narsil:
We're moving back on `transformers@main` https://github.com/huggingface/text-generation-inference/actions/runs/5059392097/jobs/9080991720?pr=344

PR #353, opened 2023-05-22
v0.7.0

OlivierDehaene:


server/text_generation_server/cli.py
```diff
@@ -71,7 +71,7 @@ def download_weights(
     model_id: str,
     revision: Optional[str] = None,
     extension: str = ".safetensors",
-    auto_convert: bool = True,
```
Narsil, line 74: I think this is what's breaking the integration tests, no ?

PR #362, opened 2023-05-23
feat(server): do not use device_map auto on single GPU

OlivierDehaene:


njhill:
@OlivierDehaene models seem to load much faster with accelerate even in single GPU case (see also https://github.com/huggingface/text-generation-inference/pull/193).

Narsil:
This reword logic will not changed the `from_pretrained` logic which is defaulted for non officially supported models though.

PR #373, opened 2023-05-28
Improve num_shard support with CUDA_VISIBLE_DEVICES=all

johnj:
# What does this PR do?

Make num_shards mirror available GPU's if CUDA_VISIBLE_DEVICES is set to "all".

Setting CUDA_VISIBLE_DEVICES=all in a podman-based (cdi) setup effectively fails to use the GPU at all, and limits  num_shard to 1.

This PR attempts to recognize CUDA_VISIBLE_DEVICES being set to "all", then counts the available GPU's and set the num_shard accordingly to what is exposed into the container.

The PR is only iterating standalone GPU's and not MIG's because you cannot use the multiple MIG's instance in the same process.

/cc @OlivierDehaene @Narsil

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -10,14 +10,13 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aes"
-version = "0.7.5"
+version = "0.8.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "9e8b47f52ea9bae42228d07ec09eb676433d7c4ed1ebdf0f1d1c29ed446f1ab8"
+checksum = "433cfd6710c9986c576a25ca913c39d66a6474107b406f34f91d4a8923395241"
 dependencies = [
  "cfg-if",
  "cipher",
  "cpufeatures",
- "opaque-debug",
 ]
 
 [[package]]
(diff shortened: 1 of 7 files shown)


johnj:
Testing environment: ubuntu 22.04 and no MIG setup (A100).

Command:

podman run --network host --shm-size 1g --rm --security-opt=label=disable --device=nvidia.com/gpu=7 -e CUDA_VISIBLE_DEVICES="all" ghcr.io/huggingface/text-generation-inference:latest --model-id ... ghcr.io/huggingface/text-generation-inference:latest --port 8080

OlivierDehaene:
Doesn't 

```shell
podman run --network host --shm-size 1g --rm --security-opt=label=disable --device=nvidia.com/gpu=all  ghcr.io/huggingface/text-generation-inference:latest --model-id bigscience/bloom-560m
```

work out of the box? Notice the removal of the env var.

johnj:
hi @OlivierDehaene, the lack of the env var in my comment is a copy/paste error. Good catch.

Without CUDA_VISIBLE_DEVICES=all this works fine, although only with CPU support and 1 shard in our setup.

Going to num_shard > 1 always results in an assertion error about num_gpu being greater than device count.

We also observed that with cuda 12 CUDA_VISIBLE_DEVICES=all throws a hard exception and with multiple shards, thought that an error within torch.

I honestly didn’t expect a response so quickly, I still need to modify this patch to translate all to individual device UUID’s if cuda 12 is detected. That is forth coming.

Narsil:
Is `all` expected at all ? 

https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#env-vars

Otherwise, there's no need to do parsing of anything, returning `None` here should take care of using all available GPUs ! 
However, unless I'm mistaken `CUDA_VISIBLE_DEVICES=all` is the bug, not the current code.

Narsil:
For the doc you linked the env variable is `NVIDIA_VISIBLE_DEVICES` not `CUDA_VISIBLE_DEVICES`. Maybe that explains it ?

Maybe we should even just not look at any nvidia env variable and use other mecanism to "read" number of GPUs so we're agnostic to it.

Narsil:
> The concern I would raise here that it may cause implicit behavior changes for existing setups - how do you feel about this?

It's already the case with `CUDA_VISIBLE_DEVICES`. Here the current code just tries to determine how many shards we should spawn when user didn't specify, so all visible GPUs should spawn. But this env variable implicitly modifies the number of visible devices. Maybe there's a better way to access all available GPUs that is maintained by nvidia and work with all the possible env variables at once.

johnj:
> Maybe there's a better way to access all available GPUs that is maintained by nvidia and work with all the possible env variables at once.

I’ll go investigate this further and report back on this PR.

johnj:
Still working through this.

I don’t want to exec out to nvidia-smi for this. So far the alternatives I’ve explored to that are:

1) emulating nvidia-smi - traverse /dev/nvidia* devices and issue ioctl’s to get device info (the card is not in a fault state, etc). I feel this is fragile because there are no guarantees or docs for this I could find.

2) use rust-cuda to get device info … this project seems somewhat inactive.

PR #406, opened 2023-06-02
feat(server): only compute prefill logprobs when asked

### LINKED ISSUE #288: Make generation details optional
### Feature request

* Add a `details` argument to the python client (it's currently hard-coded to `True`).
* Forward the `details` argument to the model through `generate_pb2.Request`.
* Only evaluate the details when requested.

### Motivation

The `details` computation comes with a significant overhead and its result is thrown away in the vast majority of cases. It should be done only when needed.

### Your contribution

I did the client part in #272, and will do an example of optional computation for `VectorizedCausalLM`. I can't work on the argument forwarding because it involves the rust part.

OlivierDehaene:
Close #288

### DIFF
--- a/benchmark/src/generation.rs
+++ b/benchmark/src/generation.rs
@@ -136,6 +136,7 @@ async fn prefill(
     let requests = (0..batch_size)
         .map(|id| Request {
             id: id.into(),
+            prefill_logprobs: false,
             inputs: sequence.clone(),
             truncate: sequence_length,
             parameters: Some(parameters.clone()),
--- a/clients/python/README.md
+++ b/clients/python/README.md
@@ -107,6 +107,40 @@ print(text)
 ### Types
 
 ```python
+# Request Parameters
+class Parameters:
+    # Activate logits sampling
+    do_sample: bool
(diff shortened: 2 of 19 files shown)


njhill:
@OlivierDehaene just a suggestion: "prefill" kind of refers to an internal detail, WDYT about referring to this in the API as e.g. "input_logprobs" or "input_token_logprobs" or "prompt_logprobs". Since from an API user's pov it's the details of the input tokens that you are requesting.

OlivierDehaene:
I agree and disagree. The name is not really good but other names imply that these values are always available while encoder-decoder models are unable to compute them.

njhill:
Good point, but then imo it just depends on whether the goal is for the API to abstract/hide the difference between encoder-decoder and decoder-only models. It seems reasonable to document that the parameter only applies to decoder-only models. If not, I feel that even "decoder_input_logprobs" would be more meaningful at the API level.

OlivierDehaene:
Is `decoder_input_details` ok with you?

njhill:
Yes, that sounds fine to me! (and this is only a suggestion of course)

PR #434, opened 2023-06-08
feat(server): optimize dist ops

OlivierDehaene:


server/text_generation_server/models/custom_modeling/flash_neox_modeling.py
```diff
@@ -265,7 +265,8 @@ def forward(
             mlp_output = self.mlp(ln2_hidden_states)
             intermediate = mlp_output + attn_output
 
-            torch.distributed.all_reduce(intermediate, group=self.process_group)
+            if self.process_group.size() > 1:
+                torch.distributed.all_reduce(intermediate, group=self.process_group)
```
Narsil, line 269: This is not needed. `all_reduce` is free.
`all_gather` is not free. (Sorry my bad when we discussed)

PR #438, opened 2023-06-09
Inference support for GPTQ (llama + falcon tested) + Quantization script

Narsil:
Let's start discussing implementation.

- Need to expose the quantization scripts (either included here or add doc on how to use https://github.com/qwopqwop200/GPTQ-for-LLaMa)
- Make sure GPTQ works for multiple models (priority to Falcon).

Currently it means that every place we use `get_{tensor|sharded}` to check for quantization.

My idea is to reintegrate as much as possible into `utils/layer.py` by expanding `load_multi` to be a bit more generic.
This might require some thinking, but ultimately the `qweight,qzeros,scales,g_idx` should be in a single place, and independant of bias presence.

# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


server/text_generation_server/utils/gptq/quant_linear.py
```diff
@@ -0,0 +1,325 @@
...
+            a_ptrs += BLOCK_SIZE_N
+            b_ptrs += BLOCK_SIZE_N
+            scales_ptrs += BLOCK_SIZE_N
+            zeros_ptrs += (BLOCK_SIZE_N // infearure_per_bits)
+
+        c_ptrs = c_ptr + stride_cm * offs_am[:, None] + stride_cn * offs_bk[None, :]
+        c_mask = (offs_am[:, None] < M) & (offs_bk[None, :] < K)
+        tl.store(c_ptrs, accumulator, mask=c_mask)
+except:
+    print('trioton not installed.')
```
gsaivinay, line 260: is it a typo? pls ignore if not.

gsaivinay: fixed the original file now https://github.com/qwopqwop200/GPTQ-for-LLaMa/blob/triton/quant/quant_linear.py#L260

Dockerfile
```diff
@@ -159,11 +159,14 @@ COPY --from=builder /usr/src/target/release/text-generation-router /usr/local/bi
...
 COPY --from=builder /usr/src/target/release/text-generation-launcher /usr/local/bin/text-generation-launcher
 
+RUN apt update && apt install build-essential g++ -y
+
 # AWS Sagemaker compatbile image
 FROM base as sagemaker
 
 COPY sagemaker-entrypoint.sh entrypoint.sh
 RUN chmod +x entrypoint.sh
+RUN pip install triton
```
OlivierDehaene, line 169: This should be installed earlier: it will only be available to the sagemaker target if you install it here.
Maybe it is not required? If I remember correctly, pytorch now also installs it but maybe I'm wrong.

server/text_generation_server/cli.py
```diff
@@ -150,6 +150,34 @@ def download_weights(
...
+    model_id: str,
+    output_dir: str,
+    revision: Optional[str] = None,
+    logger_level: str = "INFO",
+    json_output: bool = False,
+    trust_remote_code: bool = False,
+):
+    extension: str = ".safetensors",
+    # Remove default handler
+    logger.remove()
```
OlivierDehaene, line 164: I'm not sure how this works. Since you call download_weights below I don't think it's needed.

server/text_generation_server/utils/gptq/quantize.py
```diff
@@ -0,0 +1,1176 @@
...
+    use_cache = model.config.use_cache
+    model.config.use_cache = False
+    try:
+        layers = model.model.layers
+        prefix = "model.layers"
+    except Exception:
+        layers = model.transformer.h
+        prefix = "transformer.h"
+
+    # embeddings = model.get_input_embeddings()
```
OlivierDehaene, line 616: To remove maybe?

Dockerfile
Narsil: Doesn't seem to be the case.
https://github.com/pytorch/pytorch/blob/main/requirements.txt

I know I had to install it somewhere, are you ok with adding directly in the requirements ? (I did that for now, will revert if needed)

server/text_generation_server/cli.py
Narsil: Removed.

server/text_generation_server/utils/gptq/quant_linear.py
```diff
@@ -0,0 +1,394 @@
...
+            'BLOCK_SIZE_M': 32,
+            'BLOCK_SIZE_N': 128,
+            'BLOCK_SIZE_K': 32,
+            'GROUP_SIZE_M': 8
+        }, num_stages=2, num_warps=4),
+    ],
+                              key=['M', 'N', 'K'],
+                              nearest_power_of_two=True)
+    @triton.jit
+    def transpose_matmul_248_kernel(a_ptr, b_ptr, c_ptr, scales_ptr, zeros_ptr, g_ptr, M, N, K, bits, maxq, stride_am, stride_ak, stride_bk, stride_bn, stride_cm, stride_cn, stride_scales,
```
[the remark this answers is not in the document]
Narsil: Removed.

jiyuanq:
Hi @Narsil amazing work!

I tried your branch with https://huggingface.co/huggingface/falcon-40b-gptq today and was able to get 44ms per token on 4 x A10G, as opposed to 180ms per token I previously got with falcon-40b-instruct + bitsandbytes

Wondering when this PR will be merged and whether you will be uploading a `falcon-40b-instruct-gptq` as well? I think many including myself don't have access to a GPU with enough vram to quantize a 40b model

Narsil:
> Wondering when this PR will be merged and whether you will be uploading a falcon-40b-instruct-gptq as well? I think many including myself don't have access to a GPU with enough vram to quantize a 40b model

It should technically fit even if you have a very simple GPU, as long a you can load a single layer on the GPU. I didn't go too extreme, but we're leveraging `accelerate` there, which should offload to CPU RAM and disk when needed, and the quantization is doing it layer per layer (You still need to fit the model in CPU RAM though + the quantized weights in order to be able to save them).

Narsil:
> Is the act-order option supported?

Is exists in code as everything is simply pulled, but not exposed yet. Any good info of what act-order does, and implications ? (If there's now a good default we'll just use that).

For now it's False by default like in https://github.com/qwopqwop200/GPTQ-for-LLaMa

0x1997:
@Narsil It's slower but has better accuracy. https://github.com/lm-sys/FastChat/blob/main/docs/gptq.md

0x1997:
@Narsil For higher speed up of LLaMA models, you can checkout the https://github.com/turboderp/exllama project.
I tested it with two 13B models, both quantized with group size 128 and activation order enabled, and got 2.6x speed up over running the 16bit version with text-generation-inference.

Narsil:
> @Narsil For higher speed up of LLaMA models, you can checkout the https://github.com/turboderp/exllama project. I tested it with two 13B models, both quantized with group size 128 and activation order enabled, and got 2.6x speed up over running the 16bit version with text-generation-inference.

Looks pretty sweet indeed. Will have to dig a bit deeper to see how much we can reuse (everything seems super llama focused for now)

jiyuanq:
> > Wondering when this PR will be merged and whether you will be uploading a falcon-40b-instruct-gptq as well? I think many including myself don't have access to a GPU with enough vram to quantize a 40b model
> 
> It should technically fit even if you have a very simple GPU, as long a you can load a single layer on the GPU. I didn't go too extreme, but we're leveraging `accelerate` there, which should offload to CPU RAM and disk when needed, and the quantization is doing it layer per layer (You still need to fit the model in CPU RAM though + the quantized weights in order to be able to save them).

Do you have any recommendation for running the quantizer on GPUs with smaller vrams? I managed to load the 40b model with offloading, but it looks like the working memory required to run `fasterquant` is larger than what I could fit in my GPU 0. I even tried reducing nsamples to 64 but still got cuda OOM

Narsil:
Try using options here:: https://huggingface.co/docs/accelerate/usage_guides/big_modeling

Notably `device_map = infer_auto_device_map(my_model, max_memory={0: "10GiB", 1: "10GiB", "cpu": "30GiB"})` seems like a good option to reserve enough memory on GPU0 (you could say 0 on GPU0 since we're manually sending there).

Actually making that the default might be better than `balanced_low_0`

jiyuanq:
> Try using options here:: https://huggingface.co/docs/accelerate/usage_guides/big_modeling
> 
> Notably `device_map = infer_auto_device_map(my_model, max_memory={0: "10GiB", 1: "10GiB", "cpu": "30GiB"})` seems like a good option to reserve enough memory on GPU0 (you could say 0 on GPU0 since we're manually sending there).
> 
> Actually making that the default might be better than `balanced_low_0`

Thanks, I actually tried that, with GPU 0’s max memory set to 2GB but still no luck. Perhaps the memory required to run quantization is more than 22GB.
Another thing I noticed is that the line  

```
layer = layers[i].to(dev)
``` 

can lead to error “ NotImplementedError: Cannot copy out of meta tensor; no data”, likely because some layers are offloaded?

jiyuanq:
> That means the layers wasn't loaded at all, probably disk offloaded. I'm not familiar enough with accelerate internals, but there must be some way to fetch the information of where the weights live, so that you could fill them before sending on the GPU.

I see. Thanks! Let me dig a bit more to figure that out

Dockerfile
```diff
@@ -159,6 +159,8 @@ COPY --from=builder /usr/src/target/release/text-generation-router /usr/local/bi
 # Install launcher
 COPY --from=builder /usr/src/target/release/text-generation-launcher /usr/local/bin/text-generation-launcher
 
+RUN apt update && apt install build-essential g++ -y
```
OlivierDehaene, line 162:
```suggestion
RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
        build-essential \
        g++ \
        && rm -rf /var/lib/apt/lists/*
```

OlivierDehaene: Maybe this should be added at the top to avoid having one more layer.
This will invalidate the cache though.

server/text_generation_server/models/flash_llama.py
```diff
@@ -34,13 +34,22 @@ def __init__(
...
-        )
+        try:
+            tokenizer = LlamaTokenizer.from_pretrained(
+                model_id,
+                revision=revision,
+                padding_side="left",
+                truncation_side="left",
+                trust_remote_code=trust_remote_code,
+            )
+        except Exception:
```
OlivierDehaene, line 45: Can this exception be more specific?
Also, where is this new try/catch coming from?

server/text_generation_server/utils/gptq/quant_linear.py
```diff
@@ -0,0 +1,359 @@
...
+            qweight.stride(1),
+            output.stride(0),
+            output.stride(1),
+            scales.stride(0),
+            qzeros.stride(0),
+        )
+        return output
+
+
+class QuantLinearFunction(torch.autograd.Function):
```
OlivierDehaene, line 240: I guess you can remove this and call `matmul248` directly instead of doing `QuantLinearFunction.apply` as we do not have the backward anymore.

psinger:
Trying out the branch locally and running into this when trying to query the endpoint:

```
  File "/home/user/text-generation-inference/server/text_generation_server/utils/gptq/custom_autotune.py", line 90, in _bench
    return triton.testing.do_bench(
  File "/home/user/miniconda3/envs/text-generation-inference/lib/python3.9/site-packages/triton/testing.py", line 144, in do_bench
    torch.cuda.synchronize()
  File "/home/user/miniconda3/envs/text-generation-inference/lib/python3.9/site-packages/torch/cuda/__init__.py", line 688, in synchronize
    return torch._C._cuda_synchronize()
RuntimeError: CUDA error: an illegal memory access was encountered
CUDA kernel errors might be asynchronously reported at some other API call, so the stacktrace below might be incorrect.
For debugging consider passing CUDA_LAUNCH_BLOCKING=1.
Compile with `TORCH_USE_CUDA_DSA` to enable device-side assertions.

 rank=0
2023-06-19T17:52:01.967782Z ERROR batch{batch_size=1}:prefill:prefill{id=0 size=1}:prefill{id=0 size=1}: text_generation_client: router/client/src/lib.rs:33: Server error: CUDA error: an illegal memory access was encountered
```

Any idea? Trying it on a model I quantized using the script in this PR.

0x1997:
I got the following error when loading a [model](url) quantized with [auto-gptq](https://github.com/PanQiWei/AutoGPTQ)
> {"timestamp":"2023-06-20T02:07:57.613035Z","level":"ERROR","fields":{"message":"Shard 0 failed to start:\nTraceback (most recent call last):\n\n  File \"/opt/conda/bin/text-generation-server\", line 8, in <module>\n    sys.exit(app())\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/cli.py\", line 67, in serve\n    server.serve(model_id, revision, sharded, quantize, trust_remote_code, uds_path)\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/server.py\", line 155, in serve\n    asyncio.run(serve_inner(model_id, revision, sharded, quantize, trust_remote_code))\n\n  File \"/opt/conda/lib/python3.9/asyncio/runners.py\", line 44, in run\n    return loop.run_until_complete(main)\n\n  File \"/opt/conda/lib/python3.9/asyncio/base_events.py\", line 647, in run_until_complete\n    return future.result()\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/server.py\", line 124, in serve_inner\n    model = get_model(model_id, revision, sharded, quantize, trust_remote_code)\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/models/__init__.py\", line 178, in get_model\n    return FlashLlama(\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/models/flash_llama.py\", line 64, in __init__\n    model = FlashLlamaForCausalLM(config, weights)\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/models/custom_modeling/flash_llama_modeling.py\", line 417, in __init__\n    self.model = FlashLlamaModel(config, weights)\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/models/custom_modeling/flash_llama_modeling.py\", line 316, in __init__\n    [\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/models/custom_modeling/flash_llama_modeling.py\", line 317, in <listcomp>\n    FlashLlamaLayer(\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/models/custom_modeling/flash_llama_modeling.py\", line 248, in __init__\n    self.self_attn = FlashLlamaAttention(\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/models/custom_modeling/flash_llama_modeling.py\", line 112, in __init__\n    self.query_key_value = TensorParallelColumnLinear.load_multi(\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/utils/layers.py\", line 226, in load_multi\n    weight = weights.get_multi_weights_col(\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/utils/weights.py\", line 95, in get_multi_weights_col\n    w = [self.get_tensor(f\"{p}.g_idx\") for p in prefixes]\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/utils/weights.py\", line 95, in <listcomp>\n    w = [self.get_tensor(f\"{p}.g_idx\") for p in prefixes]\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/utils/weights.py\", line 47, in get_tensor\n    filename = self.get_filename(tensor_name)\n\n  File \"/opt/conda/lib/python3.9/site-packages/text_generation_server/utils/weights.py\", line 34, in get_filename\n    raise RuntimeError(f\"weight {tensor_name} does not exist\")\n\nRuntimeError: weight model.layers.0.self_attn.q_proj.g_idx does not exist\n\n"},"target":"text_generation_launcher"}
{"timestamp":"2023-06-20T02:07:57.613081Z","level":"INFO","fields":{"message":"Shutting down shards"},"target":"text_generation_launcher"}
Is it possible to support these models without requantization?

Narsil:
@psinger 

Illegal access seems like a  triton bug. Which GPU are you using ? I'm guessing if it's old, triton might be creating invalid kernels

@0x1997 Absence of `g_idx` means it's lacking the weight. I have no idea how they called it.
Please note that we also modified our source https://github.com/qwopqwop200/GPTQ-for-LLaMa to include `bits` and `groupsize` directly in the safetensors file so that we don't need to ask the user to know in advance to specify at load time.
I don't think either method does it that way.

psinger:
@Narsil I am using A100, so not old.

I tried running it locally and via generated docker image to make sure it is not an environment issue, but same issue happens in both.

Narsil:
@0x1997 But the error happens when it looks for `g_idx`. This has nothing to do with group_size=-1, has it ? (It's always defined no ?). IIUC, `groupsize=-1` simply means full row. It should be relatively easy to add support back in, but I'm not sure this is the issue at hand.

@psinger cuda version ? Do you mind creating an issue with all your setup + stacktrace. It's not really possible to help without all the information (and being able to reproduce)

psinger:
@Narsil If I am building the docker image, my local cuda shouldn't matter, or will it still have an impact?

server/text_generation_server/models/flash_llama.py
Narsil: For the quantized model I save the output of `AutoTokenizer.from_pretrained(..).save_pretrained()` which kills the slow tokenizer.

For the exception, I'm not sure what kind it is.

We could also just `AutoTokenizer` which would drop the exception and just work.

server/text_generation_server/utils/gptq/quantize.py
```diff
@@ -0,0 +1,866 @@
...
+    trust_remote_code: bool,
+    upload_to_model_id: Optional[str],
+    percdamp: float,
+    act_order: bool,
+):
+    print("loading model")
+    model = AutoModelForCausalLM.from_pretrained(
+        model_id,
+        torch_dtype=torch.float16,
+        device_map="balanced_low_0",
```
psinger, line 785: I believe this will break if one only has a single GPU.

Narsil: Check it ?

psinger: I did, it fails. Either you make this a setting, or change it to `"auto"` by default.

psinger:
@Narsil 

unfortunately I also cant get it to work on a fresh H100 system due to some mismatch

I documented all the steps, please let me know if I am missing something

```
git clone https://github.com/huggingface/text-generation-inference.git
git checkout support_gptq

curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh

conda create -n text-generation-inference python=3.9 
conda activate text-generation-inference

sudo apt-get install cargo
sudo apt-get install libssl-dev gcc -y

BUILD_EXTENSIONS=True make install

# ouch this takes some time...
cd server && install-flash-attention

source ~/.bashrc
conda activate text-generation-inference

# somehow those were missing
pip install texttable
pip install einops
pip install datasets

conda install cudatoolkit

text-generation-server quantize tiiuae/falcon-7b ../data/falcon-7b-gptq --tru
st-remote-code

text-generation-launcher --model-id /home/user/models/falcon-7b-gptq/ --quantize gptq --trust-remote-code --port 8080
```

Error in this case is different to before:

`python: /project/lib/Dialect/TritonGPU/Transforms/Combine.cpp:870: int {anonymous}::{anonymous}::computeCapabilityToMMAVersion(int): Assertion `false && "computeCapability > 90 not supported"' failed.`

Seems to be some cuda / package mismatch

psinger:
@jgcb00 yes, switching it to `"auto"` will fix it

please let me know if you manage to start the endpoint with that model

psinger:
This looks like A100? I managed to quantize a 7b model on a single GPU there.

jgcb00:
But I don't get it, we cannot convert the model but then it will easily run on one signle A100 ? 
I will try with the device cpu see if I can convert the model with that !

psinger:
You need to fully load the model in float16 to quantize it. After quantization you will need way less memory.
Quantizing it on CPU will be too slow I am afraid.

Narsil:
> @jgcb00 yes, switching it to "auto" will fix it

`auto` doesn't fix anything, since it will cram your GPU 0, and then there's not enough room to create the quantizations.

We need at least a single layer to load on GPU for everything correctly. `balanced_low_mem_0` means we're keeping GPU as free as possible so we can do everything we need there (and we do need the room).

The error you're seeing is imho an `accelerate` bug (at least the error message should be descriptive), even if there might be better balance. When using `accelerate` we want to load what we can, offload the rest to CPU and whatever is left to DISK.
I don't want to add support for handling DISK atm since it requires more `accelerate` internal (and this PR is already huge, we need to fix later, not here, this is not what this PR is about).

We could bypasse `accelerate`, but that requires even more logic, which is out of scope of this PR imo.

psinger:
`auto` definitely fixes if it fits on a single GPU - `balanced_low_0` assumes there are multiple GPUs, it will not offload to CPU afaik.

But I agree that `auto` is also not the best default, as larger models will require offloading. That's why I suggested this could be a parameter the user can set.

Narsil:
No. Adding flags everywhere is not good. If you know what you're doing you can edit the code yourself.
For the vast majority, we need to figure out the sanest defaults. In this case neither `auto` nor  `balanced_low_0` (actually maybe balanced_low_mem_0 if accelerate things it's ok to fix.

jgcb00:
At the end of the quantization I have : 
```
Done.

  File "/root/text-generation-inference/tgi_env/bin/text-generation-server", line 8, in <module>
    sys.exit(app())

  File "/root/text-generation-inference/server/text_generation_server/cli.py", line 174, in quantize
    quantize(

  File "/root/text-generation-inference/server/text_generation_server/utils/gptq/quantize.py", line 827, in quantize
    save_file(

  File "/root/text-generation-inference/tgi_env/lib/python3.9/site-packages/safetensors/torch.py", line 232, in save_file
    serialize_file(_flatten(tensors), filename, metadata=metadata)

  File "/root/text-generation-inference/tgi_env/lib/python3.9/site-packages/safetensors/torch.py", line 394, in _flatten
    raise RuntimeError(

RuntimeError: 
            Some tensors share memory, this will lead to duplicate memory on disk and potential differences when loading them again: [{'transformer.word_embeddings.weight', 'lm_head.weight'}].
            A potential way to correctly save your model is to use `save_model`.
            More information at https://huggingface.co/docs/safetensors/torch_shared_tensors
```

Narsil:
Do you mind opening issues for everything, with all the added info ? This is a PR, it would be easier to keep it about the content of the PR directly.

And fixes will come later don't worry but currently your giving incomplete issue descriptions, which makes it impossible to reproduce. 

Just specify that you're using this branch in the issue, we'll figure it out.

tienthanhdhcn:
Thanks @Narsil so much. Do you have any idea on when this PR will be merged?

Narsil:
@OlivierDehaene should review and merge probably soon

TheBloke:
Disappointed you're looking to PR GPTQ-for-LLaMa and not [AutoGPTQ](https://github.com/PanQiWei/AutoGPTQ).  Is there a reason for that?

AutoGPTQ is much cleaner and better integrated with transformers.  It also supports more model types and has far fewer breaking changes.

flozi00:
> Disappointed you're looking to PR GPTQ-for-LLaMa and not [AutoGPTQ](https://github.com/PanQiWei/AutoGPTQ). Is there a reason for that?
> 
> AutoGPTQ is much cleaner and better integrated with transformers. It also supports more model types and has far fewer breaking changes.

Totally agree with that
The only problem is that autogptq with falcon is very slow, TGI with bnb is still faster, all other models are really good

Narsil:
> Is there a reason for that?

Yes, not pulling the modeling code which we cannot use because this repo uses Flash Attention, which is a very important part of the speedup. And we need a quantize script which is as much as possible agnostic to the model under use, which autogptq isn't (afaik).

Pretty agnostic to using the cuda kernel vs the triton one to be honest.

> The only problem is that autogptq with falcon is very slow

Really ? Latency should be much better with GPTQ always compared to bnb there's no question about that. What kind of latency are you seeing ? 

If you could include outputs from `text-generation-benchmark` for both (ideally with the non quantized version too) that would be awesome.

flozi00:
I have no numbers at the moment, but with autogptq - falcon 7b model it was around 1 word per second only
There also was an issue in the autogptq repo where they spoke about that problem, i can search that later

Narsil:
> l it was around 1 word per second only

On what hardware ?

flozi00:
RTX A600 ada lovelace

> > l it was around 1 word per second only
> 
> On what hardware ?

RTX A6000 ada lovelace
Are the autogptq models compatible with the actual PR ?

flozi00:
https://github.com/PanQiWei/AutoGPTQ/issues/130#issue-1738555663

For example that issue in AutoGPTQ Repo

ronald-d-rogers:
Is there a new Docker container for this already or does that only come when a release is cut?

PR #453, opened 2023-06-14
feat(router): add ngrok integration

OlivierDehaene:
What do you think @Narsil?

Maybe we can hide behind a cargo feature, but then it's a bit of mess in the docker container. We will need to build multiple times with different feature set.


Narsil:
As discussed orally.

I'm OK with no cargo feature, but imho it should be a flag in the launcher itself, so there's a central place to activate ngrok or not.
By default our docker should build ngrok (it's convenient, even for us sometimes).

For users that are sensitive having a feature flag disabling ngrok would be a nice touch (so that someone could distribute the docker internally without being worried of users opening ngrok tunnels everywhere)

PR #529, opened 2023-07-03
Implement NTK-Aware scaled and dynamically scaled  RoPE for PositionRotaryEmbedding

### LINKED ISSUE #512: NTK-Aware Scaled RoPE allows LLaMA models to have extended (8k+) context size without any fine-tuning and minimal perplexity degradation.
### Feature request

Longer context up to 8k tokens, the given discussion and notebook generate promising results

### Motivation

Discussion: https://www.reddit.com/r/LocalLLaMA/comments/14lz7j5/ntkaware_scaled_rope_allows_llama_models_to_have/

Colab Notebook: https://colab.research.google.com/drive/1VI2nhlyKvd5cw4-zHvAIk00cAVj2lCCC#scrollTo=d2ceb547

### Your contribution

As it's only 3 lines of code to change it would be pretty easy to change

I will start training an model and give an example demo

iantbutler01:
# What does this PR do?
Implements NTK-Aware scaled and dynamically scaled RoPE for the PositionRotaryEmbedding to allow models to scale beyond their default max_tokens.

https://www.reddit.com/r/LocalLLaMA/comments/14lz7j5/ntkaware_scaled_rope_allows_llama_models_to_have/
https://www.reddit.com/r/LocalLLaMA/comments/14mrgpr/dynamically_scaled_rope_further_increases/

Fixes #512 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -35,9 +35,9 @@ struct Args {
     max_total_tokens: usize,
     #[clap(default_value = "1.2", long, env)]
     waiting_served_ratio: f32,
-    #[clap(default_value = "4096", long, env)]
+    #[clap(default_value = "2048", long, env)]
     max_batch_prefill_tokens: u32,
-    #[clap(default_value = "16000", long, env)]
+    #[clap(default_value = "8192", long, env)]
     max_batch_total_tokens: u32,
     #[clap(default_value = "20", long, env)]
     max_waiting_tokens: usize,
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -369,7 +369,7 @@ try:
     import rotary_emb
 
     class PositionRotaryEmbedding(nn.Module):
-        def __init__(self, inv_freq):
+        def __init__(self, inv_freq, scale_factor=1, dynamic_scaling=False, max_seq_len=2048, dim=None, base=None):
             super().__init__()
 
             self.inv_freq = inv_freq
@@ -379,32 +379,62 @@ try:
             self._cos_k_cached = None
             self._sin_k_cached = None
 
-        @classmethod
-        def static(cls, dim, base, device):
-            inv_freq = 1.0 / (
-                base
-                ** (torch.arange(0, dim, 2, device=device, dtype=torch.float32) / dim)
-            )
-            return cls(inv_freq)
+            self.scale_factor = scale_factor
+            self.dynamic_scaling = dynamic_scaling
+            self.original_max_seq_len = max_seq_len
+            self.max_seq_len = max_seq_len * scale_factor
+            self.dim = dim
+            self.base = base
 
+        @classmethod
+        def static(cls, dim, base, device, scale_factor=1, dynamic_scaling=False, max_seq_len=2048):
+            inv_freq = cls._get_inv_freq(dim, base, device, scale_factor)
+            return cls(inv_freq, scale_factor, dynamic_scaling, max_seq_len, dim, base)
+        
         @classmethod
         def load(cls, prefix, weights):
             # XXX: Always load this in float32 !
             dtype = weights.dtype
             weights.dtype = torch.float32
+
             inv_freq = weights.get_tensor(f"{prefix}.inv_freq")
             weights.dtype = dtype
             return cls(inv_freq)
 
+        @staticmethod
+        def _get_inv_freq(dim, base, device, scale_factor=1):
+            base = base * scale_factor ** (dim / (dim-2))
+
+            inv_freq = 1.0 / (
+                base
+                ** (torch.arange(0, dim, 2, device=device, dtype=torch.float32) / dim)
+            )
+
+            return inv_freq
+
         def _update_cos_sin_cache(self, dtype, device, seqlen):
             # Reset the tables if the sequence length has changed,
             # or if we're on a new device (possibly due to tracing for instance)
+            
+            length = seqlen
+            max_seq_len = self.max_seq_len
+            inv_freq = self.inv_freq
+
+            if self.dynamic_scaling:
+                scale_factor = (self.scale_factor * length / self.original_max_seq_len) - (self.scale_factor - 1)
+                max_seq_len = self.original_max_seq_len * scale_factor
+                inv_freq = self._get_inv_freq(self.dim, self.base, inv_freq.device, scale_factor)
+                self.register_buffer("inv_freq", inv_freq)
+
+            if self.scale_factor > 1:
+               length =  max(seqlen, max_seq_len)
+
             if (
-                seqlen > self._seq_len_cached
+                length > self._seq_len_cached
                 or self._cos_cached.device != device
                 or self._cos_cached.dtype != dtype
             ):
-                self._seq_len_cached = seqlen
+                self._seq_len_cached = length
                 t = torch.arange(seqlen, device=device, dtype=self.inv_freq.dtype)
                 # Don't do einsum, it converts fp32 to fp16
                 # freqs = torch.einsum("i,j->ij", t, self.inv_freq)
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -423,10 +423,9 @@ try:
             if self.dynamic_scaling:
                 scale_factor = (self.scale_factor * length / self.original_max_seq_len) - (self.scale_factor - 1)
                 max_seq_len = self.original_max_seq_len * scale_factor
-                inv_freq = self._get_inv_freq(self.dim, self.base, inv_freq.device, scale_factor)
-                self.register_buffer("inv_freq", inv_freq)
+                self.inv_freq = self._get_inv_freq(self.dim, self.base, inv_freq.device, scale_factor)
 
-            if self.scale_factor > 1:
+            if self.scale_factor > 1 and not self.dynamic_scaling:
                length =  max(seqlen, max_seq_len)
 
             if (
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -162,7 +162,7 @@ struct Args {
     /// Limits the number of tokens for the prefill operation.
     /// Since this operation take the most memory and is compute bound, it is interesting
     /// to limit the number of requests that can be sent.
-    #[clap(default_value = "4096", long, env)]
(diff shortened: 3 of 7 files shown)


iantbutler01:
I've tested NTK-Aware scaling on a project I'm working on and was successfully generating at 2400 tokens which is about the limit my RTX 6000 Ada can handle from VRAM with falcon 40BN instruct, but it was entirely coherent generation above the original 2048 token context.

I still need to test dynamic scaling and clean up the PR further to comply with guidelines and the checklist, but wanted to open this up in the meantime.

ssmi153:
Just a note that Huggingface Transformers natively supports this now: https://github.com/huggingface/transformers/commit/34d94094279d2c903d9d8a51a65edb265f22c849. Does this make it easier to implement here?

iantbutler01:
@ssmi153 Not particularly, most of the attention modules in this repo are custom to support flash attention. The work is good to review my implementation and that's about it from what I see.

server/text_generation_server/utils/layers.py
```diff
@@ -369,7 +369,7 @@ def forward(self, hidden_states, residual=None):
     import rotary_emb
 
     class PositionRotaryEmbedding(nn.Module):
-        def __init__(self, inv_freq):
+        def __init__(self, inv_freq, scale_factor=1, dynamic_scaling=False, max_seq_len=2048, dim=None, base=None):
```
Narsil, line 372: Can we have at most 1 extra argument.

A lot of information should be extractable directly from `inv_freq`.

server/text_generation_server/utils/layers.py
```diff
@@ -379,32 +379,61 @@ def __init__(self, inv_freq):
...
             # or if we're on a new device (possibly due to tracing for instance)
+            
+            length = seqlen
+            max_seq_len = self.max_seq_len
+            inv_freq = self.inv_freq
+
+            if self.dynamic_scaling:
+                scale_factor = (self.scale_factor * length / self.original_max_seq_len) - (self.scale_factor - 1)
+                max_seq_len = self.original_max_seq_len * scale_factor
+                self.inv_freq = self._get_inv_freq(self.dim, self.base, inv_freq.device, scale_factor)
```
Narsil, line 426: This is not really OK I think.

You ditching entirely the original `self.inv_freq` which unfortunately for us is sometimes different from the calculation proposed (that's why not all models are `static` and some are `load`.

Llama most notably has different saved `inv_freq` (not sure why but it's indeed the case).

Narsil (commented):
Thanks for the PR.

- Overall, can we remove a lot of the complexity ?
From what I read, dynamic scaling seems just better than static scaling, so let's just use dynamic scaling, no ?
The current code has a lot of pathways, can we keep them to a minimum ?
Keep the code as close to the original as possible
Nothing should be directly in `custom_modeling` file. This behavior it seems should be entirely agnostic of modeling code.

This can go in the config for instance (like quantize) and be in `models/flash_llama.py` for instance (this is not modeling code, but wrapping the model itself, this will probably be factored away at some point, but here would be a good place for now).

I'm happy to make those changes if you want, as they are mostly stylistic choices rather than business logic.

router/src/main.rs
```diff
@@ -35,9 +35,9 @@ struct Args {
     max_total_tokens: usize,
     #[clap(default_value = "1.2", long, env)]
     waiting_served_ratio: f32,
-    #[clap(default_value = "4096", long, env)]
+    #[clap(default_value = "2048", long, env)]
     max_batch_prefill_tokens: u32,
-    #[clap(default_value = "16000", long, env)]
+    #[clap(default_value = "8192", long, env)]
```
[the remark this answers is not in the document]
iantbutler01: Oh yup fair, I don't want to change them I meant to clean this out. I'll remove!

server/text_generation_server/utils/layers.py
iantbutler01: Part of dynamic scaling is calculating the inv_freq, looking at the implementation in Transformers I don't see them preserving this value either.

iantbutler01: What would you suggest alternatively?

server/text_generation_server/utils/layers.py
iantbutler01: Yup, I can try to simplify this

iantbutler01:
@Narsil thanks for the review!

So my only reason for suggesting we keep static scaling is that it's much easier to consider VRAM usage of a statically scaled context window. If you have a model with 2048 context default, and scale it by 4 to 8192 you can much more easily consider the VRAM consumption.

Otherwise I agree from what I've read as well dynamic is better performing.

That said it's like you mentioned having both adds complexity! If you still feel that's not enough of a reason to keep static I'll remove it 👍

server/text_generation_server/utils/layers.py
Narsil: And so let's keep rewriting inv_freq. It has some indesirable effects on those models, but the other way is even worse.

iantbutler01: That sounds reasonable, I'll make this change after work.

M-Chris:
Would love to see this exposed now that the https://github.com/huggingface/transformers/pull/24653 merge is complete. 

I understand there are various complexities with flash attention, and if flash attention 2 will be implemented.. 
but maybe this can still follow the same implementation of transformers merge to reduce confusion and use the "rope_scaling={"type": "dynamic", "factor": 2.0}," args?  or at least a variation there of?

--rope_scaling {"type": "dynamic", "factor": 2.0} or however seen fit by you guys ?

🙌

gante:
@iantbutler01 @Narsil some information for decision-making 🤗 

> Overall, can we remove a lot of the complexity ?
From what I read, dynamic scaling seems just better than static scaling, so let's just use dynamic scaling, no ?

The current state of scaling techniques in `transformers`:
- dynamic scaling is the best for scaling without fine-tuning
- linear scaling is the best for scaling with fine-tuning

What's out there that I'm adding next in `transformers`:
- The folks behind dynamic scaling have created a technique that is the best in both regimes, before and after fine-tuning (see [this comment](https://github.com/huggingface/transformers/pull/24653#issuecomment-1633027893))

So... perhaps we can jump straight into the best technique in TGI? :D It should only need one flag in practice, the `scale`

iantbutler01:
@gante That sounds good to me! I'll work on this over the next few days.

M-Chris:
FWIW I tested against Llama via a simple overwrite on the Docker image on the latest TGI 0.9.3, using a flash attn v2 compatible GPU and works good 😄 

I pulled against this pr branch and layered it into the overwrite, with a few adjustments and assumptions.
- Assumed SCALE is the only ENV
- Assumed dynamic as the only option since this is only inference.
- Applied changes to FlashLlamaAttention since it rotary_emb was move from the original pr


_Couple thoughts//concerns.._
- The user will be required to set their --max-total-tokens, maybe something you have already a plan to address iantbutler01? also maybe its fine if a user manipulates these as seen fit.
- Not sure if there is a specific reason to be holding back the current version of transformers? (deployed that the docker as a test, doesn't seem to have any affect from testing )


Here is the docker used to quickly overwrite for testing if its helpful:
`
FROM ghcr.io/huggingface/text-generation-inference:latest

WORKDIR /usr/src

RUN apt-get update -y

RUN pip install transformers==4.31.0
RUN pip install scipy
RUN pip install bitsandbytes==0.40.2
RUN pip install --upgrade accelerate

COPY ./app_overwrites/server/text_generation_server/. /opt/conda/lib/python3.9/site-packages/text_generation_server
`

Id be happy to share anything else if needed 🍻

iantbutler01:
@gante Looking at @jquesnelle's repo, and the comment you linked to it looks like there is actually both a [standard by parts](https://github.com/jquesnelle/scaled-rope/pull/1/commits) as well as a [dynamic by parts](https://github.com/jquesnelle/scaled-rope/commit/b6aed01702e6e05ba4586b53b89b74d8fdf754c2) method. So it looks like the improvement you were talking about applies to both types of NTK aware scaling? In that case I'm inclined to make this PR just the dynamic by parts method to save on some of the complexity.

iantbutler01:
@Narsil @gante I spent some time tonight working to implement the dynamic parts by method I mentioned in my last comment. I'm coming to realize, that with this new method and the comment here: https://github.com/huggingface/text-generation-inference/issues/512#issuecomment-1648129166 suggesting there are now models that have been fine tuned with scaling that the complexity here is ballooning.

Even just the parts by method itself is more gnarly and requires supporting a whole bunch of parameters on the attention module.

Before I continue, at the risk of the complexity putting this in review hell, I'd like some guidance on what you all think I should proceed with. Personally if I add the dynamic parts by method linked in my previous comment it will have effectively set up the ability to implement the other methods here anyway, but maybe a follow up PR for those makes sense.

gante:
@iantbutler01 you raised good points: as users fine-tune their models with rope scaling, they may lose compatibility with TGI (depending on how we decide to do things here). And yes, let's settle on a path that avoids review hell!

I'd suggest separating the two use cases and making two separate decisions/PRs:
1. loading models with `rope_scaling`, such as the one linked [here](https://github.com/huggingface/text-generation-inference/issues/512#issuecomment-1648129166).
    - Would NOT be in this PR
    - Should consist in reading `model.config` and loading the right RoPE class. 
    - Support for each scaling strategy could be added progressively, depending on demand. For instance, all current relevant cases of fine-tuned RoPE scaling rely on linear scaling, so there is only 1 small modification needed in practice.
2. enabling dynamic RoPE scaling
    - Would be in this PR
    - Triggered by (e.g.) a simple env var
    - Implementing the NTK-by-parts, since it is the best-performing dynamic method so far
    
@Narsil @iantbutler01 WDYT?

iantbutler01:
@gante I am fine with that approach, that's basically what I started last night but I wanted to make sure that's what everyone had in mind.

iantbutler01:
Given the license change I am no longer comfortable contributing my work.

OliverFM:
@iantbutler01 would you be willing to contribute this change to a fork?
I was planning on adding support for Speculative Decoding, see the issue #729, but will no longer do that with the license changes. 

I am strongly considering maintaining a fork of the repo from the commit before the license change. I would be adding support for speculative decoding there.

PR #543, opened 2023-07-04
feat(server): use latest flash attention commit

OlivierDehaene:
@njhill FYI


server/pyproject.toml
```diff
@@ -26,7 +26,7 @@ hf-transfer = "^0.1.2"
 sentencepiece = "^0.1.97"
 tokenizers = "0.13.3"
 huggingface-hub = "^0.14.1"
-transformers = "^4.29.2"
+transformers = "4.29.2"
```
njhill, line 29: Why not transformers 4.30?

OlivierDehaene: It seems to have a max recursion issue with some llama tokenizers.

PR #557, opened 2023-07-06
fix(server): decrease memory fragmentation

OlivierDehaene:


ankit201:
@OlivierDehaene 
I was facing the same memory fragmentation issue (OOM error) while using MPT-30 with :

max_total_token = 1000
GPU - 2 x 80GB(A100)
I tried the same fix on causal_lm.py as you mentioned in flash_causal_lm.py by clearing the CUDA cache and performing a load test with 50 concurrent users. On watching the GPU usage, it tries to clear the cache since but doesn't do it effectively as the number of users increases and the model size still grows resulting in an OOM error, although this issue doesn't occur if I run a falcon-40b model and it effectively clears out the cache and model memory on GPU is managed effectively.

Is there a different fix for MPT-30 models which derives causal_lm.py?

PR #567, opened 2023-07-09
Memory Fragmentation for Causal LM

ankit201:
# Memory Fragmentation for Causal LM(eg: MPT-30B models for which Flash is still not implemented)

## Who can review?
@OlivierDehaene @Narsil


OlivierDehaene:
Hello
Thanks for the PR but we have a similar setup on main now.

PR #590, opened 2023-07-12
Bug fixes for GPTQ_BITS environment variable passthrough

ssmi153:
# What does this PR do?

This fixes a typo and extends the GPTP_BITS environment variables through to the second method which requires the same logic. Please let me know if there's anything I've misunderstood in this change.

Thanks @Narsil for the original fix.

### DIFF
--- a/server/text_generation_server/utils/weights.py
+++ b/server/text_generation_server/utils/weights.py
@@ -127,8 +127,8 @@ class Weights:
                 try:
                     import os
 
-                    bits = int(os.getenv("GTPQ_BITS"))
-                    groupsize = int(os.getenv("GTPQ_GROUPSIZE"))
+                    bits = int(os.getenv("GPTQ_BITS"))
+                    groupsize = int(os.getenv("GPTQ_GROUPSIZE"))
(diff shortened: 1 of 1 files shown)


Narsil (approved):
LGTM Thanks !

ssmi153:
@Narsil, just did some further testing with the new release and I'm still getting `RuntimeError: weight gptq_bits does not exist` errors. I think the issue here is that we call gettensor() (for the gptq_bits tensor), which in turn calls get_filename(), and this get_filename() function throws a RuntimeError if it can't find the file the tensor should be in. In our try / catch block we try to catch a SafetensorError rather than a RuntimeError, and so the RuntimeError sails right past the try/catch and gets thrown to global scope. I'm not sure exactly how you'd like to fix this. One option would be to catch both SafetensorErrors and RuntimeErrors.

ssmi153:
@Narsil, I've made the (very small) change and created a pull request with it here: https://github.com/huggingface/text-generation-inference/pull/596 . Wasn't sure how to continue to make changes to this pull request.

PR #600, opened 2023-07-13
Convert example docker command in README to use :latest rather than being pegged to :0.9

bealbrown:
# What does this PR do?

I had been encountering some issues with running a derivative of falcon over 4 GPU shards, and discovered that the example docker command I had been using from the README pegged the release version to 0.9.

Removing this limitation and switching to latest solved my issue (by the way, remarkable and great work making so many improvements on creating a general purpose environment for serving these models!)

Not sure what the rationale was behind pegging to 0.9 but if there's no remaining rationale I think this could certainly help people gain more value from this repo.

All the best!
bealbrown

Fixes # (issue)

## Before submitting
- [ x] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ model=bigscience/bloom-560m
 num_shard=2
 volume=$PWD/data # share a volume with the Docker container to avoid downloading weights every run
 
-docker run --gpus all --shm-size 1g -p 8080:80 -v $volume:/data ghcr.io/huggingface/text-generation-inference:0.9 --model-id $model --num-shard $num_shard
+docker run --gpus all --shm-size 1g -p 8080:80 -v $volume:/data ghcr.io/huggingface/text-generation-inference --model-id $model --num-shard $num_shard
 ```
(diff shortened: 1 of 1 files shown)


Narsil:
`:latest` is bad practice:

The reason is that you never know what `latest` means after a while. And we've had already a few issues where users where using `latest` and didn't pull so where not actually on `latest`. Lots of confusing back&forth ensued.
`docker` will not check latest if you don't ask it too.

You're free to use it still but I would recommend avoiding it (basically everywhere, unless you're just trying out something.).

bealbrown:
Ok! Will of course peg to a known working version for production usage, but for getting started demo seemed to me that using latest would be most appropriate. 

In my case 0.9 was unable to run a model with some obscure errors, so I almost abandoned using the project, but discovered that the latest version was able to. 

At any rate, thanks again for your great work on this project! All the best!

Narsil:
> In my case 0.9 was unable to run a model with some obscure errors,

We indeed may want to make sure this is up-to-date.

PR #605, opened 2023-07-13
Add optional cert param to python client - allows use of client cert based authentication

matth:
# What does this PR do?

For hosting `text-generation-inference` behind a server that uses SSL client certs for authentication, this PR adds an optional `cert` parameter to the python client libraries constructor. 

The format of this follows the underlying `requests` library and is passed onto `requests` when performing HTTP posts.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/clients/python/text_generation/client.py
+++ b/clients/python/text_generation/client.py
@@ -3,7 +3,7 @@ import requests
 
 from aiohttp import ClientSession, ClientTimeout
 from pydantic import ValidationError
-from typing import Dict, Optional, List, AsyncIterator, Iterator, Union
+from typing import Dict, Optional, List, AsyncIterator, Iterator, Union, Tuple
 
 from text_generation.types import (
     StreamResponse,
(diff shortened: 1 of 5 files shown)


clients/python/text_generation/client.py
```diff
@@ -41,6 +41,7 @@ def __init__(
         headers: Optional[Dict[str, str]] = None,
         cookies: Optional[Dict[str, str]] = None,
         timeout: int = 10,
+        cert: Optional[Union[str, tuple[str, str]]] = None,
```
ciaranby, line 44: Could you use Tuple from typing instead of tuple so it works with older version of python? This will cause an error on import with python <3.9.

matth:
Updated to add Tuple from typing, removed cert from async client

OlivierDehaene:
Thanks for the PR!
Is it possible to also add the cert to the AsycnClient, and the generate_stream methods? 
Also, is it possible to add a test to test the logic?

matth:
Thanks. 

Not sure there is a great way to test the cert functionality in the requests based client,  it's similar to the `timeout` and `cookies` parameters where it is just passed on to the `requests.post` call.  

To fully test would require a server that does the client cert authentication or mocking of the requests call - both of which introduce more complexity to the currently very straightforward tests.

We can add the cert to `generate_stream` and  the async client - it's a slightly different approach, there may be stuff to more easily test here.

Put PR back as draft while we do this.

Narsil:
Closing this as stale.

For reference we are soft deprecating this library's client in favor to use this one https://huggingface.co/docs/huggingface_hub/guides/inference which is now the main source of inference for all our tools/products.

PR #624, opened 2023-07-17
feat(server): flash attention v2

OlivierDehaene:


Narsil (dismissed):
LGTM

krzim:
It looks like this will remove compatibility with compute capability 7.5. That means T4 cards and AWS G4 instance types will no longer be supported.

Narsil (dismissed):

krzim:
That would be great. The motivation for my 4bit bnb PR was to improve throughout on Falcon models on g4s. We have workloads that run in regions that only have g4 availability so maintaining compatibility would be best.

OlivierDehaene:
I know. That's why the PR is not merged.

PR #630, opened 2023-07-18
feat(server): auto max_batch_total_tokens for flash att models

OlivierDehaene:


flozi00:
since this PR is merged I am running OOM all the time, even when the cli arg is set
using the docker container and 48GB GPU
7B model is the smallest i tested

OlivierDehaene:
Can you open a separate issue with the issue template?

flozi00:
found more details and opened issue

PR #684, opened 2023-07-23
Adding ctranslate2 quantization and inference

### LINKED ISSUE #496: Adding support for CTranslate2 acceleration
## Feature request

The feature would be to support accelerated inference with the CTranslate2 framework
https://github.com/OpenNMT/CTranslate2

## Motivation

Reasons to CTranslate2

#### faster float16 generation
In my case, outperforms VLLM #478 by a factor of 180% and Transformers by 190%.
This is measured for gpt_bigcode_starcoder, with input length of 10, batch_size of 2, max_new_tokens=32

#### even faster int8 quantization 
- int8 quantization that in most cases outperforms float16 inference, quality loss is minimal while being identical to 

#### Wide range of model support
This could be a new "best effort" option - https://opennmt.net/CTranslate2/guides/transformers.html

#### Stability
Compared to other frameworks, cross-platform unit tests.
It is the backend of LibreTranslate https://github.com/LibreTranslate/LibreTranslate

#### Streaming Support 
https://opennmt.net/CTranslate2/generation.html?highlight=stream

## Your contribution

Here is an example script, comparing the three frameworks with ```ray[serve]``` (replacement of the dynamic batching in TGI).

With the code below, my timings are:
In: `input: "def hello_world"` -> 32 new tokens
Median time of 10:

* Transformers : 287.6ms
*  VLLM : 278.8ms
* CTranslate2 : 193.5ms

```python

from fastapi import FastAPI
from ray import serve, get_gpu_ids
import time
from typing import List, Dict
from pydantic import BaseModel
from enum import Enum

app = FastAPI(docs_url="/docs") # lets not do any authentication for now 

class HFAPIRequest(BaseModel):
    inputs: str
    parameters: dict = {}

class Framework(Enum):
    ctranslate2 = 1
    transformers = 2
    vllm = 3


@serve.deployment
@serve.ingress(app)
class FastAPIWrapper:
    def __init__(self, model_handle) -> None:
        self.model_handle = model_handle
    
    @app.get("/v1/heartbeat")
    async def complete(self) -> float:
        """check that the application is running"""
        return time.time()
        
    @app.post("/v1/complete")
    async def complete(self, data: HFAPIRequest) -> Dict[str,str]:
        """send json-data per http request. returns a completion"""
        try:
            d_dict=data.dict()
            # send the call, get ray_ref as receipt
            ray_ref = await self.model_handle.predict.remote(inputs=d_dict["inputs"])
            # wait until ray_ref finished
            completion = await ray_ref
            return {"generated_text": completion}
        except Exception as ex:
            return  {"error": str(ex)}

@serve.deployment(ray_actor_options={"num_gpus": 1})
class CodeCompletionModel:
    def __init__(self, model_name = "michaelfeil/ct2fast-gpt_bigcode-santacoder", framework: Framework = Framework.ctranslate2 ) -> None:
        self.framework = framework
        gpus = bool(get_gpu_ids())
        
        if self.framework == Framework.ctranslate2:
            # 1784MiB VRAM, 314.3ms for generation
            from hf_hub_ctranslate2 import GeneratorCT2fromHfHub
            
            self.model = GeneratorCT2fromHfHub(
                    # use a pre-quantized model here
                    model_name_or_path=model_name, 
                    device="cuda" if gpus else "cpu",
                    compute_type="int8_float16" if gpus else "int8",
            )
        elif self.framework == Framework.transformers:
            # 2918MiB VRAM, ~1146.8ms for generation
            from transformers import pipeline
            import torch
            self.model = pipeline(
                task="text-generation",
                model="bigcode/gpt_bigcode-santacoder", 
                device="cuda:0" if gpus else "cpu",
                torch_dtype=torch.float16
            )
        elif self.framework == Framework.vllm:
            from vllm import LLM, SamplingParams
            self.sampling_params = SamplingParams(max_tokens=32)
            self.model  = LLM(model="bigcode/gpt_bigcode-santacoder")
        else:
            raise ValueError(
                f"{self.framework} is not a supported framework."
                f"Please use one of the following: {list(Framework)}"
            )
    
    
    @serve.batch(max_batch_size=4)
    async def predict(self, inputs: List[str]) -> List[str]:
        if self.framework == Framework.ctranslate2:
            return self.model.generate(
                text=inputs,
                max_length=32, 
                include_prompt_in_result=False
            )
        elif self.framework == Framework.transformers:
            out = self.model(inputs, max_new_tokens=32, return_full_text=True)
            return [o[0]["generated_text"] for o in out]
        elif self.framework == Framework.vllm:
            outputs = self.model.generate(inputs, self.sampling_params)
            return [o.outputs[0].text for o in outputs]
    
deploy_handle = FastAPIWrapper.bind(model_handle=CodeCompletionModel.bind(framework=Framework.ctranslate2))

if __name__ == "__main__":
    serve.run(deploy_handle, port=5000)
```

### LINKED ISSUE #680: server: Incorrect logit process for mixed CPU/GPU collection in StaticWarper - Logit process
### System Info

This is not exactly a bug, that should occur in production with the current code as it is.

### Information

- [ ] Docker
- [X] The CLI directly

### Tasks

- [ ] An officially supported command
- [X] My own modifications

## Reproduce:
Modify causal_lm.py:
- disable in __init__.py that model explicitly moved to CPU -> device cpu https://github.com/huggingface/text-generation-inference/blob/1da642bd0e6de28ef499f17cd226264f3ccdc824/server/text_generation_server/models/causal_lm.py#L477-L488
- run the santacoder_model unit tests, with a modified temperature or top_k to make sure that non-greedy sampling is selected.

## Observation
- Here the returned logits will be torch.Tensor(cpu, float32) instead of torch.Tensor(cuda, float16)  https://github.com/huggingface/text-generation-inference/blob/1da642bd0e6de28ef499f17cd226264f3ccdc824/server/text_generation_server/models/causal_lm.py#L532

- This causes a silent error here in next token selection: https://github.com/huggingface/text-generation-inference/blob/1da642bd0e6de28ef499f17cd226264f3ccdc824/server/text_generation_server/models/causal_lm.py#L576
The first generated token is repeated. I modified the logits to be esentially a "one-hot-tensor", setting the argmax() token to +10 and the others to -10, still getting a different token selected by the next_token_selection function.

I suspect the bug here.
https://github.com/huggingface/text-generation-inference/blob/1da642bd0e6de28ef499f17cd226264f3ccdc824/server/text_generation_server/utils/logits_process.py#L45

## Proposed changes
I would modify https://github.com/huggingface/text-generation-inference/blob/1da642bd0e6de28ef499f17cd226264f3ccdc824/server/text_generation_server/utils/logits_process.py#L45 to check if the scores tensor is not on cpu, instead of relying that the tensor is on gpu if gpu is available.

```python
# something like:
if torch.cuda.is_available() and scores.device != "cpu"
```

## Effect:
Because the code is relying on the fact, that all models are moved to GPU if cuda is available, the code works as is in production. Also the code works if no temperature sampling is used by greedy sampling instead.
However it would save significant debugging time (as for me 2-3h) 

### Expected behavior

The next_token selection should have no side-effects, and operate only on its configuration and the logits tensor.

michaelfeil:
# What does this PR do?
@OlivierDehaene - brace yourself!
This PR extends introduces a new quantization method [ctranslate2](https://github.com/OpenNMT/CTranslate2) (MIT-Licence), which is one of the remaining inference engines that are faster than plain PyTorch.
For comparison, ctranslate2 does everything in plain C++ and cuBLAS (chaining multiple forward passes in C++, next_token), making it up to 1.6x faster than [TGI with Custom Kernels](https://huggingface.co/michaelfeil/ct2fast-Llama-2-13b-chat-hf/discussions/1#64bb913e796f20daad439875) in Llama-2-13b-chat, for static (non-continuous) batch sizes, and also faster than [vllm](https://github.com/huggingface/text-generation-inference/issues/496)

### The following features are used:
- ctranslate2's [decoder-only forward  ](https://opennmt.net/CTranslate2/python/ctranslate2.Generator.html#ctranslate2.Generator.forward_batch)method working for all supported (decoder-only architectures)[https://opennmt.net/CTranslate2/guides/transformers.html]
- ctranslate2's quantization and native loading int8 / int8_float16 / float16 / bfloat16
- devices: cpu and cuda

### The following features are not available in the forward call:
- despite no KV caching running quite fast (no cache supported in Python API of CT2
- sharding. Implementing tensor parallel / model shared in CTranslate2 is still a open enchancement. This means that model needs to fit in memory once.

### My measured improvements:

- Plain GPU speedup (Codegen-2B) runs around **325% faster** (measured time per token) while taking only 50% of the cuda memory. This is valid for models without special kernel / best effort /  `AutoModel` with  `fp16` .
- lower latency: 
       - First token latency: Due to the fact, that are no KV caches, especially short generations (max_new_token) are fast, e.g. 
         first token. 
       - Startup latency: Converted model weights are limited by SSD speed. My case: GPU **10x** and CPU **4x** faster startup times.
- CPU-speedup:
         -  support of native int8 e.g. with intel MKL -> memory savings
         -  should be a faster (20-80%) compared to current torch implementations.
- Wide compatibility: ctranslate2 has cross platform tests, and works across multiple Nvidia architectures (almost everything that supports cuda 11 or mkl)

# How is this done:
I duplicated the causal_lm, and overwrote some parts of forward and init.

# Summary:
This quantization and inference method is if you have one of the following:
- inference on one GPU + no custom kernel available (not yet added or kernel compatibility with gpu)
- inference on CPU 
- balance between low-memory usage and fast inference 
- low latency: server startup or first token latency

Fixes #496 
Fixes  #680 

## Disclaimers
- installs a new dependency with ctranslate2 (MIT) - itself it only bundles pybind11
- I don't have a option to build the docker image on my laptop

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@OlivierDehaene - let me know, if there are any proposed changes before merge

### DIFF
--- a/server/text_generation_server/cli.py
+++ b/server/text_generation_server/cli.py
@@ -89,6 +89,7 @@ def download_weights(
     auto_convert: bool = True,
     logger_level: str = "INFO",
     json_output: bool = False,
+    trust_remote_code: bool = False
 ):
     # Remove default handler
     logger.remove()
@@ -168,6 +169,7 @@ def download_weights(
             config = AutoConfig.from_pretrained(
                 model_id,
                 revision=revision,
+                trust_remote_code=trust_remote_code
             )
             architecture = config.architectures[0]
 
--- a/server/text_generation_server/cli.py
+++ b/server/text_generation_server/cli.py
@@ -14,6 +14,7 @@ app = typer.Typer()
 class Quantization(str, Enum):
     bitsandbytes = "bitsandbytes"
     gptq = "gptq"
+    ct2 = "ct2"
 
 
 class Dtype(str, Enum):
@@ -71,7 +72,7 @@ def serve(
     # Downgrade enum into str for easier management later on
     quantize = None if quantize is None else quantize.value
     dtype = None if dtype is None else dtype.value
-    if dtype is not None and quantize is not None:
+    if dtype is not None and quantize is not None and quantize != Quantization.ct2:
         raise RuntimeError(
             "Only 1 can be set between `dtype` and `quantize`, as they both decide how goes the final model."
         )
--- a/server/text_generation_server/models/__init__.py
+++ b/server/text_generation_server/models/__init__.py
@@ -18,6 +18,7 @@ from text_generation_server.models.galactica import GalacticaSharded
 from text_generation_server.models.santacoder import SantaCoder
 from text_generation_server.models.t5 import T5Sharded
 from text_generation_server.models.gpt_neox import GPTNeoxSharded
+from text_generation_server.models.ct2_causal_lm import CT2CausalLM
 
 # The flag below controls whether to allow TF32 on matmul. This flag defaults to False
 # in PyTorch 1.12 and later.
@@ -74,6 +75,7 @@ def get_model(
     dtype: Optional[str],
     trust_remote_code: bool,
 ) -> Model:
+    dtype_ct2 = dtype
     if dtype is None:
         dtype = torch.float16
     elif dtype == "float16":
@@ -82,6 +84,15 @@ def get_model(
         dtype = torch.bfloat16
     else:
         raise RuntimeError(f"Unknown dtype {dtype}")
+
+    if quantize is not None and "ct2" in quantize:
+        return CT2CausalLM(
+            model_id,
+            revision,
+            quantize=quantize,
+            dtype=dtype_ct2,
+            trust_remote_code=trust_remote_code,
+        )
 
     if "facebook/galactica" in model_id:
         return GalacticaSharded(
--- a/server/text_generation_server/models/ct2_causal_lm.py
+++ b/server/text_generation_server/models/ct2_causal_lm.py
@@ -18,6 +18,7 @@
 import torch
 import numpy as np
 import os
+import multiprocessing
 from pathlib import Path
 
 from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
@@ -71,10 +72,12 @@ class CT2CausalLM(Model):
         )
 
         # Start CT2
+        ct2_generator_kwargs = {"inter_threads": 1}
         if torch.cuda.is_available():
             self.ct2_device = "cuda"
         else:
             self.ct2_device = "cpu"
+            ct2_generator_kwargs["intra_threads"] = multiprocessing.cpu_count() // 2
 
         if dtype == torch.float16 and self.ct2_device == "cuda":
             ct2_compute_type = "float16"
@@ -127,7 +130,8 @@ class CT2CausalLM(Model):
 
         # Start CT2
         self.ct2_model = ctranslate2.Generator(
-            str(out_dir), device=self.ct2_device, compute_type=ct2_compute_type
+            str(out_dir), device=self.ct2_device, compute_type=ct2_compute_type,
+            **ct2_generator_kwargs
         )
 
         class DummyModel(torch.nn.Module):
@@ -210,7 +214,7 @@ class CT2CausalLM(Model):
             .flatten(1)
             .to(torch.int32)
         )
-        # lengths of the padded ids_input, i.e. how often 1234567 is used.
+        # lengths of the padded ids_input, i.e. how often not pad=1234567 is used.
         lengths = np.array(input_lengths, dtype=np.int32)
         
         if self.ct2_device == "cuda":
--- a/server/text_generation_server/models/ct2_causal_lm.py
+++ b/server/text_generation_server/models/ct2_causal_lm.py
@@ -81,14 +81,15 @@ class CT2CausalLM(Model):
         elif dtype == torch.float16:
             ct2_compute_type = "bfloat16"
         else:
-            # default, int8 quantization.
+            # default, int8 quantization. 
+            # Fastest and lowest
             if "cuda" in self.ct2_device:
                 ct2_compute_type = "int8_float16"
             else:
                 ct2_compute_type = "int8"
-                # raise ValueError("cpu is currently experimental due to"
-                #                  " sampling based / non-greedy next_token"
-                #                  " of code only working in float16.")
+                # print("cpu is currently experimental due to"
+                #       " sampling based / non-greedy next_token"
+                #       " of code only working in float16.")
         # Start CT2 - conversion
         out_dir = (
             Path(HUGGINGFACE_HUB_CACHE)
@@ -103,7 +104,7 @@ class CT2CausalLM(Model):
                     load_as_float16=True,
                     revision=None,
                     low_cpu_mem_usage=True,
-                    trust_remote_code=True,
+                    trust_remote_code=trust_remote_code,
                 )
                 converter.convert(
                     output_dir=out_dir,
@@ -200,12 +201,12 @@ class CT2CausalLM(Model):
         # CT2 forward requires a list of list of input tokens ids and lengths
         ids_input = (
             torch.nested.to_padded_tensor(
-                torch.nested.nested_tensor(all_input_ids), 1234
+                torch.nested.nested_tensor(all_input_ids), 1234567
             )
             .flatten(1)
             .to(torch.int32)
         )
-        # lengths of the padded ids_input, i.e. how often 1234 is used.
+        # lengths of the padded ids_input, i.e. how often 1234567 is used.
         lengths = torch.from_numpy(np.array(input_lengths, dtype=np.int32)).to(
             ids_input.device
         )
--- a/server/text_generation_server/models/ct2_causal_lm.py
+++ b/server/text_generation_server/models/ct2_causal_lm.py
@@ -2,7 +2,9 @@ import torch
 import inspect
 import numpy as np
 import os
+from pathlib import Path
 
+from huggingface_hub.constants import HUGGINGFACE_HUB_CACHE
 from dataclasses import dataclass
 from opentelemetry import trace
 from transformers import (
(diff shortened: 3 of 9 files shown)


server/text_generation_server/cli.py
```diff
@@ -71,7 +72,7 @@ def serve(
     # Downgrade enum into str for easier management later on
     quantize = None if quantize is None else quantize.value
     dtype = None if dtype is None else dtype.value
-    if dtype is not None and quantize is not None:
+    if dtype is not None and quantize is not None and quantize != Quantization.ct2:
```
michaelfeil, line 75: Should I define float16 / bfloat / int8 over dtype -- or -- make them three items in the Quanitzation Enum?

server/text_generation_server/models/ct2_causal_lm.py
```diff
@@ -0,0 +1,381 @@
...
+    def decode(self, generated_ids: List[int]) -> str:
+        return self.tokenizer.decode(
+            generated_ids, skip_special_tokens=True, clean_up_tokenization_spaces=False
+        )
+
+    # def forward_slowtokenize_ct2(
+    #     self, all_input_ids,
+    # ) -> Tuple[torch.Tensor, List[Tuple[torch.Tensor, torch.Tensor]]]:
+    #     # Model Forward, by copy between cpu and cuda
+    #     tokens_in  = [self.tokenizer.convert_ids_to_tokens(i) for i in all_input_ids]
```
michaelfeil, line 169: This code also works quite fast, however moving the tokens from cuda -> cpu -> cuda. On the other hand does not require padding.

server/text_generation_server/models/ct2_causal_lm.py
```diff
@@ -0,0 +1,381 @@
...
+        # if sum := torch.isnan(logits).sum():
+        #     sum2 = torch.isnan(logits2).sum()
+        #     raise ValueError(f"logits {sum}, {sum2}")
+        # if sum2 := torch.isnan(logits2).sum():
+        #     raise ValueError(f"logits2 {sum}")
+        # torch.testing.assert_close(logits, logits2)
+        # raise ValueError(f"all_input_ids={len(batch.all_input_ids)},{batch.all_input_ids[0].shape}, logits={logits.shape}, tokens_in={len(tokens_in)},{len(tokens_in[0])}")
+
+        # Results
+        generations: List[Generation] = []
```
michaelfeil, line 251: Code below should be identical to causal_lm.py

server/text_generation_server/models/__init__.py
```diff
@@ -74,6 +75,7 @@ def get_model(
     dtype: Optional[str],
     trust_remote_code: bool,
 ) -> Model:
+    dtype_ct2 = dtype
```
michaelfeil, line 78: Fix to not default to float16 for a "dtype=None" arg.

OlivierDehaene:
Hello
I will look into it. 
Any benchmarks you can share on this PR vs og TGI?

The benchmark you ran in hf.co is biased as for TGI you used: 

```
GenerateParameters { best_of: Some(1), temperature: Some(0.5),
 repetition_penalty: Some(1.03), top_k: Some(10), top_p: Some(0.95), typical_p: Some(0.95), do_sample: true, max_new_tokens: 128,
 return_full_text: Some(false), stop: ["photographer"], truncate: None, watermark: true,
 details: true, decoder_input_details: true, seed: None }
 ```
 
 Sampling is always slower. Also, you used stop sequences, watermarking (very slow) and details + decoder_input_details  (also very slow but only for the prefill step).

File: server/text_generation_server/models/ct2_causal_lm.py [hunk 5]
```diff
@@ -0,0 +1,394 @@
...
+            trust_remote_code=trust_remote_code,
+        )
+
+        # Start CT2
+        ct2_generator_kwargs = {"inter_threads": 1}
+        if torch.cuda.is_available():
+            self.ct2_device = "cuda"
+        else:
+            self.ct2_device = "cpu"
+            ct2_generator_kwargs["intra_threads"] = multiprocessing.cpu_count() // 2
```
Yard1, line 80: Would be great if this could be also set by an env var.

michaelfeil: I would suggest to default to this value (default is 0, whereas taking the physical cores has ~50-70% faster inference speed)

However I like the idea of being able to set this an env variable.

michaelfeil: Updated now, if you are happy with the solution, can you resolve this comment.

Yard1: LGTM, though you'll probably need to convert the string obtained from os.environ into an int!

michaelfeil:
@OlivierDehaene @Narsil You have done a great job developing TGI until now, its a great piece of technology.

Apart from the code itself, I am quite disappointed about the licence change, and in particular about the prior communication (i did read about it first on social media). I hope internal monetization pressure will not do similar things to related huggingface repos that works in close cooperation with customers, e.g. accelerate.

This PR is licenced Apache 2.0 . I have worked on it in my free-time. I really believe in the open source mission, and will contribute it to forked projects, e.g. https://github.com/Preemo-Inc/text-generation-inference. In effect, you can expect no more commits from my side here.

Narsil:
>  In effect, you can expect no more commits from my side here.

No worries, it's OK. 
I think we didn't necessarily do a great job at communicating the license change.
@OlivierDehaene did send prior mails to all previous contributors to notify of the change, sorry you didn't get one. We used the github ones we found.

Also I'd like to reaffirm that the change only applies to Hosted services that provide cloud like services, not to people using it, even as part as products like let's say someone using it with his own personal fine-tuned model specifically for RPG adventure generation let's say.
Not everyone seems to have understood that.

I'm avoid all the grey zones on purpose in that example.

https://twitter.com/ClementDelangue/status/1685286529147502592

PR #725, opened 2023-07-28
chore: update license to HFOIL

OlivierDehaene:


OlivierDehaene:
See #726

julien-c:
reverted in #1714

PR #727, opened 2023-07-28
v1.0.0

OlivierDehaene:


njhill:
@OlivierDehaene curious about the reason for reverting the vllm update as part of this?

OlivierDehaene:
Initial tests were fine but LLama v2 70b was not happy about it.

njhill:
Thanks for the explanation

PR #789, opened 2023-08-07
Add docker-compose example

mallorbc:
# What does this PR do?

The repo has an example running this with plain docker, but it's likely to want to run this with other services.  Using docker-compose can help with that and this example shows how to do it


mallorbc:
Wasn't aware of the license change until recently.  This PR is Apache 2.0 and I will not contribute under the custom license you added a week ago.  I will make any related PRs [here](https://github.com/Preemo-Inc/text-generation-inference)

PR #617, opened 2023-07-14
Add WIP support for returning top tokens

Vinno97:
# What does this PR do?

This PR add support for return the log probabilities of the top N tokens (besides the chosen model).

The token selection code is loaned from the [IBM's fork](https://github.com/IBM/text-generation-inference/) (much thanks @njhill) and I merged into the current `main` branch. I chose to name the input parameter `top_n_tokens` and it can be any (positive) integer. If it's set, a `top_tokens` field in the result `details` will contain a LxN list of tokens where L is the sequence length and N is the amount of requested tokens.

Note that it is currently very much WIP.  

* It is currently only implemented for seq-to-seq models (though it's modular and should be easy to implement for other models as well). 
* I haven't implemented this functionality for the prefill. **Do we want to?**
* I sort the tokens based on their `logprobs`, because IBM's implementation had the functionality. However, it may cause unnecessary overhead. 
* IBM's code doesn't just return the "top n" tokens. It actually (explicitly) returns more tokens when there are tokens with equal probabilities. I am not sure if this is desired.
* I have not yet implemented input validation of the `top_n_tokens` field.
* I have also done no benchmarking to test the impact on performance. 
  * I haven't given the benchmarking tool the capability to benchmark this feature.

Fixes #604 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
- [x] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ ] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. 
Specifically @OlivierDehaene and @njhill

Issue #505, opened 2023-06-28
Guidance acceleration


Atry:
### Feature request

[Guidance](https://github.com/microsoft/guidance) can control the generated format, which could be a nice feature if it is built-in

- Add extra parameter to `/generate` and `/generate_stream` protocol to specify a template
- Use guidance-like mask to generate tokens and pre-fill known tokens, alternately

### Motivation

See https://github.com/microsoft/guidance#guidance-acceleration-notebook 

### Your contribution

Just an idea for now

Vinno97:
For now, won't it be better to add text-generation-inference as a supported LLM into Guidance, in the same way as it did for OpenAI's models?

Atry:
Similar feature in another text generation project: https://github.com/ggerganov/llama.cpp/pull/1773

Atry:
See also https://github.com/go-skynet/LocalAI/issues/354

yubuyuabc:
> 



> Similar feature in another text generation project: [ggerganov/llama.cpp#1773](https://github.com/ggerganov/llama.cpp/pull/1773)

I wonder konw if it can or not get a stream with batch process

OlivierDehaene:
This is one of our priority for the next release.

KreshLaDoge:
Do you happen to have any updates on this matter?
Coming from [LMQL](https://github.com/eth-sri/lmql/issues/190)

PR #810, opened 2023-08-10
Added support for `logit_bias` in token generation [WIP / Looking for Feedback]

### LINKED ISSUE #240: [Feature] support for general logit processors
Are there any plans to support logit processors via an additional API parameters? For instance, the OpenAI API provides a `logit_bias` parameter, that is applied to the token distribution during generation.

This would be great for http://github.com/eth-sri/lmql, where we heavily rely on token masking to enforce constraints during decoding.

Thank you for your hard work on this project.

MarcusDunn:
# What does this PR do?

Adds support for sending `logit_bias`'s during token generation using [SequenceBiasLogitsProcessor](https://huggingface.co/docs/transformers/main/en/internal/generation_utils#transformers.SequenceBiasLogitsProcessor).

Fixes #240 

## Before submitting
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests) Pull Request section?
- [ ] Did you make sure to update the documentation with your changes.

## Tasks still remaining.

- [ ] Fix `FlashCausalLMBatch.concatenate`

   As we need to tokenize the inputs to `logit_bias` in order to create a `SequenceBiasLogitsProcessor` there needs to be a  tokenizer passed into `NextTokenChooser.from_pb` and `HeterogeneousNextTokenChooser.from_pb` this was no issue everywhere but `FlashCausalLMBatch.concatenate`. 

    I'm looking for feedback on how to resolve this nicely - the obvious solution would be to store a `tokenizer` in `FlashCausalLMBatch` or `HeterogeneousNextTokenChooser` similar to how `device` is selected when concatenating at the moment.

- [ ] Add Python tests

- [ ] Update client

- [ ] Documentation

    Not 100% if this is complete. I've updated the rust openapi schema and added docs to the protobuf files - not sure if this is sufficient.

## Who can review?

@OlivierDehaene

### DIFF
--- a/server/text_generation_server/models/flash_causal_lm.py
+++ b/server/text_generation_server/models/flash_causal_lm.py
@@ -303,7 +303,7 @@ class FlashCausalLMBatch(Batch):
             max_length = max(max_length, input_length + max_new_tokens)
 
         next_token_chooser = HeterogeneousNextTokenChooser.from_pb(
-            next_token_chooser_parameters, dtype, device
+            next_token_chooser_parameters, dtype, device, tokenizer
         )
         start_slots = torch.tensor(start_slots, dtype=torch.int64)
 
@@ -634,6 +634,8 @@ class FlashCausalLMBatch(Batch):
             next_token_chooser_parameters,
             dtype=batches[0].next_token_chooser.dtype,
             device=batches[0].next_token_chooser.device,
+            # todo - determine how to obtain access to a tokenizer here
+            tokenizer=...
         )
 
         # Needed to avoid dropping blocks when the batches will go out of scope
--- a/server/text_generation_server/utils/tokens.py
+++ b/server/text_generation_server/utils/tokens.py
@@ -33,7 +33,7 @@ class NextTokenChooser:
             do_sample=False,
             seed=0,
             device="cpu",
-            logit_bias=None,
+            logit_bias={},
     ):
         self.watermark_processor = (
             WatermarkLogitsProcessor(device=device) if watermark else None
@@ -85,6 +85,7 @@ class NextTokenChooser:
             cls,
             pb: generate_pb2.NextTokenChooserParameters,
             device: torch.device,
+            tokenizer: PreTrainedTokenizerBase,
     ) -> "NextTokenChooser":
         return NextTokenChooser(
             watermark=pb.watermark,
@@ -96,7 +97,7 @@ class NextTokenChooser:
             do_sample=pb.do_sample,
             seed=pb.seed,
             device=device,
-            logit_bias=pb.logit_bias,
+            logit_bias=dict([(tuple(tokenizer.encode(bias.string, add_special_tokens=False).input_ids[0]), bias.bias) for  bias in pb.logit_bias]),
         )
 
 
@@ -285,6 +286,7 @@ bias
             pb: List[generate_pb2.NextTokenChooserParameters],
             dtype: torch.dtype,
             device: torch.device,
+            tokenizer: PreTrainedTokenizerBase,
     ) -> "HeterogeneousNextTokenChooser":
         return HeterogeneousNextTokenChooser(
             watermark=[pb_.watermark for pb_ in pb],
@@ -297,7 +299,7 @@ bias
             seeds=[pb_.seed for pb_ in pb],
             device=device,
             dtype=dtype,
-            logit_bias={},
+            logit_bias=[dict([(tuple(tokenizer.encode(bias.string, add_special_tokens=False).input_ids[0]), bias.bias) for bias in pb_.logit_bias])  for pb_ in pb],
         )
 
 
--- a/server/text_generation_server/utils/tokens.py
+++ b/server/text_generation_server/utils/tokens.py
@@ -23,17 +23,17 @@ from text_generation_server.utils.logits_process import (
 
 class NextTokenChooser:
     def __init__(
-            self,
-            watermark=False,
-            temperature=1.0,
-            repetition_penalty=1.0,
-            top_k=None,
-            top_p=None,
-            typical_p=None,
-            do_sample=False,
-            seed=0,
-            device="cpu",
-            logit_bias={},
+        self,
+        watermark=False,
+        temperature=1.0,
+        repetition_penalty=1.0,
+        top_k=None,
+        top_p=None,
+        typical_p=None,
+        do_sample=False,
+        seed=0,
+        device="cpu",
+        logit_bias={},
     ):
         self.watermark_processor = (
             WatermarkLogitsProcessor(device=device) if watermark else None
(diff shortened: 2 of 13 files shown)


server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -634,6 +634,7 @@ def concatenate(cls, batches: List["FlashCausalLMBatch"]) -> "FlashCausalLMBatch
             next_token_chooser_parameters,
             dtype=batches[0].next_token_chooser.dtype,
             device=batches[0].next_token_chooser.device,
+            tokenizer=batches[0].next_token_chooser.tokenizer
```
MarcusDunn, line 637: *here* is where I need access to a tokenizer but one was not conveniently already available in scope.

server/text_generation_server/utils/tokens.py
```diff
@@ -218,12 +241,15 @@ def __init__(
         self.do_sample = do_sample
         self.dtype = dtype
         self.device = device
+        self.tokenizer = tokenizer
```
MarcusDunn, line 244: so unfortunately I had to add a field to `HeterogeneousNextTokenChooser`

server/text_generation_server/utils/tokens.py
```diff
@@ -275,8 +302,12 @@ def from_pb(
             typical_p=[pb_.typical_p for pb_ in pb],
             do_sample=[pb_.do_sample for pb_ in pb],
             seeds=[pb_.seed for pb_ in pb],
+            logit_bias=[dict(
+                [(tuple(tokenizer.encode(bias.string, add_special_tokens=False).input_ids[0]), bias.bias) for bias in
+                 pb_.logit_bias]) for pb_ in pb],
```
MarcusDunn, line 307: my python is pretty bad - this felt really ugly, not certain how to improve it.

MarcusDunn:
@Narsil this is probably ready for review (I understand @OlivierDehaene is off for August?).

It would be great if you could approve workflows, I cannot build the docker image for the life of me and cannot test CUDA stuff without it.

Narsil:
Hi thanks for the PR.

This will likely wait for @OlivierDehaene return.
There are a lot of ways to push models in certain directions (guidance, token healing, speculative decoding, logits bias).

I'm relatively convinced, that modifying biases like you're suggesting is not desirable. This is something that should either be in the model itself, or access higher level of controls.

That being said, all methods mentioned above will likely need to find a common underlying arch (so we don't have to recreate everything all the time.)

Efficiency is also an important factor, adding new logits modifiers adds overhead for all clients on the machine.
Check how any modification can impact performance: https://github.com/huggingface/text-generation-inference/pull/617

MarcusDunn:
We sometimes call out to a LLM to ask a true/false question (or select from a series of dynamically generated options) to make some decision. The context can get fairly long and it takes some time to prompt engineer and test our way into getting an expected answer consistently.

This prompt engineering process is repeated every time we swap out a model. (Falcon -> LLama 2 chat -> LLama 2 instruction-finetuned), and we're never 100% confident that our prompt is sufficient to guarantee the output format we want. It slows down testing the latest and greatest from the open source community.

I've seen guidance/lmql, it would 100% do what we need, but seems like gross overkill and I do not want to maintain a HTTP interface in front of it if I can avoid it. 

I would love a way to guide output, I can see that being totally out of scope for this project though, this felt like a reasonable medium.

Narsil:
> what do you mean by "access higher level of controls"?

Typically guidance. You would specific something higher level that raw token_id (which most user do not have access to).
Also because specifying ids require specific knowledge of the model, which is something most users do not have. The prompt is already quite tricky for finetuned models.

ggbetz:
Hello! @narsil wrote:

> Could you share maybe more information about the problem you're trying to solve ?

I'm working with [LMQL](https://github.com/eth-sri/lmql) and if I understand correctly, `logit_bias` support would be essential for using TGI - HF Inference Endpoints as an open cloud alternative to OpenAI. (See issue https://github.com/eth-sri/lmql/issues/190.)

Narsil:
I read a bit lmql, seems to be doing pretty much what I would like to avoid, knowing the model internals (here the tokenizer) to control the output via input_id biases.
Even worse, you usually don't want the same biases depending on where you are in the generation.

This seems a bit higher level and closer to something which would make sense IMO: https://github.com/guidance-ai/guidance
I don't think we want to expose anything with regard to model internals. It makes things quite hard to use.

Still a lot of work when we're not working with a single model, and a bunch, which use different tokenizers with different properties. But the higher level the control, the easier it is for us to maintain ideally.

MarcusDunn:
Just to clarify - the PR as it currently is takes in strings to bias towards, not tokens. The caller does not need to know token_ids and can pass in strings larger than one token.

MarcusDunn:
Just to clarify - the PR as it currently is takes in strings to bias towards, not tokens. The caller does not need to know token_ids and can pass in strings larger than one token.

Is something like guidance within the scope for tgi?

Where could I find the roadmap you mentioned?

charles-dyfis-net:
> I read a bit lmql, seems to be doing pretty much what I would like to avoid, knowing the model internals (here the tokenizer) to control the output via input_id biases.

That's not a burden LMQL puts on its users, but one it takes on itself -- much as Guidance does.

> Even worse, you usually don't want the same biases depending on where you are in the generation.

Correct, and that's why LMQL breaks down an operation into smaller requests to the model, again, as Guidance does.

Narsil:
@charles-dyfis-net . But that means stopping and restarting queries all the time. Meaning forgetting KV all the time pretty much.

This seems like a very high cost latency wise.
Keeping KV around hoping for client to reconnect is also not great. (How long do you keep it around, how do you cache hit, etc..).

charles-dyfis-net:
> @charles-dyfis-net . But that means stopping and restarting queries all the time. Meaning forgetting KV all the time pretty much.
> 
> This seems like a very high cost latency wise. Keeping KV around hoping for client to reconnect is also not great. (How long do you keep it around, how do you cache hit, etc..).

So when you describe guidance as something that you want to support, are you saying you'd want to run it server-side, built into TGI? If you run Guidance server-side... well, you could run LMQL queries server-side as well, though that's going to leave a lot of users (like ones who want to implement tool-calling workflows) dissatisfied. And unless y'all have a route to Microsoft's team that isn't open to the public, I'd worry about maintainer accessibility; one of the significant factors behind my decision to use LMQL over Guidance is the availability of a highly responsive and accessible upstream.

One way to split the difference is to divide into multiple operations whenever tool use or something else that needs to happen on the remote end takes place; LMQL upstream has been toying with the idea of supporting that explicitly.

With respect to the KV cache -- personally, I'd bind it to a specific [LMTP](https://github.com/eth-sri/lmql/blob/main/src/lmql/models/lmtp/README.md)-over-websocket connection, since a given LMQL client will be reusing a single transport over time in nontrivial cases. (The interactive debugger/playground creates a new transport instance per call, but it's not typical of more serious applications in that respect).

MarcusDunn:
Looks like we'll be getting one better than this PR.
 https://github.com/huggingface/text-generation-inference/issues/505#issuecomment-1708367609

I'll close for now.

PR #848, opened 2023-08-15
Update README.md

adarshxs:
@Narsil

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,10 +94,6 @@ To see all options to serve your models (in the [code](https://github.com/huggin
 ```
 text-generation-launcher --help
 ```
-For a basic run, you can start the server as follows:
-```shell
-text-generation-launcher --model-id tiiuae/falcon-7b-instruct --num-shard 1 # You can change the model-id accordingly
-```
 
 You can then query the model using either the `/generate` or `/generate_stream` routes:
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -90,10 +90,14 @@ docker run --gpus all --shm-size 1g -p 8080:80 -v $volume:/data ghcr.io/huggingf
...
 
-To see all options to serve your models (in the [code](https://github.com/huggingface/text-generation-inference/blob/main/launcher/src/main.rs) or in the cli:
+To see all options to serve your models (in the [code](https://github.com/huggingface/text-generation-inference/blob/main/launcher/src/main.rs) or in the cli):
 ```
 text-generation-launcher --help
 ```
+For a basic run, you can start the server as follows:
+```shell
+text-generation-launcher --model-id tiiuae/falcon-7b-instruct --num-shard 1 # You can change the model-id accordingly
+```
```
Narsil, line 100:
```suggestion
```

We're trying not to push too much the local CLI. it has many more caveats in dependencies and hardware.

Narsil (commented):
Thanks for the typo fix.

I think we should remove the local CLI example, we're really trying to push the docker CLI instead (which avoids a lot of headaches for dependencies for users).

Narsil (commented):
Thanks for the typo fix.

I think we should remove the local CLI example, we're really trying to push the docker CLI instead (which avoids a lot of headaches for dependencies for users).

adarshxs:
That's cool

Narsil (approved):
Thanks

PR #852, opened 2023-08-15
Misc minor improvements for InferenceClient docs

osanseviero:
List of changes

- Add mention that it just works with inference api
- No need to specify `model` in `text_generation` if it's already specified in `InferenceClient`
- I separated the explanation of `stream=True` and `details=True`
- I found the details explanation a bit repetitive (it says two times what it returns), so removed a sentence
- Add mention of async client


Wauplin (approved):
LGTM! Thanks for clarifying the docs :)

merveenoyan (approved):
Thanks!

PR #884, opened 2023-08-19
Remove redundant content from stream guide

osanseviero:


docs/source/conceptual/streaming.md
```diff
@@ -143,6 +143,4 @@ SSEs are different than:
...
-
-One of the limitations of Server-Sent Events is that they limit how many concurrent requests can handle by the server. Instead of timing out when there are too many SSE connections, TGI returns an HTTP Error with an `overloaded` error type (`huggingface_hub` returns `OverloadedError`). This allows the client to manage the overloaded server (e.g., it could display a busy error to the user or retry with a new request). To configure the maximum number of concurrent requests, you can specify `--max_concurrent_requests`, allowing clients to handle backpressure.
+One of the limitations of Server-Sent Events is that they limit how many concurrent requests can handled by the server. Instead of timing out when there are too many SSE connections, TGI returns an HTTP Error with an `overloaded` error type (`huggingface_hub` returns `OverloadedError`). This allows the client to manage the overloaded server (e.g., it could display a busy error to the user or retry with a new request). To configure the maximum number of concurrent requests, you can specify `--max_concurrent_requests`, allowing clients to handle backpressure.
```
Narsil, line 146: Reading this again, I don't see what's the link with SSE.... 
concurrent requests would work the same without SSE. 

It's purely backpressure and because typical queries of ML systems are quite long lasting compared to standard web timings. (Backpressure also has to be dealt with on standard web servers, but the volumes are typically much higher)

osanseviero: I updated the content. WDYT?

For context, when reading about SSEs vs webhooks, a frequently mentioned con for SSEs is that they have a limit on how many concurrent connections they can have. By mentioning that we handle the number of concurrent requests to handle backpressure, I was trying to build a connection that this would not be an issue. I hope this new structure is clearer!

PR #893, opened 2023-08-21
Add ignore_eos_token

chiragjn:


Narsil:
Can you provide more background on this ?
What is the original problem ? What are you trying to achieve ?
Can you give real examples of the behavior you expect ?

chiragjn:
Oh my bad, I didn't realize I opened a PR against huggingface origin 🙈 
I don't really want this merged, but anyway, I noticed grpc API takes in ignore_eos_token, I wanted to run some benchmarking via the HTTP rest interface instead of the rust utility

I'll close the PR

PR #898, opened 2023-08-22
Expose `ignore_eos_token` to client

caiyesd:
# What does this PR do?

`ignore_eos_token` is only supported by benchmark tool, but cannot be used in client library.
This PR modifed the `router` to let `ignore_eos_token` can be set in the client.
The PR also add `ignore_eos_token` to the client APIs, and keep it be false by default for compatibility.

### DIFF
--- a/clients/python/text_generation/client.py
+++ b/clients/python/text_generation/client.py
@@ -75,6 +75,7 @@ class Client:
         typical_p: Optional[float] = None,
         watermark: bool = False,
         decoder_input_details: bool = False,
+        ignore_eos_token: bool = False,
     ) -> Response:
         """
         Given a prompt, generate the following text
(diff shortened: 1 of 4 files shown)


Narsil:
Do you mind trying to explain a little more background on this ?

Why do you need it ? What is the problem you are having without it ?

caiyesd:
I'm doing some work on inference performance test. And I developed a stress test tool based on TGI client. But I found that when the batch size goes greater, less tokens will be generated. The generated tokens' count cannot always reach the `max_new_tokens` of the request. 

My requirement is to control the number of generated tokens, but `max_new_tokens` in parameters can only be used to limit the max number of generated tokens. 

I also investigated vLLM which supports this feature. 
```
`ignore_eos`
Whether to ignore the EOS token and continue generating tokens after the EOS token is generated.
```

I see `text-generation-bentchmark` tool of TGI uses `ignore_eos_token` and set it to `True` during the benchmark test. But it do the benchmark test without router, it test with server directly. So I want to expose this parameter to the client, and then we can use client to do benchmark test with router and server.

Narsil:
When benchmarking, you could use `details: True` and recover the number of tokens from there. So you don't have to guess the amount of tokens and you're still using more natural load (some requests don't generate a lot of tokens, it's good if you can model that in the benchmarking).

varad0309:
I believe this feature will come in handy when we don't want to make use of our own benchmarking method (it is needed when benchmarking with say vLLMs). Can we reconsider this PR once again? @Narsil

PR #953, opened 2023-08-30
Fix Falcon weight mapping for H2O.ai checkpoints

Vinno97:
# What does this PR do?
During the safetensor conversion, duplicate weights are removed. However, which of the duplicates gets removed, differs per checkpoint. In some, like `h2oai/h2ogpt-oig-oasst1-falcon-40b`, the weight `transformer.word_embeddings.weightSafetensor` gets removed. In others, `lm_head.weight` gets removed. Long story long, we need to support both.

Originally, f018143 mapped `lm_head` to `word_embeddings`. Then ac736fd switched this around. This commit merges them and allows for both.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@Narsil, you wrote both commits I referenced in this PR. I think you'll understand this change :)

### DIFF
--- a/server/text_generation_server/models/flash_rw.py
+++ b/server/text_generation_server/models/flash_rw.py
@@ -54,7 +54,10 @@ class FlashRWSharded(FlashCausalLM):
             device,
             dtype,
             process_group=self.process_group,
-            aliases={"lm_head.weight": ["transformer.word_embeddings.weight"]},
+            aliases={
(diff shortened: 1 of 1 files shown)


Narsil:
@Vinno97 The supposedly randomness should have been removed by earlier commits (using ttransformers hints).
Ofc the hints will not work for falcon, let's go with this.

Issue #962, opened 2023-08-31
Transformers based GPTQ Models missing bits weight


flozi00:
### System Info

latest docker

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [X] An officially supported command
- [ ] My own modifications

### Reproduction

docker run --pull always --gpus all -d --shm-size 1g -p 8080:80 -v ./data:/data ghcr.io/huggingface/text-generation-inference:latest --max-total-tokens 4096 --max-batch-prefill-tokens 4096 --max-input-length 3000 --model-id flozi00/codellama-34b-german-assistant-v1-4bit-autogptq --quantize=gptq --cuda-memory-fraction 0.5

### Expected behavior

The gptq bits value is not recognized from config.json what is needed when using gptq models from transformers instead of Autogptq which has its own config file for that

mallorbc:
I can confirm that I am having the same issue trying to use models converted using transformers/accelerate.  Gonna try different methods of conversion.

flozi00:
PR to solve is linked here

PR #963, opened 2023-08-31
add transformers gptq support

flozi00:
Proposal to fix https://github.com/huggingface/text-generation-inference/issues/962


mallorbc:
perhaps I am doing something wrong, but when I made this change and volumed in the change, this did not fix the issue.

docer-compose volume:
      - ./server/text_generation_server:/usr/src/server/text_generation_server

Narsil (approved):
LGTM.

PR #966, opened 2023-08-31
Fixing top_k tokens when k ends up < 0

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/utils/tokens.py
+++ b/server/text_generation_server/utils/tokens.py
@@ -363,9 +363,10 @@ def batch_top_tokens(
     # Find the new "fuzzy" top n values
     top_n_indices = (logprobs >= nth_highest).nonzero()
     _, top_n_ishes = torch.unique_consecutive(top_n_indices[:, 0], return_counts=True)
-
+ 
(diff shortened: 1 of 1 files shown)


Narsil:
@Vinno97

Narsil:
This seems to be triggerable on some GPTQ models (it was for me on a very bad checkpoint).

Vinno97:
Hi @Narsil do you have a way to reproduce this? Might be good to also add it to the tests.
How I'm reading the logic:

* This fixes `top_n_ishes` being empty for a certain batch item (request).
* Which should only happen when no logprob is higher or equal to the `nth_highest` logprob.
* Which may happen when all logprobs for thatrequest are are `-inf`?

Narsil:
Yes that's basically it. It was a trash model outputting garbage, therefore we could hit this.

However the fix shouldn't have any impact except let the queries go and not crash.

Vinno97:
Okay fair enough. Was thinking of catching this earlier in the process, but this works and is good enough

PR #981, opened 2023-09-05
fit for baichuan models

XiaoBin1992:
As more and more people begin to use Baichuan's open-source models, the influence of Baichuan models is growing, especially in China. Many community members are interested in adding support for Baichuan models to TGI. Meanwhile, Baichuan is a very open company, and in the future, it plans to open-source more and more models, taking all this into consideration, we would like to add support for the Baichuan model to TGI. To do this, we need to make some changes, which we hope can be merged into the main branch of TGI. In the future, we would be happy to help maintain support for Baichuan models in TGI. We sincerely hope that our pull request can be accepted. Thank you.

By the way, the changes of this time mainly for supporting Baichuan-7B.

### DIFF
--- a/server/text_generation_server/models/__init__.py
+++ b/server/text_generation_server/models/__init__.py
@@ -182,7 +182,7 @@ def get_model(
                 trust_remote_code=trust_remote_code,
             )
 
-    elif model_type == "llama":
+    elif model_type == "llama" or model_type == "baichuan":
         if FLASH_ATTENTION:
             return FlashLlama(
                 model_id,
--- a/server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
@@ -208,13 +208,37 @@ class FlashLlamaAttention(torch.nn.Module):
         if config.num_attention_heads != config.num_key_value_heads:
             self.query_key_value = _load_gqa(config, prefix, weights)
         else:
-            self.query_key_value = TensorParallelColumnLinear.load_multi(
-                config,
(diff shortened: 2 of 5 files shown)


Narsil:
@XiaoBin1992 

Thanks a lot for this PR. 
I modified things a bit in order to move the logic into `Weights` so we could support GPTQ and sharding together out of the box too:
[TheBloke/baichuan-7B-GPTQ](https://huggingface.co/TheBloke/baichuan-llama-7B-GPTQ)

For instance.

Narsil:
Llama tests ran manually.

No adding Baichuan in integration test because of the high overhead, and it's all about the loading rather than inference here.

jessiewiswjc:
@XiaoBin1992 @Narsil Thanks for your works. I want to know if this also supports Baichuan-13B whose attention mask is alibi while Baichuan-7B is rope.

razrLeLe:
@XiaoBin1992 I tried with Baichuan2-7b-Chat, the model can be loaded without error, but the generated text from api is wierd, is there any params missed during requesting model?
image

zTaoplus:
I encountered the same issue while using `baichuan2-13B-chat`..
I extracted the chat parameters from baichuan2's [generation_config.json](https://huggingface.co/baichuan-inc/Baichuan2-13B-Chat/blob/main/generation_config.json), and when I call the tgi interface, the result is as follows. 
image


When I invoke the `chat` method, the result is as follows. 
image



Here are the deployment parameters for the tgi.

``` bash
--max-batch-prefill-tokens 4096 --max-input-length 4096 --max-total-tokens 4608
```

PR #994, opened 2023-09-07
Added note on weight-cache-override

merveenoyan:
Added note on serving supported models from a different folder without re-downloading them.

### DIFF
--- a/docs/source/supported_models.md
+++ b/docs/source/supported_models.md
@@ -28,6 +28,10 @@ AutoModelForCausalLM.from_pretrained(<model>, device_map="auto")`
 AutoModelForSeq2SeqLM.from_pretrained(<model>, device_map="auto")
 ```
 
+If you wish to serve a different version of a model that exists in a local folder, you can use `weight-cache-override` flag like below 👇 
+```bash
+text-generation-launcher --model-id bigscience/bloom --weights-cache-override <PATH-TO-LOCAL-BLOOM>
+```
 
 ## Supported Hardware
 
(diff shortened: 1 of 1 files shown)


docs/source/supported_models.md
```diff
@@ -28,6 +28,11 @@ AutoModelForCausalLM.from_pretrained(<model>, device_map="auto")`
 AutoModelForSeq2SeqLM.from_pretrained(<model>, device_map="auto")
 ```
 
+If you wish to serve a supported model that already exists on a local folder, you can use `weight-cache-override` flag like below. Otherwise, it will be downloaded to Hugging Face Hub cache.
```
Narsil, line 31:
```suggestion
If you wish to serve a supported model that already exists on a local folder, Just point to the local folder.

text-generation-launcher --model-id <PATH-TO-LOCAL-BLOOM>
```

This is a bit simpler no ?

merveenoyan: I think it's better to show the opposite of what will happen if they don't do that given this is against what's happening by default. WDYT?

Narsil: I don't understand.

Default use case: -> Point to hub `model_id`.
Local model -> use the folder as `model_id`.

`weight-cache-override` does exist but it doesn't seem to be the easiest way to point users to.
For docker you can use `-v $volume:/data` which basically already cover the case, and in case of CLI and classifc `huggingface_hub` override will work (there are numerous env flags for it)

merveenoyan: > Default use case: -> Point to hub `model_id.`
Local model -> use the folder as `model_id`.

this is what I meant as well, but if you think my point doesn't go across this way I can commit your suggestion.

Issue #781, opened 2023-08-07
Add support for AWQ quantized models


0x1997:
Compared to GPTQ, [AWQ](https://github.com/mit-han-lab/llm-awq) is more accurate and has much better inference performance.

Benchmark: https://github.com/lm-sys/FastChat/blob/main/docs/awq.md#benchmark

~Note: Multi-Query Attention is [not yet supported](https://github.com/mit-han-lab/llm-awq/issues/53#issuecomment-1659417305).~

abhinavkulkarni:
I have released a bunch of AWQ quantized models here: https://huggingface.co/abhinavkulkarni?sort_models=downloads#models

Instructions on how to run these with HuggingFace API are in the model cards.

Narsil:
Can anyone run benchmarks against TGI + exllama kernels ? 

Those are supposed to provide a similar speedup.

We don't want to support every quantization scheme in TGI, just the best possible subset:

- No quantization: best PPL
- bitsandbytes: Low vram - no quantization steps - works on every model
- GPTQ: Low VRAM - fasted inference (should be ~2x if I'm not mistaken) with exllama.

abhinavkulkarni:
@Narsil: Any kernel optimizations done for GPTQ should translate to AWQ since they both are based on similar zero-point quantization schemes - they simply differ on how those exact zero-point weights are found and admittedly AWQ is superior to GPTQ. 

So, someone needs to simply write a "_translator_" from AWQ to GPTQ state dicts and everything else should work as is.

Narsil:
> Besides, awq can be applied to LLMs other than Llama, where exllama cannot.

This statement is wrong `exllama` is just because the kernels where created here: https://github.com/turboderp/exllama it has nothing to do with llama. (In general current quantization techniques juste replace `Linear` with `QuantLinear` basically).

edwardzjl:
> > Besides, awq can be applied to LLMs other than Llama, where exllama cannot.
> 
> This statement is wrong `exllama` is just because the kernels where created here: https://github.com/turboderp/exllama it has nothing to do with llama. (In general current quantization techniques juste replace `Linear` with `QuantLinear` basically).

My apologies, I made an assumption about exllama based solely on its name. I mistakenly thought it was specifically for Llama models. :sweat_smile:

Narsil:
No worries.

casperbh96:
[...]

PR #1019, opened 2023-09-13
Add AWQ quantization inference support

abhinavkulkarni:
# Add AWQ quantization inference support

Fixes https://github.com/huggingface/text-generation-inference/issues/781

This PR (partially) adds support for AWQ quantization for inference. More information on AWQ [here](https://arxiv.org/abs/2306.00978). In general, AWQ is faster and more accurate than GPTQ, which is currently supported by TGI.

This PR installs 4-bit GEMM custom CUDA kernels released by AWQ authors (in `requirements.txt`, just one line change).

Quick way to test this PR would be bring up TGI as follows:

```
text-generation-server download-weights abhinavkulkarni/codellama-CodeLlama-7b-Python-hf-w4-g128-awq

text-generation-launcher \
--huggingface-hub-cache ~/.cache/huggingface/hub/ \
--model-id abhinavkulkarni/codellama-CodeLlama-7b-Python-hf-w4-g128-awq \
--trust-remote-code --port 8080 \
--max-input-length 2048 --max-total-tokens 4096 --max-batch-prefill-tokens 4096 \
--quantize awq
```

Please note:
* This PR was tested with FlashAttention v2 and vLLM.
* This PR adds support for AWQ inference, not quantizing the models. That needs to be done outside of TGI, instructions [here](https://github.com/mit-han-lab/llm-awq/tree/f084f40bd996f3cf3a0633c1ad7d9d476c318aaa).
* This PR only adds support for `FlashLlama` models for now.
* Multi-GPU setup has not been tested. 
* No integration tests have been added so far, will add later if maintainers are interested in this change.
* This PR can be tested on any of the models released [here](https://huggingface.co/abhinavkulkarni?sort_models=downloads#models).

Please refer to the linked issue for benchmarks for [abhinavkulkarni/meta-llama-Llama-2-7b-chat-hf-w4-g128-awq](https://huggingface.co/abhinavkulkarni/meta-llama-Llama-2-7b-chat-hf-w4-g128-awq) vs [TheBloke/Llama-2-7b-Chat-GPTQ](https://huggingface.co/TheBloke/Llama-2-7b-Chat-GPTQ).

Please note, AWQ has released faster (and in case of Llama, fused) kernels for 4-bit GEMM, currently at the top of the `main` branch at https://github.com/mit-han-lab/llm-awq, but this PR uses an older commit that has been tested to work. We can switch to latest commit later on.

## Who can review?

@OlivierDehaene OR @Narsil

### DIFF
--- a/server/requirements.txt
+++ b/server/requirements.txt
@@ -73,3 +73,5 @@ win32-setctime==1.1.0 ; python_version >= "3.9" and python_version < "3.13" and 
 wrapt==1.15.0 ; python_version >= "3.9" and python_version < "3.13"
 xxhash==3.3.0 ; python_version >= "3.9" and python_version < "3.13"
 yarl==1.9.2 ; python_version >= "3.9" and python_version < "3.13"
+# Custom 4-bit GEMM AWQ kernels
+git+https://github.com/mit-han-lab/llm-awq.git@f084f40bd996f3cf3a0633c1ad7d9d476c318aaa#subdirectory=awq/kernels
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -256,12 +256,7 @@ def get_linear(weight, bias, quantize):
             raise NotImplementedError(
                 f"The passed weight is not `awq` compatible, loader needs to be updated."
             )
-        in_features = qweight.shape[0]
-        out_features = qweight.shape[1] * 32 // bits
-        linear = WQLinear(w_bit=bits, group_size=groupsize, in_features=in_features, out_features=out_features, bias=bias is not None, dev=qweight.device)
-        linear.qweight = qweight
-        linear.qzeros = qzeros
-        linear.scales = scales
+        linear = WQLinear(w_bit=bits, group_size=groupsize, qweight=qweight, qzeros=qzeros, scales=scales, bias=bias is not None)
     else:
         raise NotImplementedError(f"Quantization `{quantize}` is not implemented yet.")
     return linear
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -17,6 +17,7 @@ except ImportError:
 from accelerate import init_empty_weights
 
 from text_generation_server.utils.gptq.quant_linear import QuantLinear
+from text_generation_server.utils.awq.quantize.qmodule import WQLinear
 
 try:
     major, _minor = torch.cuda.get_device_capability()
@@ -248,6 +249,19 @@ def get_linear(weight, bias, quantize):
                 bits,
                 groupsize,
             )
+    elif quantize == "awq":
+        try:
+            qweight, qzeros, scales, bits, groupsize = weight
+        except Exception:
+            raise NotImplementedError(
+                f"The passed weight is not `awq` compatible, loader needs to be updated."
+            )
+        in_features = qweight.shape[0]
+        out_features = qweight.shape[1] * 32 // bits
+        linear = WQLinear(w_bit=bits, group_size=groupsize, in_features=in_features, out_features=out_features, bias=bias is not None, dev=qweight.device)
+        linear.qweight = qweight
+        linear.qzeros = qzeros
+        linear.scales = scales
     else:
         raise NotImplementedError(f"Quantization `{quantize}` is not implemented yet.")
     return linear
@@ -283,8 +297,8 @@ class TensorParallelHead(SuperLayer):
             weight = weights.get_tensor(f"{prefix}.weight")
             should_gather = False
 
-        # GPTQ doesn't quantize heads (nor embeddings)
-        if config.quantize == "gptq":
+        # GPTQ and AWQ don't quantize heads (nor embeddings)
+        if config.quantize in ["gptq", "awq"]:
             quantize = None
         else:
             quantize = config.quantize
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -251,7 +251,7 @@ def get_linear(weight, bias, quantize):
             )
     elif quantize == "awq":
         try:
-            qweight, qzeros, scales, bits, groupsize = weight
+            qweight, qzeros, scales, _, bits, groupsize, _ = weight
         except Exception:
             raise NotImplementedError(
                 f"The passed weight is not `awq` compatible, loader needs to be updated."
--- a/server/text_generation_server/utils/weights.py
+++ b/server/text_generation_server/utils/weights.py
@@ -150,6 +150,19 @@ class Weights:
 
             bits, groupsize = self._get_gptq_params()
             weight = (qweight, qzeros, scales, g_idx, bits, groupsize, False)
+        if quantize == "awq":
+            try:
+                qweight = self._get_qweight(f"{prefix}.qweight") 
+            except RuntimeError:
+                raise RuntimeError(
+                    "Cannot load `awq` weight, make sure the model is already quantized"
+                )
+            qzeros = self._get_qweight(f"{prefix}.qzeros") 
+            scales = self._get_qweight(f"{prefix}.scales") 
+            scales = scales.to(dtype=self.dtype)
+
+            bits, groupsize = self._get_gptq_params()
+            weight = (qweight, qzeros, scales, bits, groupsize)
         else:
             slice_ = self._get_slice(f"{prefix}.weight") 
             total_size = slice_.get_shape()[0]
@@ -194,6 +207,25 @@ class Weights:
 
             bits, groupsize = self._get_gptq_params()
             weight = (qweight, qzeros, scales, g_idx, bits, groupsize, False)
+        elif quantize == "awq":
+            try:
+                qweight = torch.cat(
+                    [self.get_sharded(f"{p}.qweight", dim=1) for p in prefixes], dim=1
+                )
+            except RuntimeError:
+                raise RuntimeError(
+                    "Cannot load `awq` weight, make sure the model is already quantized"
+                )
+
+            qzeros = torch.cat(
+                [self.get_sharded(f"{p}.qzeros", dim=1) for p in prefixes], dim=1
+            )
+            scales = torch.cat(
+                [self.get_sharded(f"{p}.scales", dim=1) for p in prefixes], dim=1
+            )
+
+            bits, groupsize = self._get_gptq_params()
+            weight = (qweight, qzeros, scales, bits, groupsize)
         else:
             w = [self.get_sharded(f"{p}.weight", dim=0) for p in prefixes]
             weight = torch.cat(w, dim=dim)
@@ -282,6 +314,20 @@ class Weights:
                 g_idx = self.get_sharded(f"{prefix}.g_idx", dim=0)
 
             weight = (qweight, qzeros, scales, g_idx, bits, groupsize, use_exllama)
+        elif quantize == "awq":
+            bits, groupsize = self._get_gptq_params()
+
+            try:
+                qweight = self.get_sharded(f"{prefix}.qweight", dim=0)
+            except RuntimeError:
+                raise RuntimeError(
+                    "Cannot load `awq` weight, make sure the model is already quantized"
+                )
+
+            qzeros = self.get_tensor(f"{prefix}.qzeros")
+            scales = self.get_tensor(f"{prefix}.scales")
+            
+            weight = (qweight, qzeros, scales, bits, groupsize)
         else:
             weight = self.get_sharded(f"{prefix}.weight", dim=1)
         return weight
--- a/server/text_generation_server/utils/weights.py
+++ b/server/text_generation_server/utils/weights.py
@@ -135,34 +135,29 @@ class Weights:
         Highly specific when the underlying tensor is a simple cat of Q,K,V instead of being
         already alternating Q,K,V within the main tensor
         """
-        if quantize == "gptq":
+        if quantize in ["gptq", "awq"]:
             try:
                 qweight = self._get_qweight(f"{prefix}.qweight") 
             except RuntimeError:
-                raise RuntimeError(
-                    "Cannot load `gptq` weight, make sure the model is already quantized, or quantize it with `text-generation-server quantize ORIGINAL_MODEL_ID NEW_MODEL_ID`"
-                )
+                if quantize == "gptq":
+                    raise RuntimeError(
+                        "Cannot load `gptq` weight, make sure the model is already quantized, or quantize it with `text-generation-server quantize ORIGINAL_MODEL_ID NEW_MODEL_ID`"
+                    )
+                else:
+                    raise RuntimeError(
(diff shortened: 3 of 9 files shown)


casper-hansen:
Amazing work @abhinavkulkarni. The original creators of AWQ recently removed all support of the GEMM kernel, however, after extensive testing of the GEMV kernel, I believe the GEMM kernel is the best for the highest throughput (large context, large batch size) as the GEMV kernel struggles with performance on anything but empty context and batch size > 1.

I am the author of [AutoAWQ](https://github.com/casper-hansen/AutoAWQ) and will maintain backward compatibility with GEMM as the main packing method and GEMV as an optional packing method (about to be merged). You can either use the old commit or switch to AutoAWQ, which has a PyPi package that is easily installable:

```
pip install autoawq
```

Narsil:
Hey thanks a LOT for this PR.

Before diving into the details, I'll drop some benchmark I did on single A100:

# f16:
Screenshot from 2023-09-22 18-47-19

# awq

Screenshot from 2023-09-22 18-45-35

# gptq

Screenshot from 2023-09-22 18-49-11


These tell a very good story.

Baseline latency for AWQ is 15ms/per, 25ms/token for f16 and 20ms/token for exllama.
But more importantly, the baseline holds much better for AWQ at higher throughput, giving more throughput at constant latency.

F16 seems to be managing to reach higher throughput but this regime is not the main one for usual TGI deployments, where latency is the key.

I will probably make a bit more testing on different TP sharding layout and model sizes but this looks really good.

server/requirements.txt
```diff
@@ -73,3 +73,5 @@ win32-setctime==1.1.0 ; python_version >= "3.9" and python_version < "3.13" and
 wrapt==1.15.0 ; python_version >= "3.9" and python_version < "3.13"
 xxhash==3.3.0 ; python_version >= "3.9" and python_version < "3.13"
 yarl==1.9.2 ; python_version >= "3.9" and python_version < "3.13"
+# Custom 4-bit GEMM AWQ kernels
+git+https://github.com/mit-han-lab/llm-awq.git@f084f40bd996f3cf3a0633c1ad7d9d476c318aaa#subdirectory=awq/kernels
```
Narsil, line 77: Is there anyway we could move to versioned releases ? 
Works currently it's a sha on main which is good enough for stability imo.

abhinavkulkarni: @Narsil: Not sure about licensing issues, but can we copy over the kernels to TGI repo from AWQ repo? That's all this `pip install` is pulling in. I saw somewhere that you have a `custom_kernels` directory in the repo.

server/text_generation_server/utils/layers.py
```diff
@@ -248,6 +249,19 @@ def get_linear(weight, bias, quantize):
...
+    elif quantize == "awq":
+        try:
+            qweight, qzeros, scales, bits, groupsize = weight
+        except Exception:
+            raise NotImplementedError(
+                f"The passed weight is not `awq` compatible, loader needs to be updated."
+            )
+        in_features = qweight.shape[0]
+        out_features = qweight.shape[1] * 32 // bits
+        linear = WQLinear(w_bit=bits, group_size=groupsize, in_features=in_features, out_features=out_features, bias=bias is not None, dev=qweight.device)
```
Narsil, line 261: Can we modify the init logit so the allocation doesn't have to happen ?

WQLinear(qweight, qzeros, scales, groupsize, bias)

And modify `__init__` such that it infers `in_features` and `out_features` directly ?

It's should lead to simpler code, avoid odd behavior of `register_buffer`, and be more in line with what we already have done.

server/text_generation_server/utils/weights.py
```diff
@@ -150,6 +150,19 @@ def get_weights_col_packed_qkv(self, prefix: str, quantize: str):
...
+            except RuntimeError:
+                raise RuntimeError(
+                    "Cannot load `awq` weight, make sure the model is already quantized"
+                )
+            qzeros = self._get_qweight(f"{prefix}.qzeros") 
+            scales = self._get_qweight(f"{prefix}.scales") 
+            scales = scales.to(dtype=self.dtype)
+
+            bits, groupsize = self._get_gptq_params()
+            weight = (qweight, qzeros, scales, bits, groupsize)
```
Narsil, line 165: Merging this with the `gptq` logic seems like it would be simpler to keep track of since I think everything is the same except `g_idx`

server/text_generation_server/utils/weights.py
```diff
@@ -194,6 +207,25 @@ def get_multi_weights_col(self, prefixes: List[str], quantize: str, dim: int):
 
             bits, groupsize = self._get_gptq_params()
             weight = (qweight, qzeros, scales, g_idx, bits, groupsize, False)
+        elif quantize == "awq":
```
Narsil, line 210: Same here.

server/text_generation_server/utils/weights.py
```diff
@@ -282,6 +314,20 @@ def get_multi_weights_row(self, prefix: str, quantize: str):
                 g_idx = self.get_sharded(f"{prefix}.g_idx", dim=0)
 
             weight = (qweight, qzeros, scales, g_idx, bits, groupsize, use_exllama)
+        elif quantize == "awq":
```
Narsil, line 317: Same here

Narsil:
Ok, my comments are mostly nits about code maintenance, but this PR looks great ! 

Thanks for it !

server/text_generation_server/utils/layers.py
abhinavkulkarni: @Narsil: This is fixed.

server/text_generation_server/utils/weights.py
abhinavkulkarni: @Narsil: This is fixed.

server/text_generation_server/utils/weights.py
abhinavkulkarni: @Narsil: This is fixed.

server/text_generation_server/utils/weights.py
abhinavkulkarni: @Narsil: This is fixed.

casper-hansen:
@Narsil @abhinavkulkarni Has this PR been tested to be compatible with the 300+ AWQ repositories that @TheBloke has pushed with AutoAWQ?

Most notably, I want to ensure that the models can be loaded in TGI with the `quant_config.json` that is provided.

Example: https://huggingface.co/TheBloke/Llama-2-13B-AWQ

abhinavkulkarni:
@TheBloke: Can we get some clarity on naming the quantization config for GPTQ and AWQ?

You seem to name it as  [`quantize_config.json`](https://huggingface.co/TheBloke/Llama-2-13B-GPTQ/blob/main/quantize_config.json) for GPTQ and [`quant_config.json`](https://huggingface.co/TheBloke/Llama-2-13B-AWQ/blob/main/quant_config.json) for AWQ.

[TGI code](https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/utils/weights.py#L314) expects it to be named `quantize_config.json`. 

Is it possible for you to rename it to `quantize_config.json` across all your AWQ repos? Or create a symlink in these repos so that both the files are present?

casper-hansen:
I would advise against renaming to `quantize_config.json` as that makes all his repositories incompatible with AutoAWQ, which was used to create the 300+ repositories.

TheBloke:
Yeah that's because AutoGPTQ and AutoAWQ create that file with a different name.  AutoGPTQ creates it as quantize_config.json, but AutoAWQ creats it as quant_config.json.

I'm happy to rename it across all, but it will require AutoAWQ updating to support that name (not quite sure why a different name was used by AutoAWQ)

abhinavkulkarni:
Okay, @TheBloke, @casper-hansen and @Narsil, please agree on one standard.

I'm happy to comply with whatever is decided. Two issues we need consensus on:

1. Wether to have `quantize_config.json` for GPTQ and `quant_config.json` for AWQ or just one for both

2. Secondly, uniformity in how we refer to quantization parameters - for e.g. in GPTQ `quantize_config.json`, we use `bits` but in `quant_config.json` of AWQ, we use `w_bit`. Same for `groupsize` for GPTQ vs `group_size` for AWQ.

Please let me know what you agree on and I'll modify the PR accordingly.

TheBloke:
@abhinavkulkarni GPTQ doesn't use `groupsize`, it's `group_size` there too.

IMHO the best option is for AutoAWQ to update to use `quantize_config.json`, and match the param names of AutoGPTQ.   Ultimately the hope is that AutoAWQ will be integrated into Transformers, like AutoGPTQ was.  If/when that happens, I am certain the Transformers team will ask for the same param names - though they won't care about the filename because the data will need to go into config.json

Here is the params that Transformers expects in config.json:
```
    "quantization_config": {
        "bits": 4,
        "group_size": 128,
        "damp_percent": 0.1,
        "desc_act": true,
        "sym": true,
        "true_sequential": true,
        "model_name_or_path": null,
        "model_file_base_name": "model",
        "quant_method": "gptq"
    }
```

Based on this, IMHO AutoAWQ should create a `quantize_config.json` matching this format:
```
"bits": 4,
"group_size": 128,
"zero_point": true,
"version": "GEMM"
```

Then that will be easily integrated into Transformers' config.json -> `quantization_config` section in future - with the addition of a `"quant_method": "awq"` key as well.

For backwards compatibility, AutoAWQ could also read from `quant_config` and the old key names, so that support for existing models is still provided.  Though I would update all my repo config files to the new format at such a time as the change was made.

casper-hansen:
I know that Huggingface takes backward compatibility very seriously. Implementing this in AutoAWQ requires quite a bit of maintenance going forward and it will break all PyPi releases of AutoAWQ (v0.0.1, v0.0.2, v0.1.0) because it will not be able to read any Huggingface repository. Additionally, the `oobabooga/text-generation-webui` will also need to update their version because it will brick the functionality in the PR. I am also not sure if this breaks the `vLLM` support for AWQ.

I am not so sure that people will be happy with this change for the reasons that I just explained, but I would gladly accept PRs that refactor the current behavior and extract the logic into a QuantizeConfig class that implements the necessary logic for backward compatibility. However, going forward, we have to support `quant_config.json` for multiple versions and raise warnings in AutoAWQ before we phase the `quant_config.json` out (just like they do in Huggingface libraries). We can finally phase it out in the next major release: `v0.2.0`.

TheBloke:
[vLLM already allows for both filenames](https://github.com/vllm-project/vllm/blob/947b794146aeae41ea17cbbe8bf9e53abc9c3f53/vllm/model_executor/quantization_utils/awq.py#L51):
```
   @classmethod
    def get_config_filenames(cls) -> List[str]:
        return [
            "quant_config.json",  # E.g., casperhansen/vicuna-7b-v1.5-awq
            "quantize_config.json",  # E.g., abhinavkulkarni/mosaicml-mpt-7b-instruct-w4-g128-awq  # pylint: disable=line-too-long
        ]
```

HF take backwards compatibility seriously once something is in a Hugging Face project - so they definitely won't want it to change after AWQ is merged.  Before it's merged I don't think they'll care one way or the other.  But they may well ask for the w_bits -> bits param change to be made in order for it to be merged, so I think you're going to end up doing at least that change eventually if you want it in Transformers.

So if this change is going to be made, I would say the time to do it is now - there's never a good time for a change that might break compatibility, but earlier is always far better than later

If the config file name does change, I'm happy to have both `quantize_config` and `quant_config.json` in all my repos for a while, which would allow ongoing support for prior AutoAWQ versions and any un-updated clients. Then I could phase it out in a a few weeks, once everyone is up-to-date. A bit like I've handled the many breaking llama.cpp changes.

server/requirements.txt
casper-hansen: Would recommend switching to `autoawq==0.1.0` as this has support for Windows as well. The `mit-han-lab` kernels break on Windows without the modifications I made.

abhinavkulkarni: Thanks @casper-hansen.

@Narsil: Okay to have a dependency on `autoawq==0.1.0`?

abhinavkulkarni:
I agree with @TheBloke that it's better to make changes now than later.

So, how about this:

1. @TheBloke, please write a "_translator_" to translate `AutoAWQ` `quant_config.json` to `quantize_config.json` and have both of them in the model repo. Please back-process all of your published models to have both of these files.

2. @casper-hansen: Over a period of time, please deprecate `quant_config.json` and start publishing `quantize_config.json` in the format HF team expects.

Agreed?

Narsil:
Ok, only nits are left. I will merge this into `dev` and add the nits I wanted to share and add an integration tests.

Thank you very much for this contribution !

Narsil:
Totally agree with all the names/config discussion, which is why I suggested reusing all the code we could for awq to match gptq.

Also in this repo, unlike in `transformers` we might want to make the necessary changes to  support the existing repos (provided they are not too hard to support as in little lines of code.)

server/text_generation_server/utils/weights.py
```diff
@@ -135,18 +135,26 @@ def get_weights_col_packed_qkv(self, prefix: str, quantize: str):
...
+                    )
 
             qzeros = self._get_qweight(f"{prefix}.qzeros") 
             scales = self._get_qweight(f"{prefix}.scales") 
             scales = scales.to(dtype=self.dtype)
-            g_idx = self.get_tensor(f"{prefix}.g_idx")
+            try:
+                g_idx = self.get_tensor(f"{prefix}.g_idx")
+            except RuntimeError:
+                g_idx = None
```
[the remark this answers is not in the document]
abhinavkulkarni: @Narsil: Can you elaborate on how to handle this? `g_idx = self.get_tensor(f"{prefix}.g_idx")` will raise an exception for `AWQ` since `g_idx` is not present in the state_dict.

So, should I modify this to the following:

```
try:
    g_idx = self.get_tensor(f"{prefix}.g_idx")
except RuntimeError as e:
    if quantize == "gptq":
        raise e
    g_idx = None
```

casper-hansen:
> Also looking at the fix PR : https://github.com/casper-hansen/AutoAWQ/commit/e1884728238bae0df16fde6094eaaed08ffc9762 (Is that correct ?) it seems the fix, is just removing the unecessary kernels.

This part is not Windows support for the GEMM kernel, those are FasterTransformer kernels.

Here I added Windows support. It's super simple. MSVC does not support `__asm__ __volatile__`, so just remove the underscores and it works. https://github.com/casper-hansen/AutoAWQ/commit/4c39a76a17d08f6d6eea95abe9f6c7ff7107ab24

Narsil:
Oh nice Thanks for the tip !

Archmilio:
There is only a 10% difference in inference between llama2 7b model quantized and unquantized on a single A100 GPU. In the autoAWQ inference example, there is a fuse_layer option that can significantly affect the speed depending on whether it is set to 'true/false'. I am curious if this option is always set to true in TGI

RonanKMcGovern:
Is flash attention v2 installed and on?

Archmilio:
I tested using the tgi-1.1.0 Docker image, which has Flash Attention v2 installed.

PR #1024, opened 2023-09-14
Fix a tokenizer issue in decoding generated tokens

bangoz:
# What does this PR do?

### TL;DR
Fixes a tokenization issue in `Client.generate()` method which mistakenly removes a blank space at the beginning of the output texts.

### Replication
The concerned issue can be replicated by running code generation with `CodeLlama-7b-Python-hf` model:

```bash
model=codellama/CodeLlama-7b-Python-hf
volume=$PWD/data
docker run --rm --gpus all --shm-size 1g \
    -p 8080:80 -v $volume:/data ghcr.io/huggingface/text-generation-inference:latest \
    --model-id $model --sharded true --num-shard 4
```

Then, with a prompt for code completion, calling `client.generate()` and `client.generate_stream()` result in different outputs:

```python
from text_generation import Client

client = Client("http://127.0.0.1:8080")
prompt = '\n\ndef truncate_number(number: float) -> float:\n    """ Given a positive floating point number, it can be decomposed into\n    and integer part (largest integer smaller than given number) and decimals\n    (leftover part always smaller than 1).\n\n    Return the decimal part of the number.\n    >>> truncate_number(3.5)\n    0.5\n    """\n'

### `client.generate()` method
print(client.generate(prompt=prompt, do_sample=True, top_k=10, temperature=0.1, top_p=0.95, max_new_tokens=10,repetition_penalty=1.05).generated_text)
# output: '  return number - int(number)'

### `client.generate_stream()` method
for response in client.generate_stream(prompt=prompt, do_sample=True, top_k=10, temperature=0.1, top_p=0.95, max_new_tokens=10,repetition_penalty=1.05):
    if not response.token.special:
        print(response.token.text, end="")
# output: '    return number - int(number)'
```

The latter output starts with 4 whitespaces, which matches the indent format in Python so the completed function runs correctly; while the former output starts with only 2 whitespaces. Such indent mismatching is DETRIMENTAL in code generation tasks, especially for indent-aware languages such as Python. The resulting indent error dramatically affects the correctness of `CodeLlama-7b-Python-hf` model on HumanEval (pass@1) from >30% to ~3%.

The root cause of the `client.generate()` case is at Line 1011 in `text-generation-inference/server/text_generation_server/models/flash_causal_lm.py` : 

```python
### text-generation-inference/server/text_generation_server/models/flash_causal_lm.py ###
# Decode generated tokens
output_text = self.decode(
    all_input_ids[-stopping_criteria.current_tokens :]
)
```

The generated tokens are jointly sent to `self.decode()` method, which includes an additional step of removing the prefix space at the first token (in our case this method executes the following `convert_tokens_to_string()` function in `transformers/models/llama/tokenization_llama.py` ):

```python
### transformers/models/llama/tokenization_llama.py ###
def convert_tokens_to_string(self, tokens):
        """Converts a sequence of tokens (string) in a single string."""
        # since we manually add the prefix space, we have to remove it when decoding
        if tokens[0].startswith(SPIECE_UNDERLINE):
            tokens[0] = tokens[0][1:]
        ...
```

Since the generated tokens do not start from the beginning of a sentence, the above space removal is redundant and problematic.

For the `client.generate_stream()` case, generated tokens are printed out one by one, bypassing the aforementioned function, so prefix spaces are retained.

### Solution
To ensure two outputs to be consistent, we add a `self.decode_generated_tokens()` method in `Model` class, which works by including an additional prefix tokenizer for decoding, and then substract it from the output text:

```python
### text-generation-inference/server/text_generation_server/models/model.py ###
def decode_generated_tokens(
        self,
        all_input_ids: List[int],
        num_tokens: int = 0,
    ) -> str:
        # Like in `decode_token()`, the prefix text is necessary only to defeat cleanup algorithms in the decode.
        prefix_text = self.tokenizer.decode(
            all_input_ids[-num_tokens-1:-num_tokens], skip_special_tokens=False
        )
        new_text = self.tokenizer.decode(
            all_input_ids[-num_tokens-1:], skip_special_tokens=False
        )
        new_text = new_text[len(prefix_text):]
        return new_text
```

The modified line in `text-generation-inference/server/text_generation_server/models/flash_causal_lm.py` becomes:

```python
### text-generation-inference/server/text_generation_server/models/flash_causal_lm.py ###
# Decode generated tokens
output_text = self.decode_generated_tokens(
    all_input_ids,
    stopping_criteria.current_tokens,
)
```

So that both generation methods yield the same output. Similar modification is also applied to `causal_lm.py` and `idefics_causal_lm.py` .

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/models/model.py
+++ b/server/text_generation_server/models/model.py
@@ -85,6 +85,21 @@ class Model(ABC):
             return new_text, read_offset, len(all_input_ids)
         else:
             return "", prefix_offset, read_offset
+    
+    def decode_generated_tokens(
+        self,
+        all_input_ids: List[int],
+        num_tokens: int = 0,
+    ) -> str:
+        # Like in `decode_token()`, the prefix text is necessary only to defeat cleanup algorithms in the decode.
+        prefix_text = self.tokenizer.decode(
+            all_input_ids[-num_tokens-1:-num_tokens], skip_special_tokens=False
+        )
+        new_text = self.tokenizer.decode(
+            all_input_ids[-num_tokens-1:], skip_special_tokens=False
+        )
+        new_text = new_text[len(prefix_text):]
+        return new_text
 
     def check_initialized(self):
         uninitialized_parameters = []
--- a/server/text_generation_server/models/causal_lm.py
+++ b/server/text_generation_server/models/causal_lm.py
@@ -641,8 +641,8 @@ class CausalLM(Model):
             if i % self.world_size == self.rank:
                 if stop:
                     # Decode generated tokens
-                    output_text = self.decode(
-                        all_input_ids[-stopping_criteria.current_tokens :, 0]
+                    output_text = self.decode_generated_tokens(
+                        all_input_ids[:, 0], stopping_criteria.current_tokens
                     )
                     # Get seed
                     if isinstance(next_token_chooser.choice, Sampling):
--- a/server/text_generation_server/models/flash_causal_lm.py
+++ b/server/text_generation_server/models/flash_causal_lm.py
@@ -1008,9 +1008,11 @@ class FlashCausalLM(Model):
             if i % self.world_size == self.rank:
                 if stop:
                     # Decode generated tokens
-                    output_text = self.decode(
-                        all_input_ids[-stopping_criteria.current_tokens :]
+                    output_text = self.decode_generated_tokens(
+                        all_input_ids,
+                        stopping_criteria.current_tokens,
                     )
+                    
                     generated_text = GeneratedText(
                         output_text,
                         stopping_criteria.current_tokens,
--- a/server/text_generation_server/models/idefics_causal_lm.py
+++ b/server/text_generation_server/models/idefics_causal_lm.py
@@ -728,8 +728,8 @@ class IdeficsCausalLM(Model):
             if i % self.world_size == self.rank:
                 if stop:
                     # Decode generated tokens
-                    output_text = self.decode(
-                        all_input_ids[-stopping_criteria.current_tokens :, 0]
+                    output_text = self.decode_generated_tokens(
+                        all_input_ids[:, 0], stopping_criteria.current_tokens
                     )
                     # Get seed
                     if isinstance(next_token_chooser.choice, Sampling):


server/text_generation_server/models/model.py
```diff
@@ -85,6 +85,21 @@ def decode_token(
...
             return "", prefix_offset, read_offset
+    
+    def decode_generated_tokens(
+        self,
+        all_input_ids: List[int],
+        num_tokens: int = 0,
+    ) -> str:
+        # Like in `decode_token()`, the prefix text is necessary only to defeat cleanup algorithms in the decode.
+        prefix_text = self.tokenizer.decode(
+            all_input_ids[-num_tokens-1:-num_tokens], skip_special_tokens=False
```
SwordYork, line 96: I found that the parameters here are different from the original. Can you confirm if there is a need to maintain consistency with the original?

` skip_special_tokens=True, clean_up_tokenization_spaces=False
` 

https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/models/seq2seq_lm.py#L580

bangoz: Thanks for pointing this out!

The decoding parameters in the current main branch are different for `generate()` and `generate_stream()`. Essentially,

- the `generate()` function decodes with `skip_special_tokens=True, clean_up_tokenization_spaces=False`: [https://github.com/huggingface/text-generation-inference/blob/c8a01d759173483efc2135c4e7506b23e14e7fc4/server/text_generation_server/models/flash_causal_lm.py#L796-L799](url);

 - the `generate_stream()` function decodes with `skip_special_tokens=False`:
[https://github.com/huggingface/text-generation-inference/blob/c8a01d759173483efc2135c4e7506b23e14e7fc4/server/text_generation_server/models/model.py#L72-L77](url).

So whether to align our modification with previous `generate()` or `generate_stream()` is well worth discussing.

g4rg:
Thank you for this. Continuing chat messages or prose is a frequent use case of my user base, currently leading to frustration with spaces between words getting removed.

Narsil:
In https://github.com/huggingface/text-generation-inference/pull/1065 I suggest a simpler fix, doing basically what you suggest, but reusing the existing logic for it.

The stripping of the space is quite necessary for regular text, maybe we should have dropped it in the conversion maybe, @ArthurZucker for codellama ?

Narsil:
@ArthurZucker I meant the last `Strip`  in the decoder. It's quite useful for text, so you generate "Hello" instead of " Hello", maybe it doesn't make as much sense for code.. I'm not sure we should have to jump through that many hoops (it *is* imho a very unique code thing, not necessarily desirable for text)

ArthurZucker:
It is desirable for code indeed! Specifically for CodeLlama. I was mostly answering to @bangoz on the origin of the issue. 
For TGI the issue is with the fast tokenizer, but the same issue exists in transformers, which is indeed coming from:
```python 
         if tokens[0].startswith(SPIECE_UNDERLINE):
             tokens[0] = tokens[0][1:]
```
Which we can / should remove. I'll try to do this for both fast and slow! 
CodeLlama is designed for code so yep let's update this

Narsil:
Should be fixed by https://github.com/huggingface/text-generation-inference/pull/1065 @bangoz you're co-author of the other PR.

Thanks a lot for this. I will close this PR, feel free to ping me if you think we should keep it open.

bangoz:
> In #1065 I suggest a simpler fix, doing basically what you suggest, but reusing the existing logic for it.
> 
> The stripping of the space is quite necessary for regular text, maybe we should have dropped it in the conversion maybe, @ArthurZucker for codellama ?

Simply reusing `decode_token` method may result in another problem. The `decode_token` contains the following statement: 
```
if len(new_text) > len(prefix_text) and not new_text.endswith("�"):
      # utf-8 char at the end means it's a potential unfinished byte sequence
      # from byte fallback tokenization.
      # If it's in the middle, it's probably a real invalid id generated
      # by the model
      new_text = new_text[len(prefix_text) :]
      return new_text, read_offset, len(all_input_ids)
else:
      return "", prefix_offset, read_offset
```
if the generated text ends with half tokens of a utf-8 character (this may occur when the number of generated tokens reach the max limit), then `output_text` will be nothing.

PR #1044, opened 2023-09-21
Automatic docs for launcher

merveenoyan:
This PR includes a github action to have the `text-generation-launcher --help` output in the documentation.


merveenoyan:
Tested on my local, will update credentials to run this here.

.github/workflows/auto_docs.yml
```diff
@@ -0,0 +1,30 @@
+name: Automatic Documentation for Launcher
+
+on:
+  push:
+    branches:
+      - main
```
osanseviero, line 6: Should this run just when the files with the parameters are changed? E.g. something like https://github.com/huggingface/text-generation-inference/blob/main/.github/workflows/build_documentation.yml#L5C5-L6C25

.github/workflows/auto_docs.yml
```diff
@@ -0,0 +1,30 @@
...
+    
+    - name: Update Docs
+      run: |
+        output="${{ steps.run-launcher-help.outputs.output }}"
+        
+        echo $output >> ../../docs/source/basic_tutorials/launcher.md
+
+        git add launcher.md
+        git commit -m "Update launcher.md"
+        git push
```
osanseviero, line 30: This would be doing a new push to main, but the workflow triggers with every new push. Wouldn't this trigger an infinite loop? (not sure)

Narsil:
Neat !  Super simple and efficient, I like it !

Don't know about our docs, but if we could include that in a code block would that make it more readable (I don't think the current output is valid markdown)

.github/workflows/auto_docs.yml
osanseviero: Personally, I think it would be better to make this check that `launcher.md` is up to date (as we do with `black` or `isort`, they just check but don't change files), and fail if running `text-generation-launcher --help >> launcher.md` would change the contents of `launcher.md`, hence forcing users to run this command locally. 
(no strong opinion, I think @Narsil had suggested the approach of pushing post-merge)

.github/workflows/auto_docs.yml
```diff
@@ -0,0 +1,30 @@
...
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    
+    - name: Run TGI Launcher docs
+      id: run-launcher-help
+      run: |
+        output=$(text-generation-launcher --help)
```
osanseviero, line 19: I don't think the GitHub Actions workflow has anything installed to run `text-generation-launcher --help`. You will need to either install this in GH Actions or use a Docker GA that allows you to do `docker run ghcr.io/huggingface/text-generation-inference:latest --help` (which I think has same output)

osanseviero (commented):
Exciting! :partying_face:

.github/workflows/auto_docs.yml
Narsil: I setup what you said of having a check instead of force-push, and it also works. Both are fine, but indeed it could trigger infinite loop I don't knwo how GH handles those.

.github/workflows/auto_docs.yml
Narsil: Maybe include it in the `build` step too (because there it's available as the docker image)

merveenoyan:
@Narsil just edited to add codeblock.

.github/workflows/auto_docs.yml
merveenoyan: ah right. @Narsil is `docker run` more lightweight compared to only building launcher? 
```
cd ../../launcher
cargo build
```

.github/workflows/auto_docs.yml
merveenoyan: Hmm, does it make sense run this everytime there's a change in `main.rs` instead, given it's where the docs come from?

.github/workflows/auto_docs.yml
merveenoyan: Isn't this always be up to date since it runs every time there's a change in `main.rs`? 
```
rm ../../docs/source/basic_tutorials/launcher.md
echo '```' >> ../../docs/source/basic_tutorials/launcher.md
echo $output >> ../../docs/source/basic_tutorials/launcher.md
echo '```' >> ../../docs/source/basic_tutorials/launcher.md
```
so whenever there's a change in the file it will always remove the launcher.md and then create new one and post output. Is this what you mean? 

About when to push, I agree I should change it to post-merge. I've never done it in GHA so I'm looking into it.

.github/workflows/auto_docs.yml
merveenoyan: fixed this.

.github/workflows/auto_docs.yml
merveenoyan: I added the above two lines to action file.

Narsil:
> @Narsil just edited to add codeblock.

By the way to test gh action, what I usually do is remove the filter so that it runs on the current PR

.github/workflows/auto_docs.yml
merveenoyan: I will think more on this, I can't push this way either :D

merveenoyan:
Apparently if we do:
```
on:
  pull_request:
    types:
      - closed

jobs:
  update_docs:
    if: github.event.pull_request.merged == true
```
it should work. I will not add this now to test the action on this branch though:
````
  pull_request:
    types:
      - closed
```

.github/workflows/auto_docs.yml
osanseviero: My suggestion was to, rather than GA updating `launcher.md` directly, just have GA check if the changes in `main.rs` would trigger a change in `launcher.md`. E.g. you can do this by checking if output == content in launcher.md. If it's the same, nothing happens. If it's different, then the GA workflow fails. 

This means that users, not GA, will be forced to run the command to update the doc (it could be a bash script encapsulating everything as well). (again, no strong opinion, it has its pros and cons)

.github/workflows/auto_docs.yml
osanseviero: Yes great idea!

merveenoyan:
continuing from this PR since GHA is not recognized in my fork (not even skipped) [#1045](https://github.com/huggingface/text-generation-inference/pull/1045/files)

.github/workflows/auto_docs.yml
merveenoyan: I realized `make install` works better for this case.

.github/workflows/auto_docs.yml
merveenoyan: I did it like so: (also started working in another PR since in this fork the action is not triggered)
 ```
   - name: Update Docs
      run: |
        output="${{ steps.run-launcher-help.outputs.output }}"
        launcher_content=$(/docs/source/basic_tutorials/launcher.md)

        if [ "$launcher_content" != "$output" ]; then
          rm /docs/source/basic_tutorials/launcher.md
          echo '```' >> /docs/source/basic_tutorials/launcher.md
          echo $output >> /docs/source/basic_tutorials/launcher.md
          echo '```' >> /docs/source/basic_tutorials/launcher.md
          echo "Docs are updated!"

          git add launcher.md
          git commit -m "Update launcher.md"
          git push
```

PR #1045, opened 2023-09-21
Automatic docs for TGI

merveenoyan:
I had to open this PR since I initially worked from my fork, and it requires a handful of work to trigger a new github action on my fork's specific branch (couldn't find a way, at least, despite trying all of them).


merveenoyan:
We can do `workflow_dispatch:` for manually running it, keeping this running for every push until validating that the code works.

.github/workflows/autodocs.yml
```diff
@@ -0,0 +1,33 @@
...
+  update_docs:
+    runs-on: ubuntu-latest
+    
+    steps:
+    - name: Checkout code
+      uses: actions/checkout@v2
+    
+    - name: Install Launcher
+      id: install-launcher
+      run: cargo install --git https://github.com/huggingface/text-generation-inference.git text-generation-launcher
```
osanseviero, line 18: I don't think GA has cargo by default. Would you be able to trigger the workflow in this PR so we can see the results? (right now `launched.md` seems empty)

merveenoyan: `text-generation-launcher` command would've failed then, I can see the output in the GHActions logs (I did echo for sanity check)

merveenoyan: `diff: launcher.md: No such file or directory` this is what fails, it's because in main branch currently it doesn't exist.

merveenoyan: I'll add launcher.md for now since in GHactions it says `diff: launcher.md: No such file or directory` so it doesn't fall into the if

merveenoyan: it's the file paths that cause it to not fall into if statement (diff says there's no file) so I fixed it

merveenoyan: it writes the file I can echo file content and verify but doesn't commit to this branch for some reason

merveenoyan:
@Narsil I can confirm I can write to file yet can't commit it to this repository, do I need anything for it? do we have some sort of an org token for github creds in GHA?

.github/workflows/autodocs.yml
```diff
@@ -0,0 +1,35 @@
...
+        output=$(text-generation-launcher --help)
+        { echo '```'; echo $output; echo '```'; } > launcher_test.md
+        if diff launcher_test.md docs/source/basic_tutorials/launcher.md; then
+          rm docs/source/basic_tutorials/launcher.md
+          { echo '```'; echo $output; echo '```'; } > docs/source/basic_tutorials/launcher.md
+          echo "$(cat docs/source/basic_tutorials/launcher.md)"
+          echo "Docs are updated!"
+          git add docs/source/basic_tutorials/launcher.md
+          git commit -m "Update launcher.md"
+          git push
```
pcuenca, line 32: I saw in the other PR that there was some discussion about using `git push` inside the action. Was the conclusion that this is safe to do?

.github/workflows/autodocs.yml
```diff
@@ -0,0 +1,35 @@
+name: Automatic Documentation for Launcher
+
+on:
+  push:
+    branches:
+      - launcher_docs
```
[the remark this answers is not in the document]
merveenoyan: No, it will be `workflow_dispatch` given we want to do it manually.

.github/workflows/autodocs.yml
```diff
@@ -0,0 +1,35 @@
...
+      run: cargo install --git https://github.com/huggingface/text-generation-inference.git text-generation-launcher
+    
+    - name: Update Docs
+      run: |
+        echo text-generation-launcher --help
+        output=$(text-generation-launcher --help)
+        { echo '```'; echo $output; echo '```'; } > launcher_test.md
+        if diff launcher_test.md docs/source/basic_tutorials/launcher.md; then
+          rm docs/source/basic_tutorials/launcher.md
+          { echo '```'; echo $output; echo '```'; } > docs/source/basic_tutorials/launcher.md
```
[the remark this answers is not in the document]
merveenoyan: It's not persisted anyway, I don't want to change the code even further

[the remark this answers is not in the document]
pcuenca: That's cool, just in case you thought it was easier to understand. No need to change, I agree.

.github/workflows/autodocs.yml
merveenoyan: The concern wasn't this but rather about triggering an infinite loop. Manually running solves this.

.github/workflows/autodocs.yml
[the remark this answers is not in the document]
merveenoyan: since this is not really very different (it's created during run and not persisted) I'd like to avoid changing the code even further

merveenoyan:
@Narsil I think this is good to go.

.github/workflows/autodocs.yml
Narsil: Don't  you need to install the current PR branch? (This install main I think)

merveenoyan: I think last time we agreed on checking it after the PR would be merged to main, let me change this.

osanseviero (commented):
I'm a bit confused by the changes between the PRs, commits, and replies in some of the comments. It would be great to clarify what we're aiming to here :)

**If we want to do the automatic check approach**:

* We should trigger on every PR when `main.rs` is changed
* We should fail if `launcher.md` would be changed
* We should prompt the user to update `launcher.md`, likely providing the command to update it in the error logs
* We don't need the `GITHUB_TOKEN` in the workflow (edit: saw it was removed in a new commit)

**If we want to do the automatic update approach**

* We should trigger manually
* We should update the `launcher.md` and open a PR automatically
* We do need the `GITHUB_TOKEN`

The current implementation seems like a mix of both. From what I understand from the last version, I **think** we're aiming for the first approach. If this is the case, we should not trigger manually. Instead, the GA should trigger on every push to a PR that modifies `main.rs`. Otherwise, we would have a GA workflow that would give an error and force us to open a new PR manually.

merveenoyan:
@osanseviero I will explain once I can get github context variables working.

merveenoyan:
Let me explain what I wanted to do bcs it doesn't work: I wanted to create a github action that prompts user to update the docs themselves, so it only does the assertion with their own branch's `text-generation-launcher --help` command and their branch's docs. I think my script as of now achieves this but github context variables don't work for some reason.

Narsil:
github template is `${{ xx }}`

merveenoyan:
Hey @Narsil my plan was to run the action to check up-to-date manually so I was going to change the trigger to`workflow_dispatch` before merger, just wanted to put it to here. The PR looks super nice, very sorry I'm in a conference+vacay that couldn't deliver earlier.

.github/workflows/autodocs.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     
     - name: Install Launcher
       id: install-launcher
-      run: cargo install --git git+https://github.com/${{ github.repository }}/text-generation-inference.git --branch ${{ github.head_ref }} text-generation-launcher
+      run: cargo install --git git+https://github.com/${{ github.repository }} --branch ${{ github.head_ref }} text-generation-launcher
```
merveenoyan: I spent two hours trying to get this right. 
image

.github/workflows/autodocs.yml
```diff
@@ -0,0 +1,21 @@
+name: Automatic Documentation for Launcher
+
+on:
```
merveenoyan, line 3: to be modified for manual trigger

update_doc.py
```diff
@@ -0,0 +1,31 @@
...
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--check", action="store_true")
+
+    args = parser.parse_args()
+
+    output = subprocess.check_output(["text-generation-launcher", "--help"]).decode("utf-8")
+    final_doc = f"# Text-generation-launcher arguments\n```\n{output}\n```"
```
merveenoyan, line 12:
```suggestion
    final_doc = f"# Text Generation Launcher Arguments\n```\n{output}\n```"
```

merveenoyan (commented):
I learn a lot from you 😀

Narsil:
> Hey @Narsil my plan was to run the action to check up-to-date manually so I was going to change the trigger to`workflow_dispatch` before merger, just wanted to put it to here. The PR looks super nice, very sorry I'm in a conference+vacay that couldn't deliver earlier. 

No worries. Don't apologies, we're all pretty busy, I understand.

Thank you for your work and getting all the doc off the ground. This is highly appreciated!

PR #1048, opened 2023-09-22
dtype default to None instead of float16 and each model could set it'…

sywangyi:
…s default type according to the platform

cuda: float16 if dtype is None
cpu: bfloat16 if dtype is None

@OlivierDehaene OR @Narsil

### DIFF
--- a/server/text_generation_server/models/__init__.py
+++ b/server/text_generation_server/models/__init__.py
@@ -76,13 +76,11 @@ def get_model(
     dtype: Optional[str],
     trust_remote_code: bool,
 ) -> Model:
-    if dtype is None:
-        dtype = torch.float16
-    elif dtype == "float16":
+    if dtype == "float16":
         dtype = torch.float16
     elif dtype == "bfloat16":
         dtype = torch.bfloat16
-    else:
+    elif dtype is not None:
         raise RuntimeError(f"Unknown dtype {dtype}")
 
     if "facebook/galactica" in model_id:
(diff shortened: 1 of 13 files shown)


Narsil:
What problem are you fixing exactly ?

sywangyi:
When you launch tgi server like "text-generation-launcher -p 80 --model-id huggyllama/llama-7b", the default dtype is None. in https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/models/__init__.py#L79-L80, if dtype is None, it will set to be float16. for cpu platform. we prefer default type to be bfloat16 since there's accelerator for bfloat16. in each model implementation. there's logic like https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/models/mpt.py#L50-L55. so the dtype should not be set to float16 if it's None in get_model()

Narsil:
I see. We cannot modify the default. I don't think there's enough bf16 support on CPU generally to make the switch on the default.

I think for the time being adding `--dtype bfloat16` should be preferred on the platforms that do support it.

To be perfectly honest the CPU support is mostly intended for us to make easy debugging on dummy models, it's interesting to see there are some real usage out there.

@OlivierDehaene For a second opinion on the flag.

chiragjn:
Would it be preferable if `dtype` is default set to auto and read from the config instead?, as I understand models trained/tuned with bfloat16 can cause unexpected quality issues with float16

sywangyi:
close according to the review comment

PR #1049, opened 2023-09-22
pass max_total_tokens info through warmup, python could get max_total…

sywangyi:
…_tokens as truncate_len+max_new_token in warmup

@OlivierDehaene OR @Narsil

### DIFF
--- a/router/client/src/client.rs
+++ b/router/client/src/client.rs
@@ -103,17 +103,19 @@ impl Client {
         &mut self,
         max_input_length: u32,
         max_prefill_tokens: u32,
+        max_total_tokens: u32,
     ) -> Result<Option<u32>> {
         let mut n_tokens = 0;
         let mut requests = Vec::new();
-
+        let mut truncate = 0;
         // Create requests
         while n_tokens < max_prefill_tokens {
+            truncate = min(max_input_length, max_prefill_tokens - n_tokens);
             requests.push(Request {
                 id: 0,
                 // We truncate the input on the server side to be sure that it has the correct size
                 inputs: "_test ".to_string().repeat(max_input_length as usize),
-                truncate: min(max_input_length, max_prefill_tokens - n_tokens),
+                truncate: truncate,
                 // Set sampling parameters to also take these ops into account in the max memory
                 parameters: Some(NextTokenChooserParameters {
                     temperature: 0.9,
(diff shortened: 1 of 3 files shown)


sywangyi:
there's needs to know the max_total_tokens in python. like
https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/utils/gptq/exllama.py#L51
https://github.com/huggingface/optimum-habana/blob/main/text-generation-inference/server/text_generation_server/models/causal_lm.py#L101
could we pass the max_total_tokens like this in warmup? warmup could get the max_total_tokens as  truncate_len+max_new_token and set it to model.

sywangyi:
Hi, @Narsil I see code like https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/utils/gptq/exllama.py#L132. For habana, I need to get the max_total_tokens info to compile the graph. or else if max_new_token is changed per request. graph recompile is needed now.

sywangyi:
Hi, @Narsil, any thought about the max_total_tokens? for habana, I use it here https://github.com/huggingface/optimum-habana/blob/main/text-generation-inference/server/text_generation_server/models/causal_lm.py#L101. or could you share me your insight about the issue?

Narsil:
In general this code tries to be as agnostic as possible to hardware specific implementations (since there are many different things that could happen).

IMO the python code should be agnostic to what warmup does, doesn't matter if users are requesting 2 new tokens or more.

Isn't there a way to fix without modifying the code here ?

sywangyi:
> max_total_tokens

Any suggestion to pass the max_total_tokens to python? I check all the python api, seems the only api I could leverage is warmup. If you have any better idea, I could follow yours. thanks for your help.

sywangyi:
add @regisss for the info.

yao-matrix:
> In general this code tries to be as agnostic as possible to hardware specific implementations (since there are many different things that could happen).
> 
> IMO the python code should be agnostic to what warmup does, doesn't matter if users are requesting 2 new tokens or more.
> 
> Isn't there a way to fix without modifying the code here ?

I think it's not the habana-specific thing. Personally, I treat it as the difference btw graph mode and eager model. Normally, to pursue better performance and lower TCO, AI HWs including CPU, GPU and AI ASICs(such as Habana) will require a hint on input tensor size so they can "warmup" and prepare for peak performance. One example is torch JIT mechanism. So in this case, we suppose it should be included in warmup phase.

I suppose pass this hint from front end(RUST code) to backend(Python code) opens the door for performance optimization for backends, and I think how each hardware use this info is hardware-specific, like eager mode device will special handle it, but graph mode device will special handle it. But I think pass the max_total_tokens from RUST frontend to Python backend is not device-specific, habana here is just an example of graph mode devices.

@Narsil , pls let me know your insights, thx.

OlivierDehaene (approved):

sywangyi:
Hi, @OlivierDehaene  I see you merge the PR to external branch. what's external branch for?

OlivierDehaene:
Running the CI. 
As an external contributor you don't have the rights to run it.

sywangyi:
ok, I see CI failure because of rust fmt check failure, should I fix it in another pr? @OlivierDehaene

PR #1052, opened 2023-09-25
fix discard_names bug in safetensors convertion

zhangsibo1129:
# What does this PR do?

Model Class attributes `_tied_weights_keys`, ` _keys_to_ignore_on_load_missing` can only be `None` or a List. `getattr(class_, "_keys_to_ignore_on_load_missing", [])` will return `None` if `_keys_to_ignore_on_load_missing` is None, and `discard_names.extend(None)` will trigger an exception, even though `_tied_weights_keys`  exists.

## Who can review?

@OlivierDehaene  @Narsil

### DIFF
--- a/server/text_generation_server/cli.py
+++ b/server/text_generation_server/cli.py
@@ -186,8 +186,11 @@ def download_weights(
             class_ = getattr(transformers, architecture)
 
             # Name for this varible depends on transformers version.
-            discard_names = getattr(class_, "_tied_weights_keys", [])
-            discard_names.extend(getattr(class_, "_keys_to_ignore_on_load_missing", []))
+            discard_names = []
+            if getattr(class_, "_tied_weights_keys", []):
+                discard_names.extend(getattr(class_, "_tied_weights_keys", []))
+            if getattr(class_, "_keys_to_ignore_on_load_missing", []):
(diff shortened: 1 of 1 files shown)


server/text_generation_server/cli.py
```diff
@@ -186,8 +186,11 @@ def download_weights(
             class_ = getattr(transformers, architecture)
 
             # Name for this varible depends on transformers version.
-            discard_names = getattr(class_, "_tied_weights_keys", [])
-            discard_names.extend(getattr(class_, "_keys_to_ignore_on_load_missing", []))
+            discard_names = []
+            if getattr(class_, "_tied_weights_keys", []):
+                discard_names.extend(getattr(class_, "_tied_weights_keys", []))
+            if getattr(class_, "_keys_to_ignore_on_load_missing", []):
+                discard_names.extend(getattr(class_, "_keys_to_ignore_on_load_missing", []))
```
Narsil, line 193: Actually I think we should just discard `_keys_to_ignore_on_load_missing`.

This should always be lists if defined, but indeed transformers recently switched everything to `_tied_weights_keys`.

Narsil:
Since it's a small fix I took the liberty of modifying the PR directly with the proposed fix. I tested on `facebook/opt-13.b` and it works.

PR #1096, opened 2023-10-04
Receive base64 encoded images for idefics.

### LINKED ISSUE #1095: Support other means of passing image to IdeficsImageProcessor
### Feature request

IdeficsImageProcessor expects the image to be passed only in the form of a URL from which it can [fetch the image](https://github.com/huggingface/text-generation-inference/blob/7a6fad6aac67d9bf21fe75c034a6bcab5dbd88d2/server/text_generation_server/models/custom_modeling/idefics_image_processing.py#L182).

It would be beneficial to allow accepting images as base64 encoded strings.

### Motivation

We want to use IDEFICS on data not directly available under the HTTP endpoint. We either need to set up another server to unpack an image from our DVC and make it available under some URL or deploy a custom docker container with patched image download logic.

### Your contribution

We can submit a PR to accept images as base64 encoded strings. Roughly looking like this:

```python
if isinstance(image_url_or_urls, str):
    decoded_bytes = None
    try:
        decoded_bytes = BytesIO(base64.b64decode(image_url_or_urls))
    except:
        print("Not a base64 encoded image")
    if decoded_bytes is not None:
        return Image.open(decoded_bytes)
    else:
        response = requests.get(image_url_or_urls, stream=True, headers=headers)
        response.raise_for_status()
        return Image.open(BytesIO(response.content))
```

Narsil:
# What does this PR do?

Fix #1095

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/models/custom_modeling/idefics_image_processing.py
+++ b/server/text_generation_server/models/custom_modeling/idefics_image_processing.py
@@ -35,6 +35,7 @@ from transformers.image_utils import (
     valid_images,
 )
 from io import BytesIO
+import base64
 import requests
 from transformers import TensorType, is_torch_available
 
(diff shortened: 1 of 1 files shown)


server/text_generation_server/models/custom_modeling/idefics_image_processing.py
```diff
@@ -194,10 +195,17 @@ def fetch_images(self, image_url_or_urls: Union[str, List[str]]):
...
+                content = response.content
+            else:
+                content = base64.b64decode(image)
+
             try:
-                image = Image.open(BytesIO(response.content))
+                image = Image.open(BytesIO(content))
                 # image.verify()
             except Exception:
                 raise ValueError(f"Could not load image from url {image_url_or_urls}")
```
leot13, line 211: If there is an exception, won't this make the url visible for anyone having access to the logs?
I think this may be undesirable for certain deployments

leot13 (commented):
Thanks for the improvement! Just one comment regarding the url display in the error message

server/text_generation_server/models/custom_modeling/idefics_image_processing.py
Narsil: Logs are supposed to be owned by the deployer. There's no need for TGI to censor any logs.

PR #1183, opened 2023-10-20
feat: paged attention v2

OlivierDehaene:


dongs0104:
Hi OlivierDehaene, I have a question about this and the flash attention kernel.
Why do you build an external kernel and why not use a pip installation?

OlivierDehaene:
@dongs0104, it's mainly to keep the final docker images small enough. Since we only use the attention/paged attention kernels, we do not want to bundle the rest as the image is already big.

For users that do not rely on the docker image, installing through pip is 100% fine.

RonanKMcGovern:
@OlivierDehaene does paged attention v2 work by default then? no additional flags or installs required? Thanks and thanks for your work on it.

OlivierDehaene:
Yes, as long as you use ghcr.io/huggingface/text-generation-inference:sha-12590fd or ulterior.

PR #1224, opened 2023-11-02
docs: set tag to `1.1`.

pcuenca:
# What does this PR do?

I could not find version `1.1.1` in ghcr.io, only saw `1.1` or `1.1.0`. Perhaps we could just omit it and have it use `latest`?

## Before submitting
      Pull Request section?
      to it if that's the case.


Narsil:
It does exist though now.

Also never EVER use latest please. Docker tends to never update itself even when newer versions are available making it impossible to know which version you are actually using (This is kind of an automated reply for us now when we have issues and people are using latest)

PR #1341, opened 2023-12-13
Text generation inference, fix offline

oOraph:
# What does this PR do?

Allows for Text Generation Inference to succeed in loading prefetched and cached private models if no token is provided at the time the text-generation-inference service is launched

### DIFF
--- a/server/text_generation_server/utils/hub.py
+++ b/server/text_generation_server/utils/hub.py
@@ -6,24 +6,29 @@ from loguru import logger
 from pathlib import Path
 from typing import Optional, List
 
-from huggingface_hub import HfApi, hf_hub_download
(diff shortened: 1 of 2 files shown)


server/text_generation_server/utils/hub.py
```diff
@@ -138,33 +179,33 @@ def download_weights(
 ) -> List[Path]:
     """Download the safetensors files from the hub"""
 
-    def download_file(filename, tries=5, backoff: int = 5):
-        local_file = try_to_load_from_cache(model_id, revision, filename)
+    def download_file(fname, tries=5, backoff: int = 5):
```
oOraph, line 182: Note: these var renaming (filename -> fname, start_time -> stime and i -> idx) are not mandatory, just renamed them to avoid pep warnings like "shadows var_name from outer scope"

Issue #1325, opened 2023-12-10
cannot import name 'cache_ops' from 'vllm'


iateadonut:
### System Info

$ text-generation-launcher --env
Runtime environment:
Target: x86_64-unknown-linux-gnu
Cargo version: 1.70.0
Commit sha: 3238c49121b02432bf2938c6ebfd44f06c5adc2f
Docker label: N/A
nvidia-smi:
Mon Dec 11 00:12:49 2023       
```
   +---------------------------------------------------------------------------------------+
   | NVIDIA-SMI 535.104.12             Driver Version: 535.104.12   CUDA Version: 12.2     |
   |-----------------------------------------+----------------------+----------------------+
   | GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
   | Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
   |                                         |                      |               MIG M. |
   |=========================================+======================+======================|
   |   0  NVIDIA GeForce RTX 4090        On  | 00000000:01:00.0 Off |                  Off |
   |  0%   40C    P8              32W / 450W |   8488MiB / 24564MiB |      0%      Default |
   |                                         |                      |                  N/A |
   +-----------------------------------------+----------------------+----------------------+
                                                                                            
   +---------------------------------------------------------------------------------------+
   | Processes:                                                                            |
   |  GPU   GI   CI        PID   Type   Process name                            GPU Memory |
   |        ID   ID                                                             Usage      |
   |=======================================================================================|
   |    0   N/A  N/A      1053      G   /usr/lib/xorg/Xorg                           18MiB |
   |    0   N/A  N/A      3068      C   ...vateGPT/conda_privateGPT/bin/python     8454MiB |
   +---------------------------------------------------------------------------------------+
```

$ cargo version
cargo 1.69.0 (6e9a83356 2023-04-12)

model: https://huggingface.co/mistralai/Mistral-7B-Instruct-v0.1

### Information

- [ ] Docker
- [X] The CLI directly

### Tasks

- [X] An officially supported command
- [ ] My own modifications

### Reproduction

$ text-generation-launcher --model-id mistralai/Mistral-7B-Instruct-v0.1 --port 8080

[...]

PR #1386, opened 2023-12-30
[fix] fix vllm import error

GaleSeLee:
# What does this PR do?
This PR is try to fix the vllm import error relaxed to the https://github.com/huggingface/text-generation-inference/issues/1325.


GaleSeLee:
According the discussion in https://github.com/huggingface/text-generation-inference/issues/1325. Close this pr.

PR #1395, opened 2024-01-03
feat: add mocked http request tests

drbh:
This PR implements two http request tests, and adds helper functions to test JSON requests and expected outputs.

```
test server::tests::test_echo_inputs_when_skip_generation ... ok
test server::tests::test_return_json_error_on_empty_inputs ... ok
```

Notes: 
- this pr adds `tower = "0.4.13"` so we can use the `tower::util::ServiceExt` trait to allow `oneshot` requests in the tests without starting an HTTP server. 
- `skip_generation` has been added to  `GenerateParameters` as a way to avoid making requests to the router. 

@Narsil and @OlivierDehaene please let me know what you think of this approach and if there is a better way to implement tests (the longer term goal is to improve TGI's API and these tests may be helpful when extending the API)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2820,6 +2820,7 @@ dependencies = [
  "tokenizers",
  "tokio",
  "tokio-stream",
+ "tower",
  "tower-http",
  "tracing",
  "tracing-opentelemetry",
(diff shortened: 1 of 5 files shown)


Narsil (commented):
Could you try not creating any surface just for tests ?

Tests are after all just for tests, they shouldn't ever require the code to change to accomodate for them.

Also adding new dependency is really a big thing imo, we should really refrain from doing so (especially since axum should provide everything we need already).

Issue #735, opened 2023-07-28
OpenAI API Wrapper


Ichigo3766:
### Feature request

Hi,

I was wondering if it would be possible to have a openai based api. 

### Motivation

Many projects have been built around openai api something similar to what vllm has and few others inference servers have. If TGI can have this, we can just swap the base url for any projects such as aider and many more and use them without any hastle of changing the code.

https://github.com/paul-gauthier/aider
https://github.com/AntonOsika/gpt-engineer
https://github.com/Significant-Gravitas/Auto-GPT

And many more.

For reference, vllm has a wrapper and text-generation webui has one too.

### Your contribution

discuss.

philschmid:
Hello @bloodsucker99, I am not sure that's possible on the server side since models have different prompts. 
So it might make sense to implement this on a client side, which converts the open AI schema (List of dicts) into a single prompt.

bloodsucker99:
Yea i kind of suspected that doing it on server side would not be possible :( 

Any chance you/anyone would be interested in building like a middle man for this? A python wrapper that just sits in middle would be cool

philschmid:
I had some time and started working on something. I will share the first version here. I would love to get feedback if you are willing to try out.

bloodsucker99:
Id love to try it out. Also possible to communicate over discord? Would make it much easier :)

philschmid:
Okay, I rushed out the first version. It is in a package i started called [easyllm](https://github.com/philschmid/easyllm/tree/main). 

Github:  https://github.com/philschmid/easyllm
Documentation: https://philschmid.github.io/easyllm/

The documentation also includes examples for streaming

## Example 

Install EasyLLM via pip:

```bash
pip install easyllm
```

Then import and start using the clients:

```python

from easyllm.clients import huggingface
from easyllm.prompt_utils import build_llama2_prompt

# helper to build llama2 prompt
huggingface.prompt_builder = build_llama2_prompt

response = huggingface.ChatCompletion.create(
    model="meta-llama/Llama-2-70b-chat-hf",
    messages=[
        {"role": "system", "content": "\nYou are a helpful assistant speaking like a pirate. argh!"},
        {"role": "user", "content": "What is the sun?"},
    ],
      temperature=0.9,
      top_p=0.6,
      max_tokens=256,
)

print(response)
```

bloodsucker99:
[...]

PR #1408, opened 2024-01-05
feat: supports openai chat completions API

drbh:
This PR adds support to make TGI a drop in replacement for OpenAI clients by exposing the same HTTP interface. 

Notes
- TGI inits a single model at startup so the `model` field is unused in that request.
- `max_tokens` and `stream` should work as expected but other params may not be (implemented or supported)

General approach
- fetch the `tokenizer_config` at startup from the hub
- pass `tokenizer_config` into `Infer` so we have it at request time
- use the `chat_template` on the config to format chat request
- parse jinja template and render chat string
- pass inputs into existing generate function
- wrap generation output in expected structure before returning

# How to test

### Streaming curl
```bash
curl localhost:3000/v1/chat/completions \
    -X POST \
    -d '{
  "model": "tgi",
  "messages": [
    {
      "role": "system",
      "content": "You are a helpful assistant."
    },
    {
      "role": "user",
      "content": "What is deep learning?"
    }
  ],
  "stream": true,
  "max_tokens": 20
}' \
    -H 'Content-Type: application/json'
```

It is also possible to use the `openai` python library and change the base url

###  🌊 STREAMING REQUEST
```python
from openai import OpenAI

# init the client but point it to TGI
client = OpenAI(
    base_url="http://localhost:3000/v1",
    api_key="not needed for a local LLM"
)

chat_completion = client.chat.completions.create(
    model="tgi",
    messages=[
        {"role": "system", "content": "You are a helpful assistant." },
        {"role": "user", "content": "What is deep learning?"}
    ],
    stream=True
)

# iterate and print stream
for message in chat_completion:
    print(message)

# ChatCompletionChunk(id='', choices=[Choice(delta=ChoiceDelta(content=' that', function_call=None, role='assistant', tool_calls=None), finish_reason=None, index=0, logprobs=None)], created=1704486761, model='', object='text_completion', system_fingerprint='')
```

### 🚗 SYNCHRONOUS REQUEST
```python
from openai import OpenAI

# init the client but point it to TGI
client = OpenAI(
    base_url="http://localhost:3000/v1",
    api_key="not needed for a local LLM"
)

chat_completion = client.chat.completions.create(
    model="tgi",
    messages=[
        {"role": "system", "content": "You are a helpful assistant." },
        {"role": "user", "content": "What is deep learning?"}
    ],
    stream=False
)

print(chat_completion)
# ChatCompletion(id='', choices=[Choice(finish_reason=None, index=0, logprobs=None, message=ChatCompletionMessage(content='\nDeep learning is a new field of research that has been gaining traction in the last ...', role='assistant', function_call=None, tool_calls=None))], created=1704486762, model='', object='text_completion', system_fingerprint='', usage=CompletionUsage(completion_tokens=100, prompt_tokens=76, total_tokens=176))
```


drbh:
@Narsil and @OlivierDehaene please let me know if any changes should be made!

router/src/infer.rs
```diff
@@ -666,3 +693,20 @@ impl InferError {
...
+#[derive(Debug, Error)]
+pub enum ChatTemplateError {
+    #[error("Template error: {0}")]
+    TemplateError(#[from] minijinja::Error),
+    #[error("Template not found")]
+    TemplateNotFound,
+}
+
+impl ChatTemplateError {
+    pub(crate) fn error_type(&self) -> &str {
```
Narsil, line 706: Why do we need this ?

drbh: thanks for catching, `ChatTemplateError` is unnecessary and has been removed in favor of adding a `InferError::TemplateError(_)`

router/src/server.rs
```diff
@@ -494,7 +515,146 @@ async fn generate_stream(
...
+    let generate_request = GenerateRequest {
+        inputs: inputs.to_string(),
+        parameters: GenerateParameters {
+            best_of: None,
+            temperature: None,
+            repetition_penalty: None,
+            top_k: None,
+            top_p: None,
+            typical_p: None,
+            do_sample: false,
```
OlivierDehaene, line 593: Should be `true`.

router/src/server.rs
```diff
@@ -494,7 +515,146 @@ async fn generate_stream(
...
+                .json_data(ChatCompletionChunk::new(
+                    stream_token.token.text,
+                    current_time,
+                    current_stream_count,
+                ))
+                .unwrap_or_else(|_| {
+                    println!("Failed to serialize ChatCompletionChunk");
+                    Event::default()
+                })
+        };
```
OlivierDehaene, line 632: Maybe this can be refactored so that the idx of the stream_token is part of the signature of the callback instead of relying on an atomic? It can also be part of the StreamResponse struct.

router/src/infer.rs
```diff
@@ -133,6 +138,24 @@ impl Infer {
...
+            .tokenizer_config
+            .chat_template
+            .as_ref()
+            .ok_or_else(|| {
+                InferError::TemplateError(minijinja::ErrorKind::TemplateNotFound.into())
+            })?;
+        env.add_template("_", chat_template)?;
+        env.get_template("_")?
+            .render(chat)
+            .map_err(InferError::TemplateError)
```
OlivierDehaene, line 156: Could we make this part of the struct instead?

router/src/server.rs
```diff
@@ -494,7 +515,146 @@ async fn generate_stream(
...
+                Json(ErrorResponse {
+                    error: err.to_string(),
+                    error_type: err.error_type().to_string(),
+                }),
+            ));
+        }
+    };
+
+    // poor man's token count (assumes that each character is a token)
+    let prompt_character_count: u32 = inputs.chars().count().try_into().unwrap_or_default();
```
OlivierDehaene, line 581: We have the true count when we validate the input. We could modify the InferResponse struct to contain it.

Michellehbn:
Noting we can probably close https://github.com/huggingface/text-generation-inference/issues/735 when this is done!

PR #1442, opened 2024-01-13
feat: adds phi model

drbh:
This PR adds basic modeling for phi-2 

run
```bash
text-generation-server \
    serve \
    microsoft/phi-2 \
    --revision 834565c23f9b28b96ccbeabe614dd906b6db551a
```

test
```bash
curl -s localhost:3000/generate \
   -X POST \
   -d '{"inputs":"What is Deep Learning?","parameters":{"max_new_tokens":20}}' \
   -H 'Content-Type: application/json' | jq .
# {
#   "generated_text": "\nDeep learning is a subset of machine learning that uses artificial neural networks to learn from data. These"
# }
```

notes 
- recently (~1 day ago) the Phi weights and model were updated to accommodate adding [GQA/MQA attention to the model.](https://github.com/huggingface/transformers/pull/28163) This impl expects the original model format so a fixed revision is required at the moment. 
- this PR only includes a basic implementation of the model and can later be extended for support Flash and Sharded versions as well as make use of better optimization

### DIFF
--- a/server/text_generation_server/models/__init__.py
+++ b/server/text_generation_server/models/__init__.py
@@ -58,6 +58,7 @@ try:
     from text_generation_server.models.idefics import IDEFICSSharded
     from text_generation_server.models.flash_mistral import FlashMistral
     from text_generation_server.models.flash_mixtral import FlashMixtral
+    from text_generation_server.models.flash_phi import FlashPhi
     from text_generation_server.utils.flash_attn import HAS_FLASH_ATTN_V2_CUDA
 
 except ImportError as e:
@@ -73,6 +74,7 @@ if FLASH_ATTENTION:
     __all__.append(IDEFICSSharded)
     __all__.append(FlashMistral)
     __all__.append(FlashMixtral)
+    __all__.append(FlashPhi)
 
 
 def get_model(
@@ -229,11 +231,22 @@ def get_model(
                 trust_remote_code=trust_remote_code,
             )
         
+    elif model_type == "phi":
+        if FLASH_ATTENTION:
+            return FlashPhi(
+                model_id,
+                revision,
+                quantize=quantize,
+                dtype=dtype,
+                trust_remote_code=trust_remote_code,
+                use_medusa=use_medusa,
+            )
+        else:
+            raise NotImplementedError(FLASH_ATT_ERROR_MESSAGE.format("Phi"))
+
     elif model_type == "phi-msft":
         if FLASH_ATTENTION:
-            raise NotImplementedError(FLASH_ATT_ERROR_MESSAGE.format("Flash Phi"))
-        elif sharded:
-            raise NotImplementedError(FLASH_ATT_ERROR_MESSAGE.format("Sharded Phi"))
+            raise NotImplementedError("Legacy phi-msft is not supported with Flash Attention")
         else:
             return Phi(
                 model_id,
--- a/server/text_generation_server/models/__init__.py
+++ b/server/text_generation_server/models/__init__.py
@@ -18,6 +18,7 @@ from text_generation_server.models.galactica import GalacticaSharded
 from text_generation_server.models.santacoder import SantaCoder
 from text_generation_server.models.t5 import T5Sharded
 from text_generation_server.models.gpt_neox import GPTNeoxSharded
+from text_generation_server.models.phi import Phi
 
 # The flag below controls whether to allow TF32 on matmul. This flag defaults to False
 # in PyTorch 1.12 and later.
@@ -221,6 +222,20 @@ def get_model(
             )
         else:
             return CausalLM(
+                model_id,
+                revision,
+                quantize=quantize,
+                dtype=dtype,
+                trust_remote_code=trust_remote_code,
+            )
+        
+    elif model_type == "phi-msft":
+        if FLASH_ATTENTION:
+            raise NotImplementedError(FLASH_ATT_ERROR_MESSAGE.format("Flash Phi"))
+        elif sharded:
+            raise NotImplementedError(FLASH_ATT_ERROR_MESSAGE.format("Sharded Phi"))
+        else:
+            return Phi(
                 model_id,
                 revision,
                 quantize=quantize,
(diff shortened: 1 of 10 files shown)


Narsil:
You need to implement the flash one, the non flash are not interesting anymore and just kept as legacy.

server/text_generation_server/models/custom_modeling/flash_phi_modeling.py
```diff
@@ -0,0 +1,420 @@
...
+
+        # NOTE: this is the main difference between Llama and Phi
+        # in llama the rotary embeddings are applied to the whole query and key.
+        # Phi uses PARTIAL rotary embeddings, which are applied to the first 32 dimensions
+        #
+        # Apply partial positional embeddings in place
+        self.rotary_emb(
+            query[:, :, :self.num_heads], kv[:, 0, :, :self.num_heads],
+            cos, sin
+        )
```
drbh, line 193: highlighting the most important lines here. Note only a portion of the query and key are rotary embedded.

drbh:
Hey @OlivierDehaene would you mind taking a look at this model (specifically flash phi)? The non flash is only for legacy purposes.


**Notes:**
I've copied the flash llama file and adjusted it to load the phi weights. Almost all layers are the same as llama except for the attention layer, Phi applies partial rotary embeddings (highlighted above). 

Outside of those changes there are a couple differences in weight naming and Phi uses bias. I think it makes sense to have Phi in a separate file rather than bloat the llama implementation. 

Please let me know if I should make any changes! Thanks 🙏

server/text_generation_server/models/__init__.py
```diff
@@ -227,6 +230,31 @@ def get_model(
...
+            return FlashPhi(
+                model_id,
+                revision,
+                quantize=quantize,
+                dtype=dtype,
+                trust_remote_code=trust_remote_code,
+                use_medusa=use_medusa,
+            )
+        else:
+            raise NotImplementedError(FLASH_ATT_ERROR_MESSAGE.format("Phi"))
```
OlivierDehaene, line 245: Could we also load it with the Phi class based on CausalLM? Or is the layout different? 
I'm not too familiar with Phi.

server/text_generation_server/models/custom_modeling/flash_phi_modeling.py
```diff
@@ -0,0 +1,420 @@
...
+        return _load_gqa(config, prefix, weights)
+    else:
+        if config.model_type == "baichuan":
+            return TensorParallelColumnLinear.load_qkv(
+                config,
+                prefix=f"{prefix}.W_pack",
+                weights=weights,
+                bias=True,
+            )
+        else:
```
OlivierDehaene, line 78:
```suggestion
```

server/text_generation_server/models/custom_modeling/flash_phi_modeling.py
```diff
@@ -0,0 +1,420 @@
...
+    TensorParallelRowLinear,
+    TensorParallelColumnLinear,
+    TensorParallelEmbedding,
+    PositionRotaryEmbedding,
+    TensorParallelHead,
+    get_linear,
+    FastLayerNorm,
+)
+
+class PhiConfig(PretrainedConfig):
```
OlivierDehaene, line 20: The config is missing some values like `partial_rotary_factor`.
See: https://huggingface.co/microsoft/phi-2/blob/main/configuration_phi.py#L136

server/text_generation_server/models/custom_modeling/flash_phi_modeling.py
```diff
@@ -0,0 +1,420 @@
...
+        self.num_heads = config.num_attention_heads
+        self.hidden_size = config.hidden_size
+        self.head_size = self.hidden_size // self.num_heads
+
+        self.rotary_emb = PositionRotaryEmbedding.static(
+            config=config,
+            dim=self.num_heads,
+            base=config.rope_theta,
+            device=weights.device,
+        )
```
OlivierDehaene, line 130:
```suggestion
        self.rotary_dim = int(config.partial_rotary_factor * self.head_size)
        self.rotary_emb = PositionRotaryEmbedding.static(
            config=config,
            dim=self.rotary_dim,
            base=config.rope_theta,
            device=weights.device,
        )
```

Not 100% sure but this is what I saw when skiming through the phi code inside the repo.

server/text_generation_server/models/custom_modeling/flash_phi_modeling.py
```diff
@@ -0,0 +1,420 @@
...
+        query = query.view(-1, self.num_heads, self.head_size)
+        kv = kv.view(-1, 2, self.num_key_value_heads, self.head_size)
+
+        # NOTE: this is the main difference between Llama and Phi
+        # in llama the rotary embeddings are applied to the whole query and key.
+        # Phi uses PARTIAL rotary embeddings, which are applied to the first 32 dimensions
+        #
+        # Apply partial positional embeddings in place
+        self.rotary_emb(
+            query[:, :, :self.num_heads], kv[:, 0, :, :self.num_heads],
```
OlivierDehaene, line 191:
```suggestion
            query[:, :, :self.rotary_dim], kv[:, 0, :, :self.rotary_dim],
```

server/text_generation_server/models/custom_modeling/flash_phi_modeling.py
```diff
@@ -0,0 +1,420 @@
...
+            else lambda x: torch.nn.functional.gelu(
+                x,
+                approximate="tanh"
+                if act in ["gelu_fast", "gelu_pytorch_tanh"]
+                else "none",
+            )
+        )
+
+        # llama weights are up_proj and down_proj and bias=False
+        self.gate_up_proj = TensorParallelRowLinear.load(
```
OlivierDehaene, line 246: Can you rename this layer to `up_proj`? It's a bit weird to see the "gate" naming when we are not gating.

server/text_generation_server/models/__init__.py
drbh: thanks for the feedback! 

yes, Phi can be loaded as a CausalLM with the `4.37.1` version of transformers. In my last PR I bumped the version and tested that Phi works on my local cpu

OlivierDehaene (approved):
Great thanks!

PR #1454, opened 2024-01-18
feat: conditionally toggle chat on invocations route

drbh:
This PR adds support for reading the `OAI_ENABLED` env var which will changes the function called when the `/invocations` is called.

If `OAI_ENABLED=true` the `chat_completions` method is used otherwise it defaults to `compat_generate`.

example running the router
```bash
OAI_ENABLED=true \
  cargo run -- \
  --tokenizer-name mistralai/Mistral-7B-Instruct-v0.2
```

example request
```bash
curl localhost:3000/invocations \
    -X POST \
    -d '{ "model": "tgi", "messages": [ { "role": "user", "content": "What is the IP address of the Google DNS servers?" } ], "stream": false, "max_tokens": 20, "logprobs": true, "seed": 0 }' \
    -H 'Content-Type: application/json' | jq 
```

**please let me know if any naming changes are needed or if any other routes need similar functionality.

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -708,6 +708,7 @@ pub async fn run(
     ngrok_authtoken: Option<String>,
     ngrok_edge: Option<String>,
     tokenizer_config: HubTokenizerConfig,
+    chat_enabled_api: bool,
 ) -> Result<(), axum::BoxError> {
     // OpenAPI documentation
     #[derive(OpenApi)]
@@ -871,7 +872,7 @@ pub async fn run(
         .route("/metrics", get(metrics));
 
     // Conditional AWS Sagemaker route
-    let aws_sagemaker_route = if std::env::var("OAI_ENABLED").map_or(false, |val| val == "true") {
+    let aws_sagemaker_route = if chat_enabled_api {
         Router::new().route("/invocations", post(chat_completions)) // Use 'chat_completions' for OAI_ENABLED
     } else {
         Router::new().route("/invocations", post(compat_generate)) // Use 'compat_generate' otherwise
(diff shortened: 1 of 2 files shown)


router/src/server.rs
```diff
@@ -856,25 +856,32 @@ pub async fn run(
...
-        // Inference API health route
-        .route("/", get(health))
-        // AWS Sagemaker health route
         .route("/ping", get(health))
-        // Prometheus metrics route
-        .route("/metrics", get(metrics))
+        .route("/metrics", get(metrics));
+
+    // Conditional AWS Sagemaker route
+    let aws_sagemaker_route = if std::env::var("OAI_ENABLED").map_or(false, |val| val == "true") {
```
Narsil, line 874: Can you move this into the CLI argument from clap ? Makes truthful values equality. more general (and easier to use for users).

`MESSAGES_API_ENABLED` is maybe a bit more explicit.
Personnally, `CHAT_API_ENABLED` seems a bit more apprioriate. 

Both are fine to me. 
at the very least `OPENAI_ENABLED` would be more explicit than OAI imho.

@OlivierDehaene any preference on the name ?

drbh: thanks for the quick review, will prefer clap in the next commit 👍

and yea I agree the `CHAT_API_ENABLED` is the most descriptive.... chat implies a conversation where messages do not, but totally open to whatever names best

Narsil (approved):
LGTM. 

I just think we should document the behavior somewhere.

@philschmid any idea of where that will be best documented for SM users ?

philschmid:
> LGTM.
> 
> I just think we should document the behavior somewhere.
> 
> @philschmid any idea of where that will be best documented for SM users ?

Good question. Should add a dedicated documentation page in TGI on how to deploy to sagemaker? Where we then can include this?

Narsil:
I'm fine with it @drbh Could you create a new page in the doc ?  This can be done in a follow-up PR.

PR #1463, opened 2024-01-20
Enable padding before sharding for tp embedding for non-divisible embedding tables.

wanderingai:
# What does this PR do?

This PR addresses an implementation issue with the `TensorParallelEmbedding`, which shards the embedding table via floor division without padding it to the next multiple of the world size to ensure divisibility. This affects models with non-divisible vocab sizes e.g. [NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO](https://huggingface.co/NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO) where the critical special tokens, `<|im_start|>` and `<|im_end|>` are out of the bounds of the computed `max_id`.

The fix has been tested with the aforementioned model and verified with the model author(s).

## Before submitting
- [x] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests), Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag members/contributors who may be interested in your PR.

### DIFF
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -499,13 +499,16 @@ class TensorParallelRowLinear(SuperLayer):
 class TensorParallelEmbedding(nn.Module):
     def __init__(self, prefix: str, weights, reduce=True):
         super().__init__()
-        weight = weights.get_partial_sharded(f"{prefix}.weight", dim=0)
-        num_embeddings = weights.get_shape(f"{prefix}.weight")[0]
-
         process_group = weights.process_group
 
         world_size = process_group.size()
         rank = process_group.rank()
+
+        weight, margin = weights.get_padded_sharded(
+            f"{prefix}.weight", dim=0, pad_multiple=world_size
+        )
+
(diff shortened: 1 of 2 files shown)


Narsil:
Hi, thanks for the PR, but this is unlikely to fix anything, padding is not necessary for this.

Having talked through other channels, the issues seems to stem from lack of system prompting, which puts Nous model out of domain.
There are no issues as soon as a system template is being used.

If there was a bug in `TensorParallelEmbedding` , which should identify it through a simple reproducer and fix it then. 
The padding solution is probably not necessary since there is already an null_token being already appended here.

Narsil:
Thanks a lot for the recap !

PR #1484, opened 2024-01-25
[WIP] FP8 support.

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/docs/source/basic_tutorials/launcher.md
+++ b/docs/source/basic_tutorials/launcher.md
@@ -66,7 +66,6 @@ Options:
           - bitsandbytes:     Bitsandbytes 8bit. Can be applied on any model, will cut the memory requirement in half, but it is known that the model will be much slower to run than the native f16
           - bitsandbytes-nf4: Bitsandbytes 4bit. Can be applied on any model, will cut the memory requirement by 4x, but it is known that the model will be much slower to run than the native f16
           - bitsandbytes-fp4: Bitsandbytes 4bit. nf4 should be preferred in most cases but maybe this one has better perplexity performance for you model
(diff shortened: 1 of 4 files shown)


flozi00:
Would it be worth to discuss about transformers engine implementation since it's compatible with ada lovelace and higher ?
The comment in your code says it needs Hopper+

Narsil:
> Would it be worth to discuss about transformers engine implementation since they say it's compatible with ada lovelace and higher ? The comment in your code says it needs Hopper+

Well it works whereever FP8 works, H100 are the most common cluster ones I think.

> 
> source: https://docs.nvidia.com/deeplearning/transformer-engine/user-guide/index.html#:~:text=Support%20for%20FP8%20on%20NVIDIA%20Hopper%20and%20NVIDIA%20Ada%20GPUs
> 
> the implementation seems to be very easy, as its already done in accelerate https://github.com/huggingface/accelerate/blob/7aafa25673f4324cbcaeef508a13e09576f12dc4/src/accelerate/utils/transformer_engine.py#L30-L41

The implementation is silently not done with dimensions are not 16 aligned, this is not acceptable in TGI.

There are 2 issues with the current PR implem.
Every call packs f16 buffer before calling gemm, that's a slowdown for everycall, + the 16 padding.

That's what makes it slow, and TransformerEngine is just as slow: https://github.com/NVIDIA/TransformerEngine/blob/main/transformer_engine/pytorch/module/linear.py#L91C13-L92

So afaik there's no win to using transformer engine over this implementation.
The real work will be removing those unecessary ops (unpacking and padding).

And since LLM mostly operate at seqlen < 16 (during decode this is pretty much batch_size), we also most likely need a customized gemv to make things faster (This limitation on size is very suprising to me)

flozi00:
The pytorch lab implementation will have inference support later

https://github.com/pytorch-labs/float8_experimental/issues/187#issue-2084480478

I am also checking the tensorrt llm implementation of fp8

Issue #1460, opened 2024-01-19
Problem loading model with GPTNeoX architecture (weight gpt_neox.layers.0.attention.rotary_emb.inv_freq does not exist)


ZQ-Dev8:
### System Info

- System Info: 4xA6000 Workstation
- Full command line causing issues: `docker run --rm --entrypoint /bin/bash -itd --name "traclm-v1-3b-instruct" -v "path/to/folder":/data --gpus '"device=3"' -p 172.20.158.30:8082:80 ghcr.io/huggingface/text-generation-inference:latest`
- OS version: Ubuntu 22.04 LTS
- Rust version: rustc 1.73.0 (cc66ad468 2023-10-03)
- Model being used: Local finetune of `togethercomputer/RedPajama-INCITE-Base-3B-v1` with `teknium/GPT4-LLM-Cleaned`, created using the HF Trainer **without flash attention** (could this be the problem?)
- Hardware used: see below for nvidia-smi output:
```
+---------------------------------------------------------------------------------------+
| NVIDIA-SMI 535.129.03             Driver Version: 535.129.03   CUDA Version: 12.2     |
|-----------------------------------------+----------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |         Memory-Usage | GPU-Util  Compute M. |
|                                         |                      |               MIG M. |
|=========================================+======================+======================|
|   0  NVIDIA RTX A6000               Off | 00000000:19:00.0 Off |                  Off |
| 30%   38C    P8              34W / 300W |     12MiB / 49140MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   1  NVIDIA RTX A6000               Off | 00000000:1A:00.0 Off |                  Off |
| 30%   49C    P8              24W / 300W |  47642MiB / 49140MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   2  NVIDIA RTX A6000               Off | 00000000:67:00.0 Off |                  Off |
| 30%   55C    P8              27W / 300W |    968MiB / 49140MiB |      0%      Default |
|                                         |                      |                  N/A |
+-----------------------------------------+----------------------+----------------------+
|   3  NVIDIA RTX A6000               Off | 00000000:68:00.0 Off |                  Off |
[...]

PR #1498, opened 2024-01-26
GPTNeoX: Use static rotary embedding

dwyatte:
# What does this PR do?

`transformers` 4.35 removed rotary embeddings from GPTNeoX's weights ([link to line diff](https://github.com/huggingface/transformers/commit/253f9a3f9716d08a81fb305fe71f983122eb608b#diff-0e2a05d86c82e96f516db8c14070ceb36f53ca44c6bc21a9cd92ad2e777b9cf1R298)). This applies the same fix as https://github.com/huggingface/text-generation-inference/pull/793 which generates them on-the-fly using the appropriate value from the config file

Fixes https://github.com/huggingface/text-generation-inference/issues/1460

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@OlivierDehaene OR @Narsil

### DIFF
--- a/server/text_generation_server/models/custom_modeling/flash_neox_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_neox_modeling.py
@@ -98,8 +98,11 @@ class FlashNeoxAttention(torch.nn.Module):
             )
         self.num_heads = self.num_heads // weights.process_group.size()
 
-        self.rotary_emb = PositionRotaryEmbedding.load(
-            config=config, prefix=f"{prefix}.rotary_emb", weights=weights
+        self.rotary_emb = PositionRotaryEmbedding.static(
+            config=config,
+            dim=self.head_size,
+            base=config.rotary_emb_base,
+            device=weights.device,
         )
 
         self.softmax_scale = self.head_size ** (-0.5)
--- a/server/text_generation_server/models/custom_modeling/flash_neox_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_neox_modeling.py
@@ -91,6 +91,8 @@ class FlashNeoxAttention(torch.nn.Module):
         self.hidden_size = hidden_size
         self.head_size = hidden_size // num_heads
 
+        self.rotary_dim = int(config.rotary_pct * self.head_size)
+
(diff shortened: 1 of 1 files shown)


drbh:
Hi @dwyatte thank you for the contribution, unfortunately this change appears to break the `neox` tests. 

tested locally with:
```bash
cd integration-tests/
pytest -s -vv models/test_flash_neox.py
```
**note you may have to comment out the `skip` flags for some tests

would you be able to review these changes and ensure they pass the tests. Thank you!!

dwyatte:
@drbh thanks for the reply

Any chance there is a guide to developing using TGI's docker image to manage all dependencies when running the integration tests? I've gotten pretty far on compiling everything but it seems like a never ending rabbit hole (on to Flash Attention now and it even recommended using TGI's image)

```
2024-01-30T00:04:51.061971Z  WARN text_generation_launcher: Could not import Flash Attention enabled models: Flash Attention is not installed.
Use the official Docker image (ghcr.io/huggingface/text-generation-inference:latest) or install flash attention with `cd server && make install install-flash-attention`
```

I can probably mount my code / integration tests, but I'm probably missing something easy (e.g., just reading through source code, maybe setting the env var `DOCKER_IMAGE`)

dwyatte:
Ok @drbh [871e5e7](https://github.com/huggingface/text-generation-inference/pull/1498/commits/871e5e73389091cd5d38507c64d4a8609cf5e5e3) fixes the actual rotary embedding logic (thanks for flagging) but I suspect there might still be failures in the local tests due to stale snapshots

Wondering if we should update these as part of the PR

drbh:
awesome thank you for resolving @dwyatte. I actually think those tests have been flakey for some time now and neox is not a current priority, so we can continue to skip them for now. Overall the PR looks good and has reasonable responses when tested locally but I think we should get a quick second look before approving.

@Narsil what do you think about approving when tests are flakey/skipped?

Thank you both!

Narsil:
Thanks a lot for this contribution.

We' ll take the contribution as-is. It possibly breaks some stuff but overall is more aligned with other modeling code, therefore i' m ok to add it.
Neox non flash is not tested, but this is neox flash

PR #1514, opened 2024-02-01
ROCm AWQ support

IlyasMoutawwakil:
# What does this PR do?

This PR adds the possibility to run AWQ models with Exllama/GPTQ kernels, specifically for ROCm devices that support Exllama kernels but not AWQ's GEMM. 

This is done by :
- un-packing, reordering and re-packing AWQ weights when `--quantize gptq` but the model's `quant_method=awq`. 
- avoiding overflows when adding 1 to zeros in exllama and triton.

Ref: https://github.com/casper-hansen/AutoAWQ/pull/313

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


IlyasMoutawwakil:
Tested with (Llama-7b + awq + sharding) on an MI250
```
ilyas@hf-amd-mi250-dev:~/text-generation-inference$ docker run --device /dev/dri/ --device /dev/kfd/ --shm-size 1g -p 9999:80 -v $volume:/data tgi-rocm-awq:latest --model-id TheBloke/Llama-2-7B-AWQ --quantize awq  --sharded true --num-shard 2
```

IlyasMoutawwakil:
Also tested with Triton Kernel
```
docker run --env DISABLE_EXLLAMA="True" --device /dev/dri/ --device /dev/kfd/ --shm-size 1g -p 9999:80 -v $volume:/data tgi-rocm-awq:latest --model-id TheBloke/Llama-2-7B-AWQ --quantize awq --sharded true --num-shard 2
```

server/text_generation_server/utils/awq/pack_utils.py
```diff
@@ -0,0 +1,146 @@
+import torch
```
Narsil, line 1: This is great, should probably a standalone conversion tool somewhere, no ?

server/text_generation_server/utils/layers.py
```diff
@@ -349,14 +350,20 @@ def get_linear(weight, bias, quantize):
...
             )
-        linear = WQLinear(
-            w_bit=bits,
-            group_size=groupsize,
-            qweight=qweight,
-            qzeros=qzeros,
-            scales=scales,
-            bias=bias is not None,
-        )
+        if HAS_AWQ:
```
Narsil, line 354: I'm really against that kind of hidden control flow.

Let's make it trivial for users to convert from AWQ to GPTQ externally, and then actually use GTPQ.
Making unasked for, on-the-fly conversions is really not ok.

Narsil (commented):
I love the idea of adding more support for Rocm, but the way of achieving it right now is deceiving to users and not really achieving what we are supposed to.

Adding tooling to do AWQ->GPTQ is great but we should not do it on behalf of users.

Narsil:
Screenshot 2024-02-01 at 14 46 29


Screenshot 2024-02-01 at 14 51 21


Here are initial benchmarks on Mistral 7b. There is a latency improvement at BS=1, but the throughput is reduced for this PR.

server/exllama_kernels/exllama_kernels/cuda_func/q4_matmul.cu
```diff
@@ -85,15 +85,15 @@ __global__ void q4_matmul_kernel
             if constexpr (use_half2)
             {
                 half2 w_scale = w_scales_.item_half2half2(group, w_column);
-                uint32_t w_zero = w_zeros_.item(group, w_column) + 1;
+                uint32_t w_zero = (w_zeros_.item(group, w_column) + 1) & 0x0F;
```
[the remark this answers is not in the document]
IlyasMoutawwakil: the problem is in GPTQ's standard of storing zeros-1 instead of original zeros, ref discussion
https://github.com/AutoGPTQ/AutoGPTQ/pull/484#discussion_r1427449006

server/text_generation_server/utils/layers.py
IlyasMoutawwakil: I see, does using exllama as a backend for awq exclusively on rocm make more sense ?

IlyasMoutawwakil:
Thank you @Narsil for the comments and the benchmark
can I get some details about the benchmark you shared, is it a comparison between llm-awq's GEMM (default behavior) vs Exllama (when llm-awq is unavailable) or Triton ? or GEMM on main vs this PR ?

server/text_generation_server/utils/layers.py
Narsil: I'm not happy of introducing a rocm vs nvidia difference either.

Don' t you think we could have instead a good error message when trying to load AWQ on rocm, and the error message includes a trivial way to use that model anyway.

```
text-generation-launcher --model-id XXX-awq --quantize awq
# Error ! AWQ on Rocm is not supported directly, you can use the GPTQ quantization to use them
text-generation-launcher --model-id XXX-awq --quantize gptq
```

For instance.Wdyt ? It seems quite obvious to users.It might be useful on nvidia targets too, an we can keep the almost transparent feeling.

Narsil:
> Thank you @Narsil for the comments and the benchmark can I get some details about the benchmark you shared, is it a comparison between llm-awq's GEMM (default behavior) vs Exllama (when llm-awq is unavailable) or Triton ? or GEMM on main vs this PR ?

Yes it' s mistral-instruct-v0.2 7B llm-awq VS your branch (without AWQ so running the conversion).

server/text_generation_server/utils/layers.py
Narsil: We *do* need to explicit log about the conversion happening though.

server/exllama_kernels/exllama_kernels/cuda_func/q4_matmul.cu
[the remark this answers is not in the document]
Narsil: Oh it's done upstream: https://github.com/AutoGPTQ/AutoGPTQ/blob/main/autogptq_extension/exllama/cuda_func/q4_matmul.cu#L90

Fine by me then.

server/text_generation_server/utils/layers.py
Narsil: We can use this flag https://huggingface.co/TheBloke/Mistral-7B-Instruct-v0.2-AWQ/blob/main/config.json#L24 to do the conversion on the fly.

Narsil:
Ignore the failing test it's unrelated to anything.

IlyasMoutawwakil:
Yes the numbers make sense since they fall in the 10-90 `total_seq_len` (`batch_size` * `seq_len`) range.
This benchmark https://github.com/AutoGPTQ/AutoGPTQ/pull/484#issue-2042020854 shows when each kernel is best.

Narsil (approved):
LGTM.
Let's wait for the tests (failure are still quite odd).

PR #1533, opened 2024-02-05
feat: use existing add_generation_prompt variable from config in temp…

drbh:
This PR adds support to read the `add_generation_prompt` from the config and use it in the chat template. If `add_generation_prompt` does not exist we default to false

### DIFF
--- a/router/src/infer.rs
+++ b/router/src/infer.rs
@@ -807,22 +807,14 @@ mod tests {
             ],
             bos_token: Some("[BOS]"),
             eos_token: Some("[EOS]"),
-            add_generation_prompt: false,
+            add_generation_prompt: true,
         };
 
         let result = tmpl.unwrap().render(chat_template_inputs).unwrap();
 
         assert_eq!(
             result,
-            r#"### User:
-Hi!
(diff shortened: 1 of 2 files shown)


router/src/infer.rs
```diff
@@ -198,6 +198,7 @@ impl Infer {
                 messages,
                 eos_token: eos_token.as_deref(),
                 bos_token: bos_token.as_deref(),
+                add_generation_prompt: true,
```
Narsil, line 201: Why is it sometimes `true`, sometimes `false`?

drbh: the `true` values will be used in all cases at runtime. The `false`s are only used in the existing template tests (and those tests all do not have `add_generation_prompt` so both true and false will return the same results).

In terms of updating the tests/add a new one I can push to this PR or maybe follow up with another small PR to avoid this blocking anything. lmk whats best! thank you!

drbh: ***update:

pushed new test and updated old ones

Narsil (approved):

PR #1539, opened 2024-02-08
Outlines guided generation

drbh:
This WIP PR starts to add grammar support via outlines, currently this PR supports very simple regex grammars and does not optimize for precompiling or caching grammar fsm's. 

todo:
- [X] add simple outlines guidance to `NextTokenChooser`
- [X] update protos for grammar
- [X] update generation params API
- [X] constrain simple grammar
- [ ] support parsing more complex grammar into fsm
- [ ] support all outline support grammar types
- [ ] explore optimizations to avoid recompiling grammars

guided request
```bash
curl -s 'http://localhost:3000/generate' \
--header 'Content-Type: application/json' \
--data-raw '{
    "inputs": "make an email for david: \n",
    "parameters": {
        "max_new_tokens": 6,
        "grammar": "[\\w-]+@([\\w-]+\\.)+[\\w-]+"
    }
}' | jq
```
response
```json
{
  "generated_text": "david@example.com"
}
```

unguided request
```bash
curl -s 'http://localhost:3000/generate' \
--header 'Content-Type: application/json' \
--data '{
    "inputs": "make an email for david: \n",
    "parameters": {
        "max_new_tokens": 6
    }
}' | jq
```
response
```json
{
  "generated_text": "    email = 'david"
}
```


drbh:
updates:

 - support parsing more complex grammar into fsm
 - support all (serializable) outline support grammar types 
 - explore optimizations to avoid recompiling grammars (rely on [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache))
 - add grammar to all token choosers (should work with all/most model architectures) 

JSON schemas are supported and can be used like:

```bash
 curl -s 'http://localhost:3000/generate' \
--header 'Content-Type: application/json' \
--data '{
    "inputs": "info: david holtz like trees and has two cats. ",
    "parameters": {
        "max_new_tokens": 100,
        "grammar": {
            "$id": "https://example.com/person.schema.json",
            "$schema": "https://json-schema.org/draft/2020-12/schema",
            "title": "Person",
            "type": "object",
            "properties": {
                "firstName": {
                    "type": "string",
                    "description": "The person'\''s first name."
                },
                "lastName": {
                    "type": "string",
                    "description": "The person'\''s last name."
                },
                "hobby": {
                    "description": "The person'\''s hobby.",
                    "type": "string"
                },
                "numCats": {
                    "description": "The number of cats the person has.",
                    "type": "integer",
                    "minimum": 0
                }
            },
            "required": ["firstName", "lastName", "hobby", "numCats"]
        }
    }
}' | jq .
```

response
```json
{
  "generated_text": "{\"firstName\": \"david\", \"hobby\": \"trees\", \"lastName\": \"holtz\", \"numCats\": 2}"
}
```

regex strings are still supported as well 

```bash
curl --location 'http://localhost:3000/generate' \
--header 'Content-Type: application/json' \
--data-raw '{
    "inputs": "name: david. email:  ",
    "parameters": {
        "max_new_tokens": 20,
        "grammar": "[\\w-]+@([\\w-]+\\.)+[\\w-]+"
    }
}'
```

```json
{
  "generated_text":"david@example.com.phone_number_1.1234567890.phone_"
}
```

notes:

building the from the grammar FSM is very computationally expensive and is required for the first generation. Wait times can be ~10 seconds with complex grammars; this performance impact (along with other thing) need to be taken into account before adding this feature

integration-tests/conftest.py
```diff
@@ -367,11 +367,18 @@ def docker_launcher(
 @pytest.fixture(scope="module")
 def generate_load():
     async def generate_load_inner(
-        client: AsyncClient, prompt: str, max_new_tokens: int, n: int
+        client: AsyncClient,
+        prompt: str,
+        max_new_tokens: int,
+        n: int,
+        **kwargs,
```
Narsil, line 374: Let's avoid kwargs if possible.

server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -910,9 +911,9 @@ def generate_token(
                 # Copy batch.input_ids to prefill_token_indices
                 if prefill_logprobs:
                     if len(batch) > 1:
-                        prefill_tokens_indices[
-                            out_start_index : out_end_index - 1
-                        ] = batch.input_ids[start_index + 1 : start_index + out_length]
+                        prefill_tokens_indices[out_start_index : out_end_index - 1] = (
```
Narsil, line 914: okay we're going to need to force lint everything with some standard instead of using each other's editor's default :)

proto/generate.proto
```diff
@@ -70,6 +70,10 @@ message NextTokenChooserParameters {
     float frequency_penalty = 9;
     /// token watermarking using "A Watermark for Large Language Models"
     bool watermark = 8;
+    /// grammar (applied if not empty)
+    string grammar = 10;
+    /// fsm_grammar_state
+    uint32 fsm_grammar_state = 11;
```
Narsil, line 76: I'm not sure we should pass the grammar_state here. It's always 0 when it's starting out, right ?

Adding anything here adds more work on other backends, so we should keep it to a strict minimum.

server/text_generation_server/utils/tokens.py
```diff
@@ -350,6 +394,19 @@ def __call__(
...
             speculative_ids = None
 
+        # advance the grammar state
+        if self.grammar_processor is not None:
+            for i in range(len(self.fsm_grammar_states)):
+                try:
+                    self.fsm_grammar_states[i] = self.grammar_processor.advance(
+                        next_ids[i].item(), self.fsm_grammar_states[i], self.grammars[i]
+                    )
+                except:
```
Narsil, line 404: Oops :)

server/text_generation_server/utils/tokens.py
```diff
@@ -350,6 +394,19 @@ def __call__(
         else:
             speculative_ids = None
 
+        # advance the grammar state
+        if self.grammar_processor is not None:
+            for i in range(len(self.fsm_grammar_states)):
+                try:
+                    self.fsm_grammar_states[i] = self.grammar_processor.advance(
+                        next_ids[i].item(), self.fsm_grammar_states[i], self.grammars[i]
+                    )
```
Narsil, line 403: The whole point of HeteregenousProcessor is to processor all the tokens at once, without making any CPU calls.

This `.item()` defeats the purpose.

I don't have any good ideas aside from moving this bit to the CPU loop in `causal_lm`/ `flash_causal_lm`.
Which is probably error prone.

@OlivierDehaene Do you have better suggestions ?

server/text_generation_server/utils/tokens.py
```diff
@@ -372,10 +429,16 @@ def filter(self, indices):
...
+        new_grammars = []
+        new_fsm_grammar_states = []
+        for i in indices:
+            new_grammars.append(self.grammars[i])
+            new_fsm_grammar_states.append(self.fsm_grammar_states[i])
+
+        self.grammars = new_grammars
+        self.fsm_grammar_states = new_fsm_grammar_states
+
+        self.choice = Greedy()
```
Narsil, line 441: Where did filter(indices) go ?

Narsil (commented):
Nice PR overall.

I've seen weird issues with the grammar leading out of regex returns (I'm guessing it has to do with the llama hack).

Narsil:
> explore optimizations to avoid recompiling grammars (rely on [lru_cache](https://docs.python.org/3/library/functools.html#functools.lru_cache))

Seems to works great.
Let's roll as-is for performance regarding the compilation we'll investigate later how to optimize this. The ideal situation would be to use the tokenization workers to do the compilation and send the compiled objects directly to the python backend (so we don't overload the python with CPU cycles which would slow down the GPUS)

server/text_generation_server/utils/logits_process.py
```diff
@@ -464,3 +468,147 @@ def filter(self, indices):
...
+            return fsm_grammar_state
+
+        if grammar == "" or grammar is None:
+            return fsm_grammar_state
+
+        fsm = self.compile_fsm(grammar, self.tokenizer)
+        return fsm.next_state(fsm_grammar_state, next_token_id)
+
+    @lru_cache(maxsize=32, typed=True)
+    def compile_fsm(self, schema, tokenizer):
```
Narsil, line 581: This can probably be put at the top level so it's written only once.

server/text_generation_server/utils/logits_process.py
```diff
@@ -464,3 +468,147 @@ def filter(self, indices):
...
+        start_time = time.time()
+        tokenizer = self.adapt_tokenizer(tokenizer)
+        is_json_string = schema.startswith("{") and schema.endswith("}")
```
Narsil, line 584: Seems a bit brittle no ?

server/text_generation_server/utils/tokens.py
```diff
@@ -78,13 +98,20 @@ def __call__(self, input_ids, scores):
 
         next_id = self.choice(scores[-1]).view(1, 1)
 
+        if self.grammar_processor is not None:
+            next_state = self.grammar_processor.advance(
+                next_id.item(), self.fsm_grammar_state, self.grammar
```
Narsil, line 103: Same remark here for the `.item()` although here it's less critical since it's already calling all the processors serially which is quite slow already.

server/text_generation_server/utils/logits_process.py
```diff
@@ -464,3 +468,147 @@ def filter(self, indices):
...
+        for i in range(len(logits)):
+            if fsm_grammar_states[i] == -1:
+                # todo mask for only eos token
+                continue
+
+            # if grammar is '' or None, return the greedy token
+            if grammars[i] == "" or grammars[i] is None:
+                continue
+
+            fsm = self.compile_fsm(grammars[i], self.tokenizer)
```
Narsil, line 562: This should be done in the `init` IMO.

server/text_generation_server/utils/logits_process.py
```diff
@@ -464,3 +468,147 @@ def filter(self, indices):
...
+        return self
+
+
+class HeterogeneousGrammarLogitProcessor(LogitsProcessor):
+    def __init__(self, tokenizer, device):
+        self.device = device
+        self.tokenizer = tokenizer
+
+    def __call__(self, input_ids: torch.Tensor, logits, fsm_grammar_states, grammars):
+        for i in range(len(logits)):
```
Narsil, line 553:
```suggestion
        for i in range(logits.shape[0]):
```

NIT: Feels slightly more obvious what we're doing here (iterating over batch size, not logits directly)

server/text_generation_server/utils/logits_process.py
```diff
@@ -464,3 +468,147 @@ def filter(self, indices):
...
+                # todo mask for only eos token
+                continue
+
+            # if grammar is '' or None, return the greedy token
+            if grammars[i] == "" or grammars[i] is None:
+                continue
+
+            fsm = self.compile_fsm(grammars[i], self.tokenizer)
+            allowed_tokens = fsm.allowed_token_ids(fsm_grammar_states[i])
+            mask = torch.full((logits.shape[-1],), -math.inf, device=self.device)
```
Narsil, line 564: WE could generate that at the start of the loop and reuse it over and over, no?

server/text_generation_server/models/flash_causal_lm.py
drbh: ahh I totally agree we should standardize our formatters. Currently I'm using Black out of the box. Is there an existing config I should use?

server/text_generation_server/utils/logits_process.py
```diff
@@ -464,3 +470,127 @@ def filter(self, indices):
...
+
+    @staticmethod
+    @lru_cache(maxsize=32, typed=True)
+    def _cached_compile_fsm(schema, tokenizer):
+        start_time = time.time()
+        try:
+            json.loads(schema)  # check if schema is a valid json
+            schema = build_regex_from_object(schema)  # convert schema to regex
+        except json.JSONDecodeError:
+            pass
```
OlivierDehaene, line 516: Is this a try catch to support schemas that are already a regex? If so can you add a comment?
It could be possible to move this compile to the router with PyO3 if this takes too much time.

drbh: yes it is and just added a comment in the latest commit. I agree we should move the compilation out of the server and into the router. 

I like the idea of using PyO3 and seems relatively straight forward (glanced at the docs). I'll start a new PR moving that logic as a follow up

drbh:
can be run with the `--grammar-support` cli flag or `GRAMMAR_SUPPORT=true` env variable

```bash
text-generation-launcher \
  --model-id HuggingFaceH4/zephyr-7b-beta \
  --grammar-support
```

### Requests

With grammar support:

```bash
 curl -s 'http://localhost:3000/generate' \
--header 'Content-Type: application/json' \
--data '{
    "inputs": "[INST]convert to JSON: text-generation-inference by huggingface is my favorite way to serve llms. [/INST]",
    "parameters": {
        "max_new_tokens": 200,
        "grammar": {
            "properties": {
                "software": {
                    "type": "string"
                },
                "rating": {
                    "type": "integer",
                    "minimum": 1,
                    "maximum": 5
                }
            },
            "required": ["software", "rating"]
        }
    }
}' | jq .
```

```json
{
  "generated_text": "{\n\n  \"rating\": 5,\n  \"software\": \"text-generation-inference by huggingface\"\n}"
}
```


Without grammar support:

If support is not toggled on then sending a `grammar` in the parameters will result in 

```json
{
  "error": "Input validation error: grammar is not supported",
  "error_type": "validation"
}
```

server/text_generation_server/models/flash_causal_lm.py
Narsil: Black is fine, we just need to enforce it repo wide in the CI. (We need to pin a revision though, black is not great at backward compatibility)

server/text_generation_server/utils/logits_process.py
Narsil: Why is this resolved ? Did you resolve and forgot to push maybe ?

server/text_generation_server/utils/logits_process.py
Narsil: While doing that, maybe we should ask more explicitness from the users, and force the grammar type to be specified.

That would avoid a try except abuse, and make error message probably more readable. Wdyt ?

Narsil:
```
{ 'grammar": {"type": "regex", "content": "..."}}
{ 'grammar": {"type": "json", "content": "..."}}
```

Also higher level API might be even better to expose to users: https://www.anyscale.com/blog/anyscale-endpoints-json-mode-and-function-calling-features

Narsil: It's okay to leave things for future PRs, but nothing that we tag in a revision can be removed later. So all the surface must be correct for us to do a release.

server/text_generation_server/utils/logits_process.py
drbh: @Narsil I think I may have misunderstood the original comment, I moved the fsm creation into the `HeterogeneousGrammarLogitProcessor.__init__` and access `self.fsms[i]` in the `HeterogeneousGrammarLogitProcessor.__call__`. 

This should reduce the number of times the fsm is generated to the number of times `HeterogeneousNextTokenChooser` is initialized instead of on each call.

Is there a different location I should move the fsm compilation too? Thank you!

Narsil: `mask` is being generated over and over. 
Reusing the mask seems better here (this kind of stuff makes a difference unfortunately).

Just create the mask once, and fill_ it to clean its values.
To be fair, resetting the values of the mask is costly anyway, maybe this is too early optimizations, but allocation in a loop is really and easy one to fix.

For reference, I wont 2ms/token on the mamba + cuda graphs by moving the n_layers tensors into a single tensor so that the copies would be a single kernel launch (that's how bad launching any single op is ).

Narsil:
> --grammar-support

I would have done it the other way personally, `--disable-grammar-support`, just so that the default is to support it.
Most users will be happy to have the features, the defence is only intended for power users, no ? (We could even remove the flag when we find a clever solution to slow down only the grammar requests down and not all users at the same time).

OlivierDehaene (commented):
Nice :)
Have you verified if it works well with speculation?

server/text_generation_server/utils/logits_process.py
```diff
@@ -464,3 +471,129 @@ def filter(self, indices):
...
+    ):
+        for i in range(logits.shape[0]):
+            fsm = self.fsms[i]
+            if fsm_grammar_states[i] == -1 or fsm is None:
+                continue
+            allowed_tokens = fsm.allowed_token_ids(fsm_grammar_states[i])
+            mask[allowed_tokens] = 0
+            biased_scores = logits[i] + mask
+            mask.fill_(-math.inf)
+            logits[i] = biased_scores
```
OlivierDehaene, line 586: Does this work?

```suggestion
        mask = torch.full_like(logits, -math.inf)
        
        for i in range(logits.shape[0]):
            fsm = self.fsms[i]
            if fsm_grammar_states[i] == -1 or fsm is None:
                continue
            allowed_tokens = fsm.allowed_token_ids(fsm_grammar_states[i])
            mask[i, allowed_tokens] = 0

        logits += mask
```

drbh: ooo yes this is great and this was very helpful! I added this to the latest commits with a bunch of other refactors mentioned above. Thank you!!

drbh:
> Nice :) Have you verified if it works well with speculation?

Now it should 🙂

drbh:
> > --grammar-support
> 
> I would have done it the other way personally, `--disable-grammar-support`, just so that the default is to support it. Most users will be happy to have the features, the defence is only intended for power users, no ? (We could even remove the flag when we find a clever solution to slow down only the grammar requests down and not all users at the same time).

great points! I've updated the PR to use `--disable-grammar-support` instead in the latest commits, thank you!

Narsil (approved):
LGTM.

I think we can still improve further, let's do this in other PRs.

PawelFaron:
> can be run with the `--grammar-support` cli flag or `GRAMMAR_SUPPORT=true` env variable
> 
> ```shell
> text-generation-launcher \
>   --model-id HuggingFaceH4/zephyr-7b-beta \
>   --grammar-support
> ```
> 
> ### Requests
> With grammar support:
> 
> ```shell
> curl -s 'http://localhost:3000/generate' \
> --header 'Content-Type: application/json' \
> --data '{
>     "inputs": "[INST]convert to JSON: I saw a puppy a cat and a raccoon during my bike ride in the park [/INST]",
>     "parameters": {
>         "max_new_tokens": 200,
>         "repetition_penalty": 1.3,
>         "grammar": {
>             "properties": {
>                 "location": {
>                     "type": "string"
>                 },
>                 "activity": {
>                     "type": "string"
>                 },
>                 "animals_seen": {
>                     "type": "integer",
>                     "minimum": 1,
>                     "maximum": 5
>                 },
>                 "animals": {
>                     "type": "array",
>                     "items": {
>                         "type": "string"
>                     }
>                 }
>             },
>             "required": ["location", "activity", "animals_seen", "animals"]
>         }
>     }
> }' | jq .
> ```
> 
> ```json
> {
>   "generated_text": "{\n\"activity\": \"biking\",\n\"animals\": [\"puppy\",\"cat\",\"raccoon\"]\n  , \"animals_seen\": 3,\n   \"location\":\"park\"}"
> }
> ```
> 
> Without grammar support:
> 
> If support is not toggled on then sending a `grammar` in the parameters will result in
> 
> ```json
> {
>   "error": "Input validation error: grammar is not supported",
>   "error_type": "validation"
> }
> ```

Did something change in the final implementation? I'm geetting this error with thix exact request:
`        }'
Failed to deserialize the JSON body into the target type: parameters.grammar: missing field `type` at line 27 column 4(base)`

drbh:
Hi @PawelFaron and @paulcx thank you both for the feedback and for testing the new feature! 

In order to resolve the issues above please note that 

1. the flag `--grammar-support` has been deprecated in favor of `--disable-grammar-support`. Gramma support is available by default and does not require a launcher flag to use
2. the request format changed slightly and is now nested inside a new object. The grammar includes a `type` and `value` at the top level; please see the updated example payload below.

**notes: the grammar can be type `json` or `regex`. The grammar needs to compile on the first request which can take a couple of seconds but subsequent requests should be faster.

```json
{
    "inputs": "[INST]convert to JSON: I saw a puppy a cat and a raccoon during my bike ride in the park [/INST]",
    "parameters": {
        "max_new_tokens": 200,
        "repetition_penalty": 1.3,
        "grammar": {
            "type": "json",
            "value": {
                "properties": {
                    "location": {
                        "type": "string"
                    },
                    "activity": {
                        "type": "string"
                    },
                    "animals_seen": {
                        "type": "integer",
                        "minimum": 1,
                        "maximum": 5
                    },
                    "animals": {
                        "type": "array",
                        "items": {
                            "type": "string"
                        }
                    }
                },
                "required": ["location", "activity", "animals_seen", "animals"]
            }
        }
    }
}
```

I hope this is helpful, please let me know if there are any other issues!

paulcx:
Thanks @drbh. This demo provided above works. btw, would you please have a look the errors?

Jason-CKY:
Hi I've tested this feature on orca-13b and llama-2-13b, both of which just generates `<unk>` tokens until the `max_new_tokens` is reached when trying to enforce grammar.

OlivierDehaene:
@Jason-CKY #1578 will fix this issue.

paulcx:
> @Jason-CKY #1578 will fix this issue.

errors exist in #1578 as showed

image

PR #1541, opened 2024-02-08
feat(server): add frequency penalty

OlivierDehaene:


router/src/lib.rs
```diff
@@ -205,10 +214,71 @@ pub(crate) struct ChatCompletion {
 pub(crate) struct ChatCompletionComplete {
     pub index: u32,
     pub message: Message,
-    pub logprobs: Option<Vec<f32>>,
```
OlivierDehaene, line 208: @drbh why was this vec f32?

router/src/server.rs
```diff
@@ -570,8 +571,8 @@ async fn chat_completions(
     let stream = req.stream;
     let max_new_tokens = req.max_tokens.or(Some(100));
     let repetition_penalty = req
```
OlivierDehaene, line 573: @drbh I think this was the wrong parameter.

server/text_generation_server/utils/logits_process.py
```diff
@@ -118,6 +118,62 @@ def filter(self, indices):
...
+        ).unsqueeze(1)
+
+    def __call__(self, input_ids: torch.Tensor, scores: torch.Tensor) -> torch.Tensor:
+        score = torch.gather(scores, 1, input_ids)
+        # if score < 0 then penalty has to be multiplied to reduce the previous token probability
+        score = -torch.where(
+            score < 0, score * self.penalty_tensor, score / self.penalty_tensor
+        )
+
+        return scores.scatter_add_(1, input_ids, score)
```
OlivierDehaene, line 167: @Narsil I'm not 100% sure of this. This seems right when compared to https://platform.openai.com/docs/guides/text-generation/parameter-details

OlivierDehaene (commented):
black messed up the PR..
@drbh, @Narsil should we add a pre-commit hook on black?

router/src/lib.rs
drbh: whoops silly mistake, thanks for resolving in this PR 🙏

router/src/server.rs
drbh: ahh yes, I misunderstood the difference between the frequency and presence and how it mapped to the `repetition_penalty`. Thank you! 🙏

PR #1542, opened 2024-02-08
feat(router): add max_batch_size

OlivierDehaene:
Some hardware require a maximum batch size.


launcher/src/main.rs
```diff
@@ -279,6 +279,11 @@ struct Args {
     #[clap(default_value = "20", long, env)]
     max_waiting_tokens: usize,
 
+    /// Enforce a maximum number of requests per batch
+    /// Specific flag for hardware targets that do not support unpadded inference
```
dacorvo, line 283: 'unpadded inference' is a bit unclear to me, but I suppose it corresponds to other configurations.

dacorvo (approved):
LGTM, thanks !

Narsil (approved):
Look much better thana my PR

launcher/src/main.rs
Narsil: Unpadded = Flash attention "varlen" (which means requests are stacked linearly without requiring padding tokens).

Padded:

input_ids = [
 [3, 53, 63],
 [0, 0, 234],
]

Unpadded:

input_ids = [3, 53, 63, 234] + [0, 3, 4] (seqlengths)

Issue #1545, opened 2024-02-09
Template Not Found When Using OpenAI format Chat Completion


binarycrayon:
### System Info

Docker Image: ghcr.io/huggingface/text-generation-inference:sha-1734540
Instance: AWS A10G via Huggingface Interfence Endpoint

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [X] An officially supported command
- [ ] My own modifications

### Reproduction

On Huggingface Inference Endpoint:

Served a finetuned model: JamAndTeaStudios/dialogue-choice-merged-01-30-sft-mistral-7b-instruct-0.2
Above model is a peft fine-tuned and merged Mistral 7B model, including tokenizer
Task: Text Generation
TGI Docker Image: ghcr.io/huggingface/text-generation-inference:sha-1734540
Instance: AWS GPU A10G in east region

### Expected behavior

Once the inference url is up and running, I followed 
https://huggingface.co/blog/tgi-messages-api
and configured Openai client with the URL 
then call the chat completion endpoint

saw 402 error:
```
UnprocessableEntityError: Error code: 422 - {'error': 'Template error: template not found', 'error_type': 'template_error'}
```

log from TGI endpoint:
```
2024/02/09 11:13:20 ~ {"timestamp":"2024-02-09T19:13:20.293485Z","level":"INFO","fields":{"message":"Shard ready in 5.304520806s"},"target":"text_generation_launcher","span":{"rank":0,"name":"shard-manager"},"spans":[{"rank":0,"name":"shard-manager"}]}
2024/02/09 11:13:20 ~ {"timestamp":"2024-02-09T19:13:20.392520Z","level":"INFO","fields":{"message":"Starting Webserver"},"target":"text_generation_launcher"}
2024/02/09 11:13:20 ~ {"timestamp":"2024-02-09T19:13:20.435457Z","level":"INFO","message":"Using the Hugging Face API","target":"text_generation_router","filename":"router/src/main.rs","line_number":175}
[...]

PR #1550, opened 2024-02-12
feat: add deserialize_with that handles strings or objects with content

drbh:
This PR adds a simple custom `deserialize_with` function that parses a string or an object with a content property. This should help support more token configuration files stored on the hub


drbh:
fixes https://github.com/huggingface/text-generation-inference/issues/1545

PR #1577, opened 2024-02-19
improve endpoint support

drbh:
small PR to add a new interface endpoint behind a feature

### DIFF
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -21,36 +21,6 @@ use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Layer};
 
-#[allow(dead_code)] // many of the fields are not used
-#[derive(Debug)]
-struct VertexAIConfig {
-    aip_http_port: u16,
-    aip_predict_route: String,
-    aip_health_route: String,
-}
-
-impl VertexAIConfig {
-    fn new(aip_http_port: u16, aip_predict_route: String, aip_health_route: String) -> Self {
-        Self {
-            aip_http_port,
-            aip_predict_route,
-            aip_health_route,
-        }
-    }
-    fn to_env(&self) {
-        // NOTE: this will only set the values for this process
-        // NOTE: child processes cannot set env vars for their parents
-        // TODO: find a way to set the values for the whole system
-        // - maybe write to a file
-        // - maybe use a shell script to set the values
-        // - maybe these values are set upstream (before this process is started)
-        //  - if set upstream maybe we read in; if we need them?
-        std::env::set_var("AIP_HTTP_PORT", self.aip_http_port.to_string());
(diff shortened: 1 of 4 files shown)


router/src/main.rs
```diff
@@ -328,6 +328,15 @@ async fn main() -> Result<(), RouterError> {
     tracing::info!("Setting max batch total tokens to {max_supported_batch_total_tokens}");
     tracing::info!("Connected");
 
+    // Determine the server port based on the feature and environment variable.
+    let port = if cfg!(feature = "google") {
```
Narsil, line 332: Let's make that compile time.

```rust
#[cfg(feature = "google")]
let port = xxx

#[cfg(not(feature = "google"))]
let port = xxxx
```

router/src/server.rs
```diff
@@ -969,10 +1058,24 @@ pub async fn run(
...
 
     // Combine routes and layers
-    let app = Router::new()
+    let mut app = Router::new()
         .merge(swagger_ui)
         .merge(base_routes)
-        .merge(aws_sagemaker_route)
+        .merge(aws_sagemaker_route);
+
+    if cfg!(feature = "google") {
```
Narsil, line 1066: Same, compile time is preferred I think.

router/src/main.rs
Narsil: Nvm, I checked the macros, it's also compile time.

drbh: nice, however if helpful I can update the syntax

Narsil:
Have you checked if the openapi is correctly created ? I think you forgot to add the structs to the utopia thing to generate them  properly, no ?

Narsil (approved):

OlivierDehaene:
This PR does not compile.

router/src/server.rs
```diff
@@ -942,8 +1036,30 @@ pub async fn run(
...
+        #[cfg(feature = "google")]
+        {
+            // limiting mutability to the smallest scope necessary
+            let mut doc = doc;
+            doc.merge(VertextApiDoc::openapi());
+            doc
+        }
+        #[cfg(not(feature = "google"))]
+        ApiDoc::openapi()
+    };
```
OlivierDehaene, line 1059: This should be:

```rust
// Define VertextApiDoc conditionally only if the "google" feature is enabled
    let doc = {
        // avoid `mut` if possible
        #[cfg(feature = "google")]
        {
            use crate::VertexInstance;

            #[derive(OpenApi)]
            #[openapi(
            paths(vertex_compatibility),
            components(schemas(VertexInstance, VertexRequest, VertexResponse))
            )]
            struct VertextApiDoc;

            // limiting mutability to the smallest scope necessary
            let mut doc = ApiDoc::openapi();
            doc.merge(VertextApiDoc::openapi());
            doc
        }
        #[cfg(not(feature = "google"))]
        ApiDoc::openapi()
    };
```

PR #1578, opened 2024-02-19
fix(router): fix openapi and add jsonschema validation

OlivierDehaene:
@drbh, can you check the validation part?

### DIFF
--- a/server/text_generation_server/utils/tokens.py
+++ b/server/text_generation_server/utils/tokens.py
@@ -328,7 +328,6 @@ class HeterogeneousNextTokenChooser:
             scores = scores.view(B, S, -1)
 
         next_ids = torch.zeros((B, S), device=scores.device, dtype=torch.long)
-        mask = torch.full((scores.shape[-1],), -math.inf, device=self.device)
 
         for j in range(S):
             _scores = scores[:, j]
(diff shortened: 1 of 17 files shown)


drbh (approved):
LGTM just tested and worked locally. Thank you!

drbh (dismissed):
LGTM thanks for adding the `draft202012` feat

server/text_generation_server/utils/tokens.py
```diff
@@ -338,10 +337,10 @@ def __call__(
                 _scores = self.repetition_processor(input_ids, _scores)
             if self.frequency_processor is not None:
                 _scores = self.frequency_processor(input_ids, _scores)
-            for warper in self.warpers:
-                _scores = warper(input_ids, _scores)
             if self.grammar_processor is not None:
                 _scores = self.grammar_processor(_scores, self.fsm_grammar_states)
+            for warper in self.warpers:
+                _scores = warper(input_ids, _scores)
```
OlivierDehaene, line 343: @drbh it's very important that logits processor are done BEFORE warpers. Otherwise you can generate nonsense. For example if top_k is set and you warp before the grammar processor, may remove all valid tokens and you cannot generate anything. See https://github.com/huggingface/text-generation-inference/pull/1539#issuecomment-1953392530

drbh: thank you for pointing out! this is very helpful and makes a lot of sense. Thanks for resolving!

PR #1579, opened 2024-02-20
fix: avoid default message

drbh:
This PR avoids setting a default message in order to avoid unexpected generations

### DIFF
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -559,7 +559,7 @@ pub(crate) struct Message {
     pub role: String,
     #[schema(example = "My name is David and I")]
     pub content: String,
-    #[serde(default)]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


router/src/lib.rs
```diff
@@ -568,6 +559,7 @@ pub(crate) struct Message {
     pub role: String,
     #[schema(example = "My name is David and I")]
     pub content: String,
+    #[serde(default)]
```
Narsil, line 562: Can we omit the field if it's empty too at the same time ?

router/src/server.rs
```diff
@@ -594,6 +594,7 @@ async fn chat_completions(
             ));
         }
     };
+    tracing::debug!("Templated input: {}", inputs);
```
Narsil, line 597: Can we remove this ?

Narsil (approved):
LGTM

PR #1586, opened 2024-02-21
fix: fix openapi schema

OlivierDehaene:


paulcx:
errors exsited in schema from image:sha-7dbaf9e

```
Resolver error at paths./v1/completions.post.responses.200.content.application/json.schema.properties.choices.items.properties.delta.properties.tool_calls.allOf.0.$ref
Could not resolve reference: Could not resolve pointer: /components/schemas/DeltaToolCall does not exist in document
Resolver error at paths./v1/completions.post.responses.200.content.application/json.schema.properties.choices.items.properties.delta.properties.tool_calls.$ref
Could not resolve reference: Could not resolve pointer: /components/schemas/DeltaToolCall does not exist in document
```

PR #1588, opened 2024-02-22
Revamp medusa implementation so that every model can benefit.

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


server/text_generation_server/models/__init__.py
```diff
@@ -193,16 +176,18 @@ def get_model(
             model_id,
             revision,
             quantize=quantize,
+            use_medusa=use_medusa,
             dtype=dtype,
             trust_remote_code=trust_remote_code,
         )
 
-    if model_type == "gpt_bigcode":
```
OlivierDehaene, line 200: How do you know if it's a true gpt2?

server/text_generation_server/utils/medusa.py
```diff
@@ -3,12 +3,6 @@
 from text_generation_server.utils.layers import TensorParallelHead, FastLinear
 
 
-@dataclass
```
OlivierDehaene, line 6: Should we remove this file?

server/text_generation_server/models/causal_lm.py
```diff
@@ -573,7 +580,7 @@ def generate_token(
         # slice the attention mask to the correct shape
         attention_mask = batch.attention_mask[:, : -batch.padding_right_offset]
 
-        logits, past = self.forward(
```
OlivierDehaene, line 576: Do we use the speculative_logits anywhere?

OlivierDehaene (commented):
I think you forgot to add galactica back, not that I care too much.

server/text_generation_server/models/__init__.py
Narsil: I do not, I assumed the code was compatible and would work.

`model_type == "gpt2" and model_id.startwith("bigcode/")` is more correct no ?

server/text_generation_server/utils/medusa.py
Narsil: Totally

server/text_generation_server/models/causal_lm.py
Narsil: Not here, I would need to implement the logic here, but at least the types are correct.

Narsil: Since we're not using the same token choosers as with flash, I would need a lot more code, which should be in a follow-up, no ?

OlivierDehaene: Ok!

OlivierDehaene (dismissed):

PR #1591, opened 2024-02-22
fix: prefer seed 1 in all tests

drbh:
This WIP PR aims to stop tests from being flakey!

Currently `seed=0` in various tests. Since this value is sent via protobuf to the model the `0` seed is ignored and equivalent to `None`. 

This PR starts the work to remove all zero seeds and update all the snapshots to avoid non deterministic output in CI

Ref: https://protobuf.dev/programming-guides/proto3/#default


OlivierDehaene:
>  Since this value is sent via protobuf to the model the `0` seed is ignored and equivalent to `None`.

I think it's the other way around. If you send a protobuf message with `seed=None` it will be interpreted as 0.

PR #1594, opened 2024-02-23
fix: avoid duplicate bos token

drbh:
this PR updates the tokenizer to avoid appending the bos token since this is now included in the config and is applied upstream by the chat template

changes to template: https://huggingface.co/google/gemma-7b-it/discussions/42/files

### DIFF
--- a/server/text_generation_server/models/custom_modeling/flash_gemma_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_gemma_modeling.py
@@ -87,7 +87,7 @@ class GemmaTokenizerFast(PreTrainedTokenizerFast):
         bos_token="<bos>",
         eos_token="<eos>",
         pad_token="<pad>",
-        add_bos_token=True,
+        add_bos_token=False,
         add_eos_token=False,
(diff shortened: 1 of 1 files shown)


OlivierDehaene:
Should the model have a BOS token when not using templates? If so we can't do that I think, right?

Narsil:
We shouldn't even use that class altogether, everything should be upstream in transformers, no ?

drbh:
great points, thank you! I seems like the issue is theres more then one place special tokens can be added once chat templates were added... maybe we can skip special tokens in all or some cases on `batch_decode` (it would be nice to skip them in all "chat" cases)

```
...
self.tokenizer.batch_decode(
  prefill_token_ids,
  clean_up_tokenization_spaces=False,
  skip_special_tokens=True,
)
```

in the Conversation pipeline in transformers special tokens seem to be expected to be handled by the template rather than the tokenizer: https://huggingface.co/google/gemma-7b-it/discussions/42#65d8904dd532b55d20355844

drbh:
stale, low priority (will re open if causes issues)

PR #1605, opened 2024-02-28
feat: starcoder2

OlivierDehaene:


server/text_generation_server/models/__init__.py
```diff
@@ -184,6 +186,16 @@ def get_model(
             trust_remote_code=trust_remote_code,
         )
 
+    if model_id.startswith("facebook/galactica"):
```
Narsil, line 189: Should we really keep that since that model has been removed from huggingface altogether ?

Narsil (approved):

PR #1617, opened 2024-02-29
Fix async client timeout

### LINKED ISSUE #1616: Incorrectly multiplied timeout by 60 in the asynchronous client
### System Info

I'm testing TGI using Docker. Below is the exact command I'm utilizing:

```console
docker run --gpus '"device=1,2"' --shm-size 1g -p 8000:80 -v ~/tgi-test:/data ghcr.io/huggingface/text-generation-inference:1.4 --model-id mistralai/Mistral-7B-v0.1 --max-input-length 8000 --max-total-tokens 8001
```

### Information

- [ ] Docker
- [ ] The CLI directly

### Tasks

- [ ] An officially supported command
- [ ] My own modifications

### Reproduction

Given the generation request:

```python
async def test():
  start = time.time()
  try:
   response = await client.generate('1' * 6_000, max_new_tokens=1_800)
  except Exception as ex:
    pass
  print(time.time() - start)
```

And this async client definition:

```python
client = AsyncClient('http://localhost:8000', timeout=1)
```

It doesn't timeout after 1 second:

```python
>>> await test()
60.88534379005432
```

But if we create a client with a timeout of 2/60:

```python
client = AsyncClient('http://localhost:8000', timeout=(2/60))
```

It does timeout after 2 seconds:

```python
>>> await test()
2.0035104751586914
```

### Expected behavior

The function should have timed out after 1 second with this client definition:

```python
client = AsyncClient('http://localhost:8000', timeout=1)
```

hugoabonizio:
# What does this PR do?

Fixes #1616

According to the [aiohttp.ClientTimeout docs](https://docs.aiohttp.org/en/stable/client_reference.html#aiohttp.ClientTimeout), the arguments should be in seconds. This PR removes the multiplication by 60.

@OlivierDehaene OR @Narsil

### DIFF
--- a/clients/python/text_generation/client.py
+++ b/clients/python/text_generation/client.py
@@ -424,7 +424,7 @@ class AsyncClient:
         self.base_url = base_url
         self.headers = headers
         self.cookies = cookies
-        self.timeout = ClientTimeout(timeout * 60)
+        self.timeout = ClientTimeout(timeout)
 
     async def chat(
         self,


OlivierDehaene (approved):
thanks

Issue #1624, opened 2024-03-05
Incorrect output for tools (function calling) accord to openai specs


puppetm4st3r:
### System Info

Hello, first of all, congratulate you for the great work you have done. I have tested the tool selection feature with the official open ai client and docker image 1.4.3, and I have noticed that the tool selection result does not talk to the open ai api result specification.

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [X] An officially supported command
- [ ] My own modifications

### Reproduction

send a request to the tgi 1.4.3 docker

	tools = [
	    {
		"type": "function",
		"function": {
		    "name": "get_current_weather",
		    "description": "Get the current weather",
		    "parameters": {
		        "type": "object",
		        "properties": {
		            "location": {
		                "type": "string",
		                "description": "The city and state, e.g. San Francisco, CA",
		            },
		            "format": {
		                "type": "string",
		                "enum": ["celsius", "fahrenheit"],
		                "description": "The temperature unit to use. Infer this from the users location.",
		            },
		        },
		        "required": ["location", "format"],
		    },
		},
	    },
	    {
		"type": "function",
		"function": {
		    "name": "get_n_day_weather_forecast",
		    "description": "Get an N-day weather forecast",
		    "parameters": {
		        "type": "object",
		        "properties": {
		            "location": {
		                "type": "string",
		                "description": "The city and state, e.g. San Francisco, CA",
		            },
		            "format": {
		                "type": "string",
		                "enum": ["celsius", "fahrenheit"],
		                "description": "The temperature unit to use. Infer this from the users location.",
		            },
		            "num_days": {
		                "type": "integer",
		                "description": "The number of days to forecast",
		            },
		        },
		        "required": ["location", "format", "num_days"],
		    },
		},
	    }
	]

	# Initialize the client, pointing it to one of the available models
	client = OpenAI(
	  base_url="http://llm_server:3000/v1",
	  api_key="_"
	)

        from openai import OpenAI
	# NOTE: tools defined above and removed for brevity
	chat_completion = client.chat.completions.create(
	  model="tgi",
	  messages=[
	      {
		  "role": "system",
		  "content": "Don't make assumptions about what values to plug into functions. Ask for clarification if a user request is ambiguous.",
	      },
	      {
		  "role": "user",
		  "content": "What's the weather like the next 3 days in San Francisco, CA?",
	      },
	  ],
	  tools=tools,
	  tool_choice="auto",  # tool selected by model
[...]

PR #1628, opened 2024-03-06
Router grammar compile

drbh:
WIP

This PR is aims to move the compilation of grammars to the router. This should improve performance and remove work from the server. Just opening a PR for visibility/feedback

Currently this PR uses py03 to create grammar workers in the validator and can parse and compile them into a state map. Work is still needed to make use of the state map on the server side and remove existing grammar compilation code.

- [X] load outlines via py03
- [X] compile object to regex
- [X] compile regex to state map
- [X] pass statemap to server via grpc
- [x] replace grammar parse with state map

**unknowns
- how much does this improve performance?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1402,6 +1402,15 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "523dc4f511e55ab87b694dc30d0f820d60906ef06413f93d4d7a1385599cc149"
 
 [[package]]
+name = "memoffset"
+version = "0.9.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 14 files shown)


drbh:
pending
- [ ] ensure python is loaded correctly in CI
- [ ] simplify grammar deserialize logic 
- [ ] cache compiled grammar on router

puppetm4st3r:
Hi @drbh can you give me a little hand with: https://github.com/huggingface/text-generation-inference/issues/1624 

I get a solution, but I need to know if it would be useful to yours, and if it is useful, how to make the PR :)

regards

drbh:
closing for now

reasons:
1. added complexity in build/runtime (adding python in rust)
2. no large performance impact
3. outlines (library used) is in flux and performance is improving (maybe contribute to rewriting parts in rust too)

**note will follow up with comment next time grammar compilation changes are made

PR #1650, opened 2024-03-18
fix: improve tool type, bump pydantic and outlines

drbh:
This PR resolves a couple 

- [X] adjusts the tool response to align with openai's tools response type
- [X] bumps pydantic to `2.6.4` in all apps (resolves dependency issue when running tests)
- [X] bump `outlines` version and fix import for new name

PR #1653, opened 2024-03-19
chore: update `outlines` function call to new name

joennlae:
First of all really lovely project 💯 

This issue has been introduced since outlines version 0.32
With this pull request https://github.com/outlines-dev/outlines/pull/556
It changes the function name from
`build_regex_from_object` to `build_regex_from_schema`.
This leads to an error in newer docker containers when starting tgi.

I did not update the poetry.lock file as there is a forced local installation of `outlines` here: https://github.com/huggingface/text-generation-inference/blob/0d72af5ab01a5b1dabd5beda953403d63b1886e0/server/Makefile#L26

To mitigate it currently in newer docker containers, just run:

```bash
pip install outlines==0.0.31
```

### DIFF
--- a/server/pyproject.toml
+++ b/server/pyproject.toml
@@ -34,7 +34,7 @@ peft = { version = "^0.9.0", optional = true }
 torch = { version = "^2.1.1", optional = true }
 scipy = "^1.11.1"
 pillow = "^10.0.0"
-outlines= { version = "^0.0.27", optional = true }
+outlines= { version = "^0.0.32", optional = true }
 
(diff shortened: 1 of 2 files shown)


joennlae:
Update: it was only in my env. But if you update outlines, then this will be an issue :-) Sorry for sounding the alarm bell quite hard.

drbh:
hi @joennlae thanks for pointing this out! outlines has been bumped and uses the new function as of this PR: https://github.com/huggingface/text-generation-inference/pull/1650

PR #1658, opened 2024-03-20
feat: support force downcast after FastRMSNorm multiply for Gemma

drbh:
This PR adds `force_downcast_after` to `FastRMSNorm.forward` which is used in the Gemma model. References https://github.com/huggingface/transformers/pull/29402 and https://github.com/huggingface/transformers/pull/29729

Setting `force_downcast_after=True` will perform the `hidden_states * weight` multiplication in f32 and then downcast to half. This differs slightly from the current implementation which first casts the `hidden_states` to a half and then multiples.

### DIFF
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -687,7 +687,7 @@ try:
             weight = weights.get_tensor(f"{prefix}.weight")
             return cls(weight, eps)
 
-        def forward(self, hidden_states, residual=None):
+        def forward(self, hidden_states, residual=None, force_downcast_after=False):
             if hidden_states.shape[-1] > 8192:
                 if residual is not None:
                     hidden_states += residual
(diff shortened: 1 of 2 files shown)


server/text_generation_server/utils/layers.py
```diff
@@ -687,7 +687,7 @@ def load(cls, prefix, weights, eps=1e-6):
             weight = weights.get_tensor(f"{prefix}.weight")
             return cls(weight, eps)
 
-        def forward(self, hidden_states, residual=None):
+        def forward(self, hidden_states, residual=None, force_downcast_after=False):
```
Narsil, line 690: I'd personally use a different method `forward_downcast_after`.

Also this only triggers when hidden_size > 8192 meaning it won't trigger for gemma.
Having a dedicated method seems simpler.

drbh: that makes alot of sense thanks for the comments. I've removed the branching logic from `FastRMSNorm` and added `forward_downcast_after` into the gemma modeling code

Narsil (approved):
Even better !

PR #1668, opened 2024-03-22
v1.4.4

OlivierDehaene:


paulcx:
Capture 035 - Swagger UI

The issue (in fastapi docs) has not been address in latest release image.

Issue #1657, opened 2024-03-20
Improve LLM's tool awareness


drbh:
### Feature request

Currently the tools feature does not return the name of the chosen function. This is due to how tools/functions are implemented in TGI (by constrained generation vs fine tuned model). 

It's been proposed that the internal structure is updated to force the generation to include names https://github.com/huggingface/text-generation-inference/pull/1650 and overall the function mechanism may be improved.

Opening this issue as a place for others thoughts/idea/uses on how to improve tools to be as useful as possible.

puppetm4st3r:
I have solved the name issue with a ref in the original json grammar, i'm not familiar with PR 's in github but will try it in the nex days, otherwise i Can leave yoy the changes there are small and a little few. With that changes the output is 100% aligned with the open ai specs.

regards

puppetm4st3r:
in the server.rs the extraction of the name and the tool parser looks like this:

```
let (tool_calls, output) = if tool_grammar.is_some() {
            // gen_text should be valid json
            let gen_text_value: Value = serde_json::from_str(&generation.generated_text).map_err(|e| {
                (
                    StatusCode::UNPROCESSABLE_ENTITY,
                    Json(ErrorResponse {
                        error: e.to_string(),
                        error_type: "Input validation error".to_string(),
                    }),
                )
            })?;
        
            let tool_call_id = generate_random_id();
            let tool_call = ToolCall {
                id: tool_call_id,
                r#type: "function".to_string(),
                function: FunctionDefinitionResponse {
                    // Extract the function name from "function" -> "name", not as a constant
                    name: gen_text_value
                        .get("function") // Access the JSON value of "function"
                        .and_then(|f| f.get("name")) // Directly access "name" inside "function"
                        .and_then(|name| name.as_str()) // Ensure "name" is a string
                        .unwrap_or("default_function_name") // Provide a default name if none is found
                        .to_string(),
                    // Serialize the JSON object obtained from "function" to an escaped JSON string
                    arguments: gen_text_value
                        .get("function") // Access the JSON value of "function"
                        .map_or_else(
                            || Ok("{}".to_string()), // Use an empty JSON object if "function" does not exist
                            |f| {
                                // Remove the "name" key from properties before serialization
                                let mut f_cloned = f.clone();
                                if let Value::Object(ref mut props) = f_cloned {
                                    props.remove("name"); // Remove the "name" key
                                }
[...]

PR #1693, opened 2024-04-02
feat: improve tools to include name and add tests

drbh:
This PR makes tool calling aware of the name of the function selected. 

Fixes: https://github.com/huggingface/text-generation-inference/issues/1657

Thank you @puppetm4st3r for the helpful snippets, large parts of this PR are simply refactors of the code shared 🙏

**opening draft PR because small tweaks are needed before merging

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -756,19 +756,34 @@ async fn chat_completions(
 ) -> Result<Response, (StatusCode, Json<ErrorResponse>)> {
     metrics::increment_counter!("tgi_request_count");
 
-    let stream = req.stream;
-    let max_new_tokens = req.max_tokens.or(Some(100));
-    let repetition_penalty = req
-        .presence_penalty
-        // rescale repetition_penalty from (-2.0, 2.0) to (0.0, 4.0)
-        .map(|x| x + 2.0);
-    let logprobs = req.logprobs.unwrap_or(false);
-    let seed = req.seed;
-    let stop = req.stop.unwrap_or_default();
-    let tool_prompt = req.tool_prompt.unwrap_or_default();
+    let ChatRequest {
+        frequency_penalty: _,
+        logit_bias: _,
+        logprobs,
+        max_tokens,
+        messages,
+        model: _,
+        n: _,
+        presence_penalty,
+        seed,
+        stop,
+        stream,
+        temperature: _,
+        tools,
+        tool_choice,
+        tool_prompt,
+        top_p: _,
+        top_logprobs: _,
+    } = req;
+
+    let repetition_penalty = presence_penalty.map(|x| x + 2.0);
+    let max_new_tokens = max_tokens.or(Some(100));
+    let logprobs = logprobs.unwrap_or(false);
+    let tool_prompt = tool_prompt.unwrap_or_default();
+    let stop = stop.unwrap_or_default();
 
     // extract tool grammar if present
-    let tool_grammar = match ToolGrammar::apply(req.tools.as_ref(), req.tool_choice.as_ref()) {
+    let tool_grammar = match ToolGrammar::apply(tools, tool_choice) {
         Ok(grammar) => grammar,
         Err(err) => {
             metrics::increment_counter!("tgi_request_failure", "err" => "validation");
@@ -784,7 +799,7 @@ async fn chat_completions(
     };
 
     // apply chat template to flatten the request into a single input
-    let mut inputs = match infer.apply_chat_template(req.messages) {
+    let mut inputs = match infer.apply_chat_template(messages) {
         Ok(inputs) => inputs,
         Err(err) => {
             metrics::increment_counter!("tgi_request_failure", "err" => "validation");
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -757,23 +757,17 @@ async fn chat_completions(
     metrics::increment_counter!("tgi_request_count");
 
     let ChatRequest {
-        frequency_penalty: _,
-        logit_bias: _,
         logprobs,
         max_tokens,
         messages,
-        model: _,
-        n: _,
         presence_penalty,
         seed,
         stop,
         stream,
-        temperature: _,
         tools,
         tool_choice,
         tool_prompt,
-        top_p: _,
-        top_logprobs: _,
+        ..
     } = req;
 
     let repetition_penalty = presence_penalty.map(|x| x + 2.0);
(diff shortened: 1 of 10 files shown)


puppetm4st3r:
thanks to you, you are doing a great job with TGI, it is an honor to be able to contribute with a little grain of sand.

I have looked the changes on this PR, and I can comment that during my local tests, the tool prompt greatly influences the final performance of the tool selection or error notification when none can be selected, adding -------------{}{} ----------- delimiting to build the final prompt influences very heavy on the results, so I recommend considering format the final prompt delimited like this:

```
// Proceed only if tool_prompt is not None
     if let Some(tool_prompt) = &req.tool_prompt {
         // Find the last message with role 'user'
         if let Some(last_user_message) = req.messages.iter_mut().rev().find(|msg| msg.role == "user") {
             // Generate the additional content combining tool_prompt and tools_str
             let additional_content = format!("\n\n---------------------------\n{}{}\n---- -----------------------", tool_prompt, tools_str);
            
             // If the last user message has existing content, append the additional content to it
             if let Some(content) = &mut last_user_message.content {
                 content.push_str(&additional_content); // Append to the existing content
             } else {
                 // If, for some reason, the content is None, replace it with the additional_content
                 last_user_message.content = Some(additional_content);
             }
         }
         // Note: If there is no message with role 'user', this block does nothing
     }
```

also changing the prompt from lib.rs
```
fn default_tool_prompt() -> Option<String> {
    Some(
        "\nBased on the conversation, please choose the most appropriate tool to use: ".to_string(),
    )
}
```

to:

`You will be presented with a JSON schema representing a set of tools.\nIf the user request lacks of sufficient information to make a precise tool selection: Do not invent any tool's properties, instead notify with an error message.\ n\nJSON Schema:\n
`

has a positive impact in the performance of the llm on the tool selection task.

to get the final prompt delimited I have to change the order of processing in the request, in the original code the tokenizer applies very on the first fragments on the code, to build the delimited final prompt I have to move the tokenizer process after the grammar generation and function definitions in order to build a final prompt with all stuff ready before tokenization process.

drbh:
Hi @puppetm4st3r thank you for the feedback on the PR , I've updated the default prompt and looked into changing how the tools are appended to the input.

Currently I'm not 100% sure what approach is best for passing tools to the LLM. 

At the moment we send:

`{inputs}{tool_prompt}{tools_str}`

where input is the users input after the template is applied to the incoming messages.

The method proposed in your PR seems to append the `{tool_prompt}{tools_str}` to the last message, and then applies the chat template.

Its unclear to me which approach is best and would love to find an alternative that enables this flexibility without adding more complexity (an extra template)...

puppetm4st3r:
yes, that part is a bit complex in Use cases, my motivation was to append to the last user message in order to try not confuse the llm, also implemented de callId property of the open ai spec because is useful in chathistory for the llm to understand previous turns of conversations that have called a function, the llm have to undestand and distinguish between turns funcion calls to not get confused,  Honestly, I didn't give much thought to what was the best strategy to include the prompt_tool and the tool definitions, I simply tried to maintain the consistency of messaging with the llm so as not to confuse it and I chose to modify the user's last message since that modification It is ephemeral and does not come back to the user (it is like a patch, I don't like it but I didn't find another direct and simple way to implement it by the moment)

puppetm4st3r:
Another option that I test a few months ago is to build a wrapper for the official OPEN AI client, which preset the chat template with the required stuff to execute function calling using json grammar guidance, the benefit you don't touch the server, the cons is that you create a new dependency/artifact to call functions and that breaks the concept of open ai compatible api., is more like open ai compatible client.

this option I feel takes us away from the original purpose

drbh:
@puppetm4st3r I think for now we'll include the current changes and keep the discussion open regarding tools and the template. It seems like there may be some convergence in using tools within templates https://github.com/huggingface/transformers/blob/76fa17c1663a0efeca7208c20579833365584889/src/transformers/models/cohere/tokenization_cohere_fast.py#L292 and this would require a larger change to the template logic.

puppetm4st3r:
great, yes Im ok that refactors on my locals with something like that at chat template level, now Im on the gym, later ar night I can share my chat templates that have result in good llm tool understansing on chat history and large multi turn conversarions, will back with more detailed ideas later! c ya

puppetm4st3r:
continuing the refinement of the idea,

Let's review the key findings we have uncovered during the implementation of this feature

1. Models are highly sensitive to the chat template, with some templates not performing well with sentences containing non-alphanumeric characters, such as a JSON schema. Through experimentation, the "ChatML" template was found to be the most effective.

3. Models not natively trained for function calling, attempting to mimic this behavior, are constrained to generating a JSON schema structure, hence incapable of returning feedback or non tool selection natural text. To mitigate this, the previous PR added the capability for the LLM to communicate errors (appending a "system" tool in a non clear or straightforward way), which is some crucial as this kind of models will not support optional use of tools.

5. Due to the sensitivity to chat templates, the user’s prompt must include instructions for the model to know which JSON schema to use for text generation. Without this, the generation process could be significantly delayed or result in empty tokens until the context window is filled.

7. Effective prompt engineering significantly enhances the model's tool selection quality, with even error notification potentially failing if the prompt is unclear or does not adhere to the chat template.
8. For chat history, only ChatML considers roles for tools (representing a tool call by the LLM), where the call ID is essential for the LLM to understand and logically sequence the calls and responses from the selected tools. However, in models not trained to understand the tool role, a workaround has been to modify the Jinja template of the chat template to render the tool call as part of the Assistant role, like "I have decided to call the following tool: …".

Conclusión:
Incorporating tool functionality into the TGI API presents significant challenges, particularly in maintaining clean integration without resorting to less desirable practices like modifying the last user role message. The interplay between chat templates, prompt engineering, and model capabilities is crucial for effective function calling, necessitating careful design and customization to achieve seamless and efficient tool integration and communication.

Thats for now, we will need to keep double clicking on these ideas and concepts.

puppetm4st3r:
ps: in my use case that work for me was modify the chatML template in order to support tools for non-pretrained LLM for tool call understanding plus appending the prompt_tools and prompt_str to the last user's role message, thats work fine, but it is not the most direct and clean solution. We now whats work but we need to implement that in a clear and best practice way.

this is my chatML modified template:
```
{% for message in messages %}
  {%- if message['role'] == 'system' %}
<|im_start|>system\n{{ message['content'] }}<|im_end|>\n
  {%- elif message['role'] == 'user' %}
<|im_start|>user\n{{ message['content'] }}<|im_end|>\n
  {%- elif message['role'] == 'tool' %}
<|im_start|>tool\ncall_id: {{ message['tool_call_id'] }}\nresult: {{ message['content'] }}<|im_end|>\n
  {%- elif message['role'] == 'assistant' %}
<|im_start|>assistant{%- if message.tool_calls -%}
      {%- for tool_call in message.tool_calls %}
tool_execution: call_id: {{ tool_call.id }}, function: {{ tool_call.function.name }}, arguments: {{ tool_call.function.arguments }}
      {%- endfor -%}
      {%- if not loop.last or not add_generation_prompt %}<|im_end|>{% endif -%}
    {%- else %}
{{message['content'] }}{%- if not loop.last or not add_generation_prompt %}<|im_end|>{% endif -%}
    {% endif -%}
  {% endif -%}
{% endfor -%}
{% if add_generation_prompt -%}
<|im_start|>assistant
{% endif %}
```

router/src/server.rs
```diff
@@ -754,19 +754,34 @@ async fn chat_completions(
...
+        seed,
+        stop,
+        stream,
+        temperature: _,
+        tools,
+        tool_choice,
+        tool_prompt,
+        top_p: _,
+        top_logprobs: _,
+    } = req;
```
Narsil, line 775:
```suggestion
   ..
   }  = req;
   ```

Narsil: You can remove all the `_` using `..`.

PR #1707, opened 2024-04-04
fix: fix CohereForAI/c4ai-command-r-plus

OlivierDehaene:
@Narsil @drbh this will update flash attention v2 and vllm.
You will need to re-install them.


server/text_generation_server/models/cache_manager.py
```diff
@@ -43,7 +43,7 @@ def __init__(
         ]
         self.free_block_mask = torch.ones(num_blocks, dtype=torch.int32, device="cpu")
         self.slots = torch.arange(
-            0, num_blocks * self.block_size, dtype=torch.int32
+            0, num_blocks * self.block_size, dtype=torch.int64
```
drbh, line 46: quick question, is there a case where `num_blocks` is really really big? just trying to understand the type change

OlivierDehaene: It's because the vllm kernel now ask for this dtype. I don't know why they changed. 
Slots is a very small tensor anyway.

server/text_generation_server/models/custom_modeling/flash_mistral_modeling.py
```diff
@@ -173,10 +173,6 @@ def __init__(
             weights=weights,
             bias=False,
         )
-        self.num_groups = self.num_heads // self.num_key_value_heads
-        self.kv_head_mapping = torch.arange(
-            0, self.num_key_value_heads, dtype=torch.int32, device=weights.device
-        ).repeat_interleave(self.num_groups)
```
OlivierDehaene, line 179: Revert

thomas-schillaci:
Hello @OlivierDehaene, @drbh this pull request slightly changes decoding, and so breaks my integration pipeline (I'm testing multiple inputs on my models and asserting their outputs don't change with do_sample=False).Do you know why this change was needed and if it's going to stay this way?

PR #1718, opened 2024-04-09
feat: explore torch.compile on MLP and other ops

drbh:
WIP 

This PR explores the differences using torch.compile on select ops and starte work on reproducible benches

### DIFF
--- a/server/bench/mlp-compare.py
+++ b/server/bench/mlp-compare.py
@@ -1,0 +1,196 @@
+import torch
+import torch._dynamo.config
+import torch._inductor.config
+from torch.nn import functional as F
+
+import torch
+
+# prefer using pre-compiled ops
+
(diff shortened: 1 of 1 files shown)


flozi00:
Please also check tensorrt as backend
I am using it for whisper and it's multiple times faster than assisted generation with distil whisper 

I think it was around factor 3 with max autotune

drbh:
closing in favor of similar work to integrate precompiled mlp (tested in this PR) to llama. 

Branch here: https://github.com/huggingface/text-generation-inference/tree/llama-fused-compiled-mlp 
PR to follow when stable

flozi00:
Hi, I am still interested in the experiments for compilation of layers to tensorrt.
Could you share some insights why you dropped this feature and the possible speed ups ?

PR #1734, opened 2024-04-12
feat: medusa v2

OlivierDehaene:


server/text_generation_server/utils/layers.py
```diff
@@ -447,12 +447,12 @@ def forward(self, x):
...
 
 class MedusaHead(torch.nn.Module):
-    def __init__(self, config, prefix, weights):
+    def __init__(self, config, medusa_config, prefix, weights):
         super().__init__()
         self.blocks = torch.nn.ModuleList(
             [
                 ResBlock(config, prefix=f"{prefix}.{i}", weights=weights)
-                for i in range(config["medusa_num_layers"])
+                for i in range(medusa_config["medusa_num_layers"])
```
Narsil, line 455: Which should probably take that information from `speculate` instead (to avoid loading layers which we're not going to use)

We can do in another PR

server/text_generation_server/utils/layers.py
```diff
@@ -467,46 +467,159 @@ def forward(self, x):
...
+            weights=weights,
+            bias=True,
+        )
+        self.process_group = weights.process_group
+        self.world_size = self.process_group.size()
+        self.rank = self.process_group.rank()
+
+        self.act = torch.nn.SiLU()
+
+        self.lm_head = TensorParallelHead.load(config, prefix, weights)
```
Narsil, line 549: I thought what I had done was cleverer by passing the `lm_head` directly (which avoids a reload).

Since we're using sharding here Weights will actually load the LM head n times with current code. (iirc)

server/text_generation_server/utils/layers.py
```diff
@@ -467,46 +467,159 @@ def forward(self, x):
...
-                    if k in routing:
-                        raise RuntimeError(
-                            f"Key {k} was found in multiple files: {filename} and {routing[k]}"
-                        )
-                    weights.routing[k] = filename
-
-            medusa = MedusaModel(config, weights)
+            lm_head = None
+            try:
+                medusa = MedusaHeadV2(config, prefix, weights)
```
Narsil, line 606: Shouldn't it be the reverse ? MedusaV2 will always load properly (since it's only a subset of V1) ?

server/text_generation_server/utils/layers.py
```diff
@@ -467,46 +467,159 @@ def forward(self, x):
         return x
 
 
-class SpeculativeHead(nn.Module):
+class MedusaHeadV1(nn.Module):
```
Narsil, line 470: Maybe it's time we put them into their own file (both in the same file is ok I think)?

Narsil (commented):
Nice.

Few nits.

Narsil (dismissed):
LGTM.

Let's move the nits as chore.

fxmarty:
It would be helpful to document it.

PR #1749, opened 2024-04-15
feat: improve temperature logic in chat

drbh:
This PR adds support for `do_sample` to chat to enable greedy sampling


router/src/validation.rs
```diff
@@ -217,7 +217,7 @@ impl Validation {
         }
 
         let temperature = temperature.unwrap_or(1.0);
-        if temperature <= 0.0 {
+        if temperature < 0.0 {
```
Narsil, line 220: No temperature = 0 still needs to yell on the regular URLs. It yields incorrect results if you really use temperature==0 as an exponent.

router/src/server.rs
```diff
@@ -775,6 +776,10 @@ async fn chat_completions(
     let logprobs = logprobs.unwrap_or(false);
     let tool_prompt = tool_prompt.unwrap_or_default();
     let stop = stop.unwrap_or_default();
+    // rescale temperature starting from 0.0 to 1.0
+    let adjusted_temperature = temperature.map_or(1.0, |t| t + 1.0);
```
Narsil, line 780: I do not understand this. Why are you adding 1.0 ?

drbh: 1 is added since openais api is essentially zero indexed. They treat `0` as deterministic and higher values as more random. 

In our case 1 is deterministic and greater is more random, so adding 1 allows users to get deterministic output with temp 0, and more random with larger values.

> temperature:
> What sampling temperature to use, between 0 and 2. Higher values like 0.8 will make the output more random, while lower values like 0.2 will make it more focused and deterministic.

router/src/validation.rs
drbh: this was removed in the latest change since `temperature` should never be 0 at this point

drbh:
tiny reproduction script

```python
import requests

headers = {
    "Content-Type": "application/json",
}

data = {
    "model": "tgi",
    "messages": [
        {
            "role": "user",
            "content": "Summarize the main ideas of Jeff Walker's Product Launch Formula into bullet points as it pertains to a growth marketing agency implementing these strategies and tactics for their clients...",
        }
    ],
    "stream": False,
    "max_tokens": 100,
    "temperature": 0,
}

url = "http://localhost:3000/v1/chat/completions"

response = requests.post(url, headers=headers, json=data)

print(response.text)
```

router/src/server.rs
Narsil: No it's not.

logits.pow(temperature).


temperature == 0 -> Pretty much forces deteministic sampling (with float issues).
temperature == 1 -> Regular distribution.

drbh: makes sense! PR updated to:
1. only set `do_sample` to `false` when `temperature` is 0
2. rescale `temperature` from 0 to 1 only when temp is 0

I believe this fits the requirements, and enables greedy ONLY when temperature is 0. Please lmk if I should change! 🙏

router/src/server.rs
```diff
@@ -1008,6 +1009,10 @@ async fn chat_completions(
     let logprobs = logprobs.unwrap_or(false);
     let tool_prompt = tool_prompt.unwrap_or_default();
     let stop = stop.unwrap_or_default();
+    // enable greedy only when temperature is 0
+    let do_sample = temperature.map_or(true, |t| t > 0.0);
+    let adjusted_temperature = temperature.map_or(1.0, |t| if t == 0.0 { 1.0 } else { t });
+    let temperature = Some(adjusted_temperature);
```
Narsil, line 1015: Isn´t that.
```rust
let temperature = if temperature == 0.0{
   None
   } else{ Some(temperature) } 
```

Narsil: Better yet, make the condition == 0.0 common to both do_sample and temperature (makes it more readable imho).

drbh: I don't think that exactly equivalent since we need the following:

1. do_sample to be true in all cases except if the incoming temp is 0
2. we need temperature to be a `Option<f32>` and we need to replace 0 with 1. 

most readable/simple I've come up with so far is...
```rust
let do_sample: bool = temperature.map_or(true, |t| t != 0.0);
let temperature = temperature.map(|t| if t == 0.0 { 1.0 } else { t });
```

whatcha think? happy to make any changes!

PR #1758, opened 2024-04-17
fix: default use_qk_norm false in cohere

drbh:
This PR simply defaults use_qk_norm to false if its not included in the config

### DIFF
--- a/server/text_generation_server/models/custom_modeling/flash_cohere_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_cohere_modeling.py
@@ -216,7 +216,9 @@ class FlashCohereAttention(torch.nn.Module):
 
         self.query_key_value = load_attention(config, prefix, weights)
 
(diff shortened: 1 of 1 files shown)


server/text_generation_server/models/custom_modeling/flash_cohere_modeling.py
```diff
@@ -216,7 +216,9 @@ def __init__(
 
         self.query_key_value = load_attention(config, prefix, weights)
 
-        self.use_qk_norm = config.use_qk_norm
+        self.use_qk_norm = (
+            config.use_qk_norm if hasattr(config, "use_qk_norm") else False
+        )
```
OlivierDehaene, line 221: I think we should update transformers instead and that should solve it.

drbh:
closing due to upstream change https://github.com/huggingface/transformers/pull/30027

PR #1760, opened 2024-04-18
Add reference to TPU support

brandonroyal:
# What does this PR do?
This PR makes a small addition to the readme that reference new TGI support for TPUs via Optimum TPU (https://huggingface.co/docs/optimum-tpu/howto/serving)


Narsil:
Thanks a lot !

PR #1798, opened 2024-04-23
Add attribute descriptions for `GenerateParameters`

Wauplin:
Once https://github.com/huggingface/huggingface.js/pull/629 gets merged, we will rely on TGI's specs to generate jsonschema types for `text_generation` and `chat_completion`.

This PR adds some documentation for `GenerationParameters`'s properties so that they get documented in the downstream tools (TGI docs but also `huggingface.js`/`huggingface_hub` inference clients). I mostly took inspiration from [the python client](https://github.com/huggingface/text-generation-inference/blob/main/clients/python/text_generation/types.py) for the descriptions.

### DIFF
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -168,9 +168,12 @@ pub struct Info {
 
 #[derive(Clone, Debug, Deserialize, ToSchema, Default)]
 pub(crate) struct GenerateParameters {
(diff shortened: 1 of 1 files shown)


Narsil (approved):
Hero PR :)

PR #1813, opened 2024-04-26
feat: add one line quickstart

drbh:
This PR adds a simple quickstart file that conditionally installs `docker` and `nvidia-container-toolkit` before fetching the TGI docker container and running it with `HuggingFaceH4/zephyr-7b-beta`

It can be run with 

```bash
bash quickstart.sh
```

additional TGI can be installed without cloning the repo

```bash
# using this branch
curl --proto '=https' --tlsv1.2 -sSf \
    https://raw.githubusercontent.com/huggingface/text-generation-inference/add-quickstart-script/quickstart.sh \
    | bash

# if merged
curl --proto '=https' --tlsv1.2 -sSf \
    https://raw.githubusercontent.com/huggingface/text-generation-inference/quickstart.sh \
    | bash
```

### DIFF
--- a/quickstart.sh
+++ b/quickstart.sh
@@ -35,4 +35,4 @@ model="HuggingFaceH4/zephyr-7b-beta"
 volume="$PWD/data"
 
 # Run the Docker container in interactive mode to allow CTRL+C to stop the container
-docker run -it --gpus all --shm-size 1g -p 8080:80 -v "$volume:/data" ghcr.io/huggingface/text-generation-inference:2.0 --model-id "$model"
+docker run -it --gpus all --shm-size 1g -p 8080:80 -v "$volume:/data" ghcr.io/huggingface/text-generation-inference:latest --model-id "$model"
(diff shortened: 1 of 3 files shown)


quickstart.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# Check if Docker is installed
+if ! command -v docker &> /dev/null; then
+  # Install Docker based on the OS
+  if [[ "$OSTYPE" == "linux-gnu"* ]]; then
+    # Ubuntu/Debian-based systems
+    sudo apt-get update && sudo apt-get install -y docker.io
```
Narsil, line 8: Not nearly enough to support all the linux variants I'm afraid :)

Narsil:
I don't think we should go that route, I feel like the `pip install` thing is more what some users might desire, not necessarily 1 liner to run docker (which is going to be annoying to maintain anyway)

PR #1828, opened 2024-04-29
Handle images in chat api

drbh:
This PR allows for messages to be formatted as simple strings, or as an array of objects including image urls. This is done by formatting content arrays into a simple string.

Example using `llava-hf/llava-v1.6-mistral-7b-hf` 

```bash
curl localhost: 3000/v1/chat/completions \
-X POST \
-d '{
    "model": "tgi",
    "messages": [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "Whats in this image?"
                },
                {
                    "type": "image_url",
                    "image_url": {
                        "url": "https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/transformers/rabbit.png"
                    }
                }
            ]
        }
    ],
    "stream": false,
    "max_tokens": 20,
    "seed": 42
}'
```

is equivlant to this more simple request

```bash
curl localhost: 3000/v1/chat/completions \
-X POST \
-d '{
    "model": "tgi",
    "messages": [
        {
            "role": "user",
            "content": "Whats in this image?\n![](https://huggingface.co/datasets/huggingface/documentation-images/resolve/main/transformers/rabbit.png)"
        }
    ],
    "stream": false,
    "max_tokens": 20,
    "seed": 42
}'
```

output
```
# {"id":"","object":"text_completion","created":1714406985,"model":"llava-hf/llava-v1.6-mistral-7b-hf","system_fingerprint":"2.0.1-native","choices":[{"index":0,"message":{"role":"assistant","content":" This is an illustration of an anthropomorphic rabbit in a spacesuit, standing on what"},"logprobs":null,"finish_reason":"length"}],"usage":{"prompt_tokens":2945,"completion_tokens":20,"total_tokens":2965}}%
```

### DIFF
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -928,7 +928,7 @@ mod message_content_serde {
                             "text" => Ok(content.text.unwrap_or_default()),
                             "image_url" => {
                                 if let Some(url) = content.image_url {
-                                    Ok(format!("\n![]({})", url.url))
+                                    Ok(format!("![]({})", url.url))
                                 } else {
                                     Ok(String::new())
                                 }
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -525,7 +525,7 @@ impl ChatCompletion {
     pub(crate) fn new(
         model: String,
         system_fingerprint: String,
-        _output: Option<String>,
+        output: Option<String>,
         created: u64,
         details: Details,
         return_logprobs: bool,
@@ -541,7 +541,7 @@ impl ChatCompletion {
                 index: 0,
                 message: Message {
                     role: "assistant".into(),
-                    content: None,
+                    content: output,
                     name: None,
                     tool_calls,
                 },
@@ -878,7 +878,7 @@ pub(crate) struct SerializedMessage {
 
 #[derive(Clone, Serialize, Deserialize, Default)]
 pub(crate) struct ChatTemplateInputs<'a> {
-    messages: Vec<SerializedMessage>,
+    messages: Vec<Message>,
     bos_token: Option<&'a str>,
     eos_token: Option<&'a str>,
     add_generation_prompt: bool,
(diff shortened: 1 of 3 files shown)


router/src/lib.rs
```diff
@@ -884,12 +884,75 @@ pub(crate) struct ToolCall {
...
+                let results: Result<Vec<String>, _> = arr
+                    .into_iter()
+                    .map(|v| {
+                        let content: Content =
+                            serde_json::from_value(v).map_err(de::Error::custom)?;
+                        match content.r#type.as_str() {
+                            "text" => Ok(content.text.unwrap_or_default()),
+                            "image_url" => {
+                                if let Some(url) = content.image_url {
+                                    Ok(format!("\n![]({})", url.url))
```
Narsil, line 931: Why the extra `\n` is shouldn't be here I think.

router/src/lib.rs
```diff
@@ -884,12 +884,75 @@ pub(crate) struct ToolCall {
...
+}
+
+#[derive(Clone, Deserialize, Serialize, ToSchema, Default, Debug)]
+pub(crate) struct Content {
+    pub r#type: String,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub text: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub image_url: Option<ImageUrl>,
+}
```
Narsil, line 906: Why not use something like

```rust
#[serde(tag = "type")]
enum ContentChunk{
    Text(Text{ text: String}),
    Image(Image {image_url: String })
}
```

?

Narsil (commented):
Good minimal approach.

1- I think we can make better use of serde's typings to get cleaner errors (for instance here you don't have checks in place that either image_url or text is set and you're using unwraps invalidly imho.
2- I think we should keep that proper structure and send it all the way to the Python level so that real markdown can still be parsed as real markdown by the python layer.
3-  Can we add some tests too ?

router/src/lib.rs
Narsil, line 931:
```suggestion
                                    Ok(format!("![]({})", url.url))
```

Narsil (approved):
Approving  since this is functional (with the deletion of the newline) and usable by users.

Let's refactor this though I think (especially to pass around real payloads to avoid the markdown issue).

PR #1850, opened 2024-05-02
[WIP] MLPSpeculator speculative decoding support

JRosenkranz:
# What does this PR do?

**_1. What is the motivation behind this feature? Is it related to a problem or frustration with the library? Is it a feature related to something you need for a project? Is it something you worked on and think it could benefit the community?_**

This feature will improve inference performance significantly with minimal effort from the user by leveraging the existing speculative decoding framework with an improved speculator architecture implementation.

**_2. Describe your requested feature in as much detail as possible. The more you can tell us about it, the better we'll be able to help you._**

This feature includes a new Speculator architecture (MLPSpeculator) to support `flash_causal_lm` models. The architecture itself is similar to medusa, with a key difference that each head conditions its outputs on the prior head. This allows it to produce better formed n-grams and increase acceptance rate.

**_3. Provide a code snippet that demonstrates the features usage._**

TBD

**_4. If the feature is related to a paper, please include a link._**

https://arxiv.org/abs/2404.19124

## Who can review?

@Narsil 

Note: More information will be added to this description as the PR becomes more mature.

Current Speculators:
- [llama3-8b-acclerator](https://huggingface.co/ibm-fms/llama3-8b-accelerator)
- [codellama-13b-accelerator](https://huggingface.co/ibm-fms/codellama-13b-accelerator)
- [llama-13b-accelerator](https://huggingface.co/ibm-fms/llama-13b-accelerator)
- [granite-7b-lab-accelerator](https://huggingface.co/ibm/granite-7b-lab-accelerator)

Reference implementation can be found: https://github.com/foundation-model-stack/fms-extras

### DIFF
--- a/server/text_generation_server/utils/layers.py
+++ b/server/text_generation_server/utils/layers.py
@@ -442,69 +442,46 @@ class ResBlock(torch.nn.Module):
     def forward(self, x):
         return x + self.act(self.linear(x))
 
-class LayerNormParameterized(nn.Module):
+class MLPSpeculatorLayerNorm(nn.Module):
     """
-    A generalized LayerNorm implementation. With all optional arguments set to True, equivalent to nn.LayerNorm up to epsilon stabilization term
-    (this class divides inputs by min(norm, eps), while nn.LayerNorm divides by norm + eps).
+    A L2 normalization implementation
     ...
     Args
     ----
     normalized_shape : int
         Dimensionality of input data (size of final tensor axis)
+    elementwise_scale_weight : torch.Tensor
+        learned scaling term after normalization?
+    elementwise_shift_bias : torch.Tensor
+        learned bias term after normalization?
     eps : float
         Safety term to prevent division by zero. Make sure the chosen value fits in the range of your encoding scheme (i.e. fp16 requires eps >= 6e-8).
-    elementwise_scale : bool
-        Include a learned scaling term after normalization?
-    elementwise_shift : bool
-        Include a learned bias term after normalization?
-    use_mean : bool
-        Recenter inputs around zero before normalizing, or just rescale?
+
     """
 
     def __init__(
         self,
         normalized_shape,
+        elementwise_scale_weight: torch.Tensor,
+        elementwise_shift_bias: torch.Tensor,
         eps=1e-06,
-        elementwise_scale=True,
-        elementwise_shift=False,
-        use_mean=False,
-        use_high_precision_pow=False,
     ):
-        super(LayerNormParameterized, self).__init__()
+        super(MLPSpeculatorLayerNorm, self).__init__()
         self.normalized_shape = normalized_shape
+        self.weight = nn.Parameter(elementwise_scale_weight)
+        self.bias = nn.Parameter(elementwise_shift_bias)
         self.eps = eps
-        self.elementwise_scale = elementwise_scale
-        self.elementwise_shift = elementwise_shift
-        self.use_mean = use_mean
-        self.use_high_precision_pow = use_high_precision_pow
-
-        if self.elementwise_scale:
-            self.weight = nn.Parameter(torch.empty(self.normalized_shape))
-        if self.elementwise_shift:
-            self.bias = nn.Parameter(torch.empty(self.normalized_shape))
-
-    def reset_parameters(self):
-        if self.elementwise_scale:
-            self.weight.data.fill_(1)
-        if self.elementwise_shift:
-            self.bias.data.zero_()
 
     def forward(self, x):
-        if self.use_mean:
-            x = x - x.mean(-1, keepdim=True)
         xf = x
-        if self.use_high_precision_pow:
-            xf = x.float()
         xf = xf * torch.rsqrt(xf.pow(2).mean(-1, keepdim=True) + self.eps)
         x = xf.type_as(x)
-        if self.elementwise_scale:
-            x = self.weight * x
-        if self.elementwise_shift:
-            x = x + self.bias
+        x = self.weight * x
+        x = x + self.bias
         return x
 
 class MLPSpeculatorModel(torch.nn.Module):
-    def __init__(self, config, emb, proj, head, ln):
+    def __init__(self, config, prefix, weights):
         super().__init__()
         self.config = config
         self.n_predict = config.n_predict
@@ -512,10 +489,30 @@ class MLPSpeculatorModel(torch.nn.Module):
         inner_dim = config.inner_dim if config.inner_dim != 0 else config.emb_dim
         self.inner_dim = inner_dim
         self.config = config.vocab_size
-        self.emb = emb
-        self.proj = proj
-        self.head = head
-        self.ln = ln
+        self.emb = nn.ModuleList(
+            [TensorParallelEmbedding(f"{prefix}.emb.{i}", weights) for i in range(config.n_predict)]
+        )
+        self.proj = TensorParallelColumnLinear.load_multi(
+            config,
+            prefixes=[f"{prefix}.proj.{i}" for i in range(config.n_predict)],
+            weights=weights,
+            bias=False,
+            dim=0
+        )
+        self.head = nn.ModuleList(
+            [TensorParallelRowLinear.load(config, f"{prefix}.head.{i}", weights, bias=False) for i in range(config.n_predict)]
+        )
+        self.ln = nn.ModuleList(
+            [
+                MLPSpeculatorLayerNorm(
+                    config.inner_dim,
+                    elementwise_scale_weight=weights.get_tensor(f"{prefix}.ln.{i}.weight"),
+                    elementwise_shift_bias=weights.get_tensor(f"{prefix}.ln.{i}.bias")
+                )
+                for i in range(config.n_predict)
+            ]
+        )
+
         # Weights ensure that state_0 accounts for 50% of state magnitude by final head in expectation
         self.state_weight = 0.5 ** (0.5 / self.n_predict)
         self.emb_weight = math.sqrt(1 - self.state_weight ** 2)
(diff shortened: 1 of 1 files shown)


server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+    """
+
+    def __init__(
+        self,
+        normalized_shape,
+        eps=1e-06,
+        elementwise_scale=True,
+        elementwise_shift=False,
+        use_mean=False,
+        use_high_precision_pow=False,
```
Narsil, line 471: Any way we could set on a definite set of values ?

Having lots of if/else in the modeling code is great when tinkering, but super confusing when running actual models, and usually only 1 set is used most of the time.
Having different pathways makes optimization also harder if we were to replace things with an actual kernel.

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+            z = self.emb[i](ind)
+            z = z.mul(self.emb_weight * math.sqrt(self.inner_dim / 2))  # b k d
+            state = self.proj[i](state) * self.state_weight + z
+            state = self.activation(self.ln[i](state))  # b k d
+            _probs = F.log_softmax(self.head[i](state), dim=2)  # b k v
+            probs, preds = _probs.topk(top_k_tokens_per_head[i], dim=2)  # b k k'
+
+            # Update candidate set with new predictions
+            out = out.unsqueeze(2).expand(-1, -1, top_k_tokens_per_head[i], -1)  # b k k' h
+            out = torch.cat([out, preds.unsqueeze(3)], dim=3)  # b k k' h+1
```
Narsil, line 548: I think we can prevent all those cats by prellocating the tensor directly (and indexing directly into it, that will save a lot of tensor allocations)

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+        out = torch.empty(b, 1, 0, device=state.device).int()  # b k h
+        log_probs = torch.zeros(b, 1, device=state.device)  # b k
+        all_probs = torch.empty(b, 1, 0, self.vsize, device=state.device)  # b k h v
+        assert (
+                len(top_k_tokens_per_head) == self.n_predict
+        ), f"You must provide a topk number for each head ({self.n_predict} heads, {len(top_k_tokens_per_head)} provided)"
+        for i in range(self.n_predict):
+            # Project and predict
+            z = self.emb[i](ind)
+            z = z.mul(self.emb_weight * math.sqrt(self.inner_dim / 2))  # b k d
```
Narsil, line 540: Couldn't this multiplication be done at init time ?

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+            log_probs = log_probs.add(probs).reshape(b, -1)  # b kk'
+
+        # Take only top n best guesses
+        best_guesses = log_probs.topk(num_candidates, dim=1)[1]  # b k
+        return all_probs.gather(
+            1, best_guesses[:, :, None, None].expand(-1, -1, self.n_predict, self.vsize)
+        )  # b n h v
+
+    def load(self, config, prefix, weights):
+        self.emb = nn.ModuleList(
```
Narsil, line 569: Let's remove all these torch allocations, and use the weights directly coming from the weights.

Doing this saves a lot of VRAM (because f32 by default in torch) and also makes bugs easier to detect (because you'll get error missing tensor instead of potentially silently keeping some tensors non set)

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+        if self.elementwise_scale:
+            self.weight = nn.Parameter(torch.empty(self.normalized_shape))
+        if self.elementwise_shift:
+            self.bias = nn.Parameter(torch.empty(self.normalized_shape))
+
+    def reset_parameters(self):
+        if self.elementwise_scale:
+            self.weight.data.fill_(1)
+        if self.elementwise_shift:
+            self.bias.data.zero_()
```
Narsil, line 490: Unused function, no ?

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+        self.eps = eps
+        self.elementwise_scale = elementwise_scale
+        self.elementwise_shift = elementwise_shift
+        self.use_mean = use_mean
+        self.use_high_precision_pow = use_high_precision_pow
+
+        if self.elementwise_scale:
+            self.weight = nn.Parameter(torch.empty(self.normalized_shape))
+        if self.elementwise_shift:
+            self.bias = nn.Parameter(torch.empty(self.normalized_shape))
```
Narsil, line 484: You should probably be getting the tensors from the weights directly. Torch allocates those tensors as f32 which is 2x bigger than necessary for f16 (more for quantized models)

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+    def load(speculator_config, prefix: str, weights):
+        from pathlib import Path
+        from safetensors import safe_open
+
+        speculator_path = speculator_config.use_speculator
+
+        filename = str(Path(speculator_path) / "*.safetensors")
+
+        routing = weights.routing
+        with safe_open(filename, framework="pytorch") as f:
```
Narsil, line 624: This should probably be done before (weights should contain all the necessary weights directly and all file handles).

This is done like that in medusa so let's keep it like that in this PR, but it will probably move at some point.

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+
+        # Take only top n best guesses
+        best_guesses = log_probs.topk(num_candidates, dim=1)[1]  # b k
+        return all_probs.gather(
+            1, best_guesses[:, :, None, None].expand(-1, -1, self.n_predict, self.vsize)
+        )  # b n h v
+
+    def load(self, config, prefix, weights):
+        self.emb = nn.ModuleList(
+            [nn.Embedding(config.vocab_size, config.inner_dim) for _ in range(config.n_predict)]
```
Narsil, line 570: `TensorParallelEmbedding` should save a bit of VRAM here (quite important for large vocab models)

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+        )
+        self.proj = nn.ModuleList(
+            [
+                nn.Linear((config.emb_dim if i == 0 else config.inner_dim), config.inner_dim, bias=False)
```
Narsil, line 574: How many such layers are there ?
We should probable TP parallelize all this computation to make it efficient.

server/text_generation_server/utils/layers.py
```diff
@@ -606,24 +800,35 @@ def forward(self, x):
...
+
+            with open(speculator_config, "r") as f:
+                speculator_config = json.load(f)
             lm_head = None
-            try:
-                medusa = MedusaHeadV1.load(config, prefix, weights)
-            except:
-                medusa = MedusaHeadV2(config, prefix, weights)
+
+            architecture = speculator_config["architectures"][0]
```
Narsil, line 818: This will fail on medusa (which doesn't define an architecture). We need to `try..except` or ` get("architecture")..`.

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+        )
+        for i in range(config.n_predict):
+            self.emb[i].weight.data.copy_(weights.get_tensor(f"{prefix}.emb.{i}.weight"))
+            self.proj[i].weight.data.copy_(weights.get_tensor(f"{prefix}.proj.{i}.weight"))
+            self.ln[i].weight.data.copy_(weights.get_tensor(f"{prefix}.ln.{i}.weight"))
+            self.ln[i].bias.data.copy_(weights.get_tensor(f"{prefix}.ln.{i}.bias"))
+            self.head[i].weight.data.copy_(weights.get_tensor(f"{prefix}.head.{i}.weight"))
+
+
+class MLPSpeculatorHeadV1(nn.Module):
```
Narsil, line 597: No need to specify V1 if there's no V2. But as you prefer.

server/text_generation_server/utils/layers.py
```diff
@@ -439,6 +442,197 @@ def __init__(self, config, prefix, weights):
...
+        return logits, speculative_logits
+
+    @staticmethod
+    def load(speculator_config, prefix: str, weights):
+        from pathlib import Path
+        from safetensors import safe_open
+
+        speculator_path = speculator_config.use_speculator
+
+        filename = str(Path(speculator_path) / "*.safetensors")
```
Narsil, line 621: Isn't there a filename already ?

server/text_generation_server/utils/layers.py
```diff
@@ -606,24 +800,35 @@ def forward(self, x):
...
         self.head = lm_head
-        self.medusa = medusa
+        self.speculator = speculator
 
     @staticmethod
     def load(config, prefix: str, weights):
-        use_medusa = config.use_medusa
-        if use_medusa:
+        use_speculator = config.use_speculator
+        if use_speculator:
```
Narsil, line 811: I feel like this should switch from a bool to a true enum value directly (making the modifications more extensive).

`config.speculator = {None, "medusa", "MLPS"}` basically.

Narsil (commented):
Do you have link to the actual models so we can try this out and see how is the performance ?

I made a lot of comments about the overall structure. I'm happy to help and make some modifications (either here or keep things for follow up PRs).
Some of the stuff is more because we didn't use to have multiple speculator, so now is a good time to clean things up a bit, but not really the main goal of this PR.

JRosenkranz:
> Do you have link to the actual models so we can try this out and see how is the performance ?
> 
> I made a lot of comments about the overall structure. I'm happy to help and make some modifications (either here or keep things for follow up PRs). Some of the stuff is more because we didn't use to have multiple speculator, so now is a good time to clean things up a bit, but not really the main goal of this PR.

Sure! Feel free to make changes in this PR. This was just a first pass, however I figured we were going to have to make some changes. Here are the links to the current speculators we have:

https://huggingface.co/ibm-fms/llama3-8b-accelerator
https://huggingface.co/ibm-fms/codellama-13b-accelerator
https://huggingface.co/ibm-fms/llama-13b-accelerator
https://huggingface.co/ibm/granite-7b-lab-accelerator

server/text_generation_server/utils/layers.py
JRosenkranz: I don't think it was ever a bool? it looked like it was a string path pointing to the speculator directory containing the config and weights?

```python
medusa_config = str(Path(use_medusa) / "config.json")
filename = str(Path(use_medusa) / "medusa_lm_head.safetensors")
```

We could keep make this an enum, or we can just keep as path and recover the class from the config? If this is possible with the current config.

server/text_generation_server/utils/layers.py
JRosenkranz: This looked consistent with what medusa was doing

```python
medusa_config = str(Path(use_medusa) / "config.json")
filename = str(Path(use_medusa) / "medusa_lm_head.safetensors")
```

Also, there could be multiple safetensors files. Let me know if there is another way I should be getting the path.

server/text_generation_server/utils/layers.py
JRosenkranz: I agree, the number of layers is based on the `n_predict` parameter to the model.

server/text_generation_server/utils/layers.py
Narsil: I meant why is there a `*` ? Shouldn't it be a real name for the file ?

JRosenkranz: Currently it looks something like this:

```
model-00001-of-00002.safetensors
model-00002-of-00002.safetensors
```

server/text_generation_server/utils/layers.py
daviswer: We can rescale `self.emb_weight` at init time, but that's just a single scalar so I assumed it wouldn't matter much. Rescaling `z=self.emb` itself is a little more problematic because we keep our layers initialized to the same scale for training purposes - we'd have to adjust the weights of our trained models to accommodate a direct rescaling of the embedding layers. What we could do though is fuse this multiply with the add that comes afterwards? Assuming I'm understanding your thought process correctly

Narsil: >  it wouldn't matter much. 

It does because it launches a new kernels for both ops. Not critical, but still something we try to avoid as much as possible (it's usually trivial)

JRosenkranz:
Closing as this has been merged via https://github.com/huggingface/text-generation-inference/pull/1865

PR #1243, opened 2023-11-07
Add RoCm support

fxmarty:
This PR adds support for AMD Instinct MI210 & MI250 GPUs, with paged attention and FAv2 support.

Remaining items to discuss, on top of possible others:
* Should we have a `ghcr.io/huggingface/text-generation-inference:1.1.0+rocm` hosted image, or is it too early?
* Should we set up a CI on MI210/MI250? I don't have access to the runners of TGI though.
* Are we comfortable with those changes being directly in TGI, or do we need a fork?

PR #1866, opened 2024-05-06
Refactor layers.

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


server/text_generation_server/utils/paged_attention.py
```diff
@@ -1,13 +1,9 @@
...
-    IS_CUDA_SYSTEM,
-    IS_ROCM_SYSTEM,
-    IS_XPU_SYSTEM,
-)
+from text_generation_server.utils.import_utils import SYSTEM
 
 _PARTITION_SIZE = 512
 
-if IS_XPU_SYSTEM:
+if SYSTEM == "xpu":
```
fxmarty, line 6: What's the motivation?

server/text_generation_server/utils/flash_attn.py
```diff
@@ -2,24 +2,54 @@
...
 HAS_FLASH_ATTN_V2_CUDA = False
 HAS_FLASH_ATTN_V2_ROCM = False
 
-if IS_XPU_SYSTEM:
+if SYSTEM == "xpu":
     import intel_extension_for_pytorch as ipex
 
-if IS_CUDA_SYSTEM or IS_ROCM_SYSTEM:
+    def attention(
+        q,
```
fxmarty, line 18: I suggested to do this kind of thing at some point but @OlivierDehaene did disagree with this design: https://github.com/huggingface/text-generation-inference/pull/1243#discussion_r1386307382

```
if IS_X_BACKEND:
    def foo():
        ...
else:
    def foo():
        ...

server/text_generation_server/utils/paged_attention.py
Narsil: Having a real enum, those are exclusive.
Meaning we cannot have both IS_XPU and IS_AMD at the same time (which was the case for the bools).

Narsil: String comparison is because python lacks good enum support unfortunately.

fxmarty: very nitpick but `Enum System` couldn't work? `SYSTEM == System.CUDA`?

Narsil: Sure it can, enum are just painful to work with in Python. I just avoid them by default.
Had too many issues with them, where simple strings are just as fine (biggest caveat being the potential mistyping something, runtime is just as fast since strings are immutable)

Narsil: Have fun https://github.com/python/cpython/issues/93910  :)

PR #1902, opened 2024-05-15
[Bug Fix] Update torch import reference in bnb quantization

DhruvSrikanth:
# What does this PR do?

Fixes `Import Error` occurring from mismatch of usage between torch.nn.Module and nn.Module.


Narsil (approved):
LGTM, thanks for the fix !

PR #1923, opened 2024-05-20
ROCm: Make CK FA2 default

fxmarty:
As per title.

Triton autotune overhead is prohibitive, as it needs to be done for each different prompt length.


fxmarty:
(merging as needed for tomorrow)

PR #1924, opened 2024-05-20
ROCm: make CK FA2 default instead of Triton

fxmarty:
As per title.

Triton autotune overhead is prohibitive, as it needs to be done for each different prompt length.


fxmarty:
(will merge directly for tomorrow)

PR #1929, opened 2024-05-21
Creating doc automatically for supported models.

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


server/text_generation_server/models/__init__.py
```diff
@@ -110,6 +111,138 @@
...
+        "type": "phi3",
+        "name": "Phi 3",
+        "url": "https://huggingface.co/microsoft/Phi-3-mini-4k-instruct",
+    }
+    LLAMA = {
+        "type": "llama",
+        "name": "Llama",
+        "url": "https://huggingface.co/meta-llama/Meta-Llama-3-8B-Instruct",
+    }
+    BAICHUAN = {
```
osanseviero, line 170: Note to self: this is not supported in transformers

server/text_generation_server/models/__init__.py
```diff
@@ -110,6 +111,138 @@
...
+        "name": "Gemma",
+        "url": "https://huggingface.co/google/gemma-7b",
+    }
+    COHERE = {
+        "type": "cohere",
+        "name": "Cohere",
+        "url": "https://huggingface.co/CohereForAI/c4ai-command-r-plus",
+    }
+    DRBX = {
+        "type": "drbx",
```
osanseviero, line 186:
```suggestion
        "type": "dbrx",
```

server/text_generation_server/models/__init__.py
```diff
@@ -110,6 +111,138 @@
     __all__.append(Mamba)
 
 
+class ModelType(enum.Enum):
+    MAMBA = {
+        "type": "ssm",
```
osanseviero, line 116: To be consistent with `transformers` https://huggingface.co/state-spaces/mamba-130m-hf/blob/main/config.json

```suggestion
        "type": "mamba",
```

server/text_generation_server/models/__init__.py
```diff
@@ -110,6 +111,138 @@
...
+        "name": "Mamba",
+        "url": "https://huggingface.co/state-spaces/mamba-2.8b-slimpj",
+    }
+    GALACTICA = {
```
osanseviero, line 120: I think this is `opt` as well based on the config file?

server/text_generation_server/models/__init__.py
```diff
@@ -110,6 +111,138 @@
...
+        "type": "galactica",
+        "name": "Galactica",
+        "url": "https://huggingface.co/facebook/galactica-120b",
+    }
+    SANTACODER = {
```
osanseviero, line 125: Note to self: not supported in `transformers`

server/text_generation_server/models/__init__.py
```diff
@@ -110,6 +111,138 @@
     __all__.append(Mamba)
 
 
+class ModelType(enum.Enum):
+    MAMBA = {
+        "type": "ssm",
+        "name": "Mamba",
+        "url": "https://huggingface.co/state-spaces/mamba-2.8b-slimpj",
```
osanseviero, line 118: Maybe `example_repo` or something like that?

osanseviero (commented):
Looking very nice!

server/text_generation_server/models/__init__.py
Narsil: Probably not. I'm not familiar with the codebase, but we tend to remove a lot of ifs from modeling code, meaning than in TGI, some model types which are the same in transformers , are different here (like gpt_bigcode which has models declared as gpt2 but the modeling is actually quite different).

server/text_generation_server/models/__init__.py
Narsil: It's just to output a link in the doc. Could be the paper page or anything. In here I'd like to keep things minimal.

server/text_generation_server/models/__init__.py
Narsil: Indeed.

Narsil: This is the original one: https://huggingface.co/state-spaces/mamba-2.8b-slimpj

Narsil:
We never remove that old logic. Simple current deployments might depend on it.
But they are indeed considered legacy.

Narsil:
I tried enabling the hf format for mamba, every tensor changed around so we have to redo the modeling code. Not fun (and not important)

PR #1940, opened 2024-05-23
[Major Change][Undecided yet] Move to FlashDecoding instead of PagedAttention kernel.

Narsil:
# What does this PR do?

This PR proposes a long standing change, which is to move towards using FlashDecoding instead of PagedAttention.

- FlashDecoding is supposedly faster than paged (on par at best in early testing, but needs for thorough testing).
- More importantly it will unlock lots of new use cases for much faster speedups (not in current PR).

FlashDecoding defines it's signature as (query, cu_seqlen_q) + (kv, cu_seqlen_kv) + block_tables (to simplify).
This means we can in a single attention pass, merge prefill and decodes, but most importantly we can have huge query_lengths at query times. With current paged kernels, 
there is a hard assumption that Q lengths = 1. For medusa speculation, we're currently faking it by duplicating "queries" in the query slots and adjusting input_lengths and slots.
The longer the query the more wasteful it is (which is ok for small sizes).

With FlashDecoding the expected upsides are:

- Overall speedups since FlashDecoding is supposed to be faster.
- KV-cache hits (to prevent recomputing kv-cache for common prefixes like system messages or assistant prompts, could have been done with Paged only).
- Cleaner+Faster with Speculation methods.
- Insane speedups for harshly constrained grammars (like JSON, fill-a-hole kind of prompting).

Current takeaways:

- Speed improvement is not really there it's ISO with Paged at best, and has slightly worse scaling seqlen or batch_size (5-10% slower) (potentially linked to the fact that FlashDecoding from FA2 implements only block_size=256, which we may be able to update).
- Not all the code has been adapted, meaning there are still a few optimizations left (at every call size we recreate cu_seqlen_{q,kv} which is more kernels and more overhead independant of FA's performance.
- API is much cleaner of varlen queries (even with KV cache).

Why not FashInfer (or others)?

API for KV-cache is much more different than FD : https://docs.flashinfer.ai/api/python/prefill.html#batch-prefill-append-attention
It requires some scratch buffers (of unclear size) , and keeping hold of them.
It requires very different bookkeeping splitting the pages and last_page_indices separately (meaning lot more changes to get it to work).

Layout being the same has FD, it can be explored if performance is there (API would allow similar features).

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


router/src/infer.rs
```diff
@@ -69,7 +69,16 @@ impl Infer {
         tokenizer_config: HubTokenizerConfig,
     ) -> Self {
         // Infer shared state
-        let queue = Queue::new(requires_padding, 16, window_size, speculate);
+        let flashdecoding = if let Ok(flashdecoding) = std::env::var("FLASH_DECODING") {
+            matches!(flashdecoding.to_lowercase().as_str(), "1" | "true")
+        } else {
+            false
+        };
+        let block_size = if flashdecoding { 256 } else { 32 };
```
Narsil, line 77:
```suggestion
        let block_size = if flashdecoding { 256 } else { 16 };
```

server/Makefile-flash-att-v2
```diff
@@ -1,11 +1,11 @@
-flash_att_v2_commit_cuda := 23e8fa5a263d1c7122bc46a86ef32030ee7130f9
+flash_att_v2_commit_cuda := v2.5.8
```
Narsil, line 1: We can actually pip install now.

server/text_generation_server/models/custom_modeling/flash_cohere_modeling.py
```diff
@@ -464,6 +469,24 @@ def forward(
...
+                dtype=torch.int32,
+            )
+            cu_seqlen_k = torch.cat(
+                [
+                    torch.zeros(
+                        (1,), device=input_lengths.device, dtype=input_lengths.dtype
+                    ),
+                    input_lengths.cumsum(dim=-1),
+                ]
+            ).to(dtype=torch.int32)
```
danieldk, line 485: Not sure if this is premature optimization, but saves two allocations:

```python
cu_seqlen_k = torch.empty(input_lengths.size(-1) + 1, device=input_lengths.device, dtype=torch.int32)
cu_seqlen_k[0] = 0
torch.cumsum(input_lengths, -1, out=cu_seqlen_k[1:])
```

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -368,6 +373,23 @@ def forward(
         cos, sin = self.layers[0].self_attn.rotary_emb.get_cos_sin(
             position_ids, max_s, hidden_states.dtype
         )
+        if cu_seqlen_prefill is None and FLASH_DECODING:
```
danieldk, line 376: Maybe we could do this in the `paged_attention` function? Then it has a non-ambiguous signature and we don't have to add this to all the models.

server/text_generation_server/layers/attention/cuda.py
```diff
@@ -32,7 +40,8 @@ def paged_attention(
     kv_head_mapping: torch.Tensor,
     softmax_scale: float,
     block_tables: torch.Tensor,
-    input_lengths: torch.Tensor,
+    cu_seqlen_q: torch.Tensor,
```
danieldk, line 43:
```suggestion
    cu_seqlen_q: Optional[torch.Tensor],
```

server/text_generation_server/models/custom_modeling/flash_gemma_modeling.py
```diff
@@ -253,6 +253,7 @@ def forward(
                 self.kv_head_mapping,
                 self.softmax_scale,
                 block_tables,
+                None,
```
danieldk, line 256: Breaks when flash decoding is enabled?

danieldk (commented):
Looks like a pretty straightforward change, added some comments.

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
Narsil: The problem is that the tensor creation adds too much overhead.

I did it that way initially  and the performance were worse than raw paged just because of that.

Narsil: We could also maybe to it all the way up in `flash_causal_lm`. That was my next best idea (but I don't like obfuscating tensor content since then the tensors might be either `cu_seqlen_q` and cu_seqlen_k` or `None` and `input_lengths` (we could dataclass stuff and all sorts of shenanigans, still obfuscation I feel.

Given the totally optional nature of flash decoding for now, I'm ok if this lives into this particular modeling code while we test, and either rollback or finish the work and put everything into causal_lm once there's only 1 format (biggest drawback will be AMD and intel which do not support FA2 with paged afaik)

danieldk (approved):

PR #1951, opened 2024-05-24
Fix GPTQ for models which do not have float16 at the default dtype

danieldk:
# What does this PR do?

Fix GPTQ for models which do not have float16 at the default dtype

Before this change GPTQ models would not work if the model's default
data type is not `float16`. For example, Gemma GPTQ models would fail
because the default dtype of Gemma is `bfloat16`. There are two issues:

1. If the default `dtype` is not `float16`, the quantizer's `float16`
   parameters get converted to that dtype. The kernels cannot deal
   with non-`float16` types.

This change resolves this issue by excluding quantizer parameters
from data type conversions.

2. Quantized models will typically have `float16` parameters. However,
   the default dtype was set to model's default. So, if a quantized
   Gemma uses `float16`, all parameters are converted to `bfloat16` since
   it is the model's default. This fails in quantized gemm, because it
   expects `float16` arguments.

This is resolved by setting the dtype of gptq/awq-quantized models to
`float16`. (We cannot use `torch_dtype` from the config, because it
often does not correspond to the dtype of the parameters.)

Also see #1953.

**Note:** draft, just checking if all models still pass.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/integration-tests/models/__snapshots__/test_flash_gemma_gptq/test_flash_gemma_gptq.json
+++ b/integration-tests/models/__snapshots__/test_flash_gemma_gptq/test_flash_gemma_gptq.json
@@ -1,0 +1,89 @@
+{
+  "details": {
+    "best_of_sequences": null,
+    "finish_reason": "length",
+    "generated_tokens": 10,
+    "prefill": [
+      {
+        "id": 2,
+        "logprob": null,
+        "text": "<bos>"
+      },
+      {
+        "id": 2015,
+        "logprob": -9.640625,
+        "text": "Test"
+      },
+      {
+        "id": 3853,
+        "logprob": -10.34375,
+        "text": " request"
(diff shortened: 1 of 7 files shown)


Narsil (approved):
LGTM.

I'm not a big fan of the `to_dtype` kwargs (because it's hard to reflect about).

IMHO this should be dealt with clearer API.

`_raw_get_xxx`  Raw getting the tensor (no to device, no dtype).
`get_xxx`  moves the dtypes and the device.

Or something along those lines.
This is ok for now but I know `to_device` is next :)

danieldk:
> Or something along those lines. This is ok for now but I know `to_device` is next :)

True :). I'll close this PR and will one with the suggested `raw` methods later.

PR #1975, opened 2024-05-29
feat: adjust attn weight loading logic

drbh:
This PR updates `load_attention` to prefer loading specific attention based on the model type. Additionally there were two cases where `TensorParallelColumnLinear.load_multi` was called and this reduces it to a single path


Narsil (approved):
LGTM

PR #1980, opened 2024-05-30
Update documentation version to 2.0.4

fxmarty:
As per title

cc @Narsil


Narsil (approved):
Thanks, this will do, but I'm so much in favor of automating this.

Too many brain cycles on this.

PR #1981, opened 2024-05-30
router: send the input as chunks to the backend

danieldk:
# What does this PR do?

Inputs are currently sent to the backend as a single string, encoding images as Base64 and packing them in Markdown-style links. We would like to switch to chunked inputs, with different chunk types corresponding to different modalities. This reduces the amount of input-parsing that needs to be done and increases safety by properly typing inputs. This change can be broken up in various steps:

1. Update the router to send chunks along with 'stringly-typed' inputs.
2. Update the backend(s) to switch to processing chunks rather than strings.
3. Update the router to directly use chunked inputs for APIs that support submitting chunks (e.g. OpenAI), rather than first joining chunks and splitting them again in the input preparation.
4. Deprecate stringly-typed inputs at some point, by removing support from the router.

This change implements the first step, adding a new chunked input representation that separates text chunks from images chunks. Image chunks contain binary data (for smaller message sizes) and the image's MIME type.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


router/client/src/client.rs
```diff
@@ -113,18 +117,36 @@ impl Client {
...
             if n_tokens == 0 {
                 // 1 request is enough to test vision heads.
                 // Sending images on other queries messes up easily with truncation.
-                inputs.push_str("![](data:image/jpeg;base64,iVBORw0KGgoAAAANSUhEUgAAABQAAAAUCAIAAAAC64paAAABg2lDQ1BJQ0MgcHJvZmlsZQAAKJF9kT1Iw0AcxV/TSotUROxQxCFDdbKLijjWKhShQqgVWnUwufQLmrQkKS6OgmvBwY/FqoOLs64OroIg+AHi7OCk6CIl/i8ptIjx4Lgf7+497t4BQqvKNDOQADTdMjKppJjLr4rBVwQQwhAERGVm1uckKQ3P8XUPH1/v4jzL+9yfY0AtmAzwicQJVjcs4g3imU2rznmfOMLKskp8Tjxh0AWJH7muuPzGueSwwDMjRjYzTxwhFks9rPQwKxsa8TRxTNV0yhdyLquctzhr1Qbr3JO/MFzQV5a5TnMUKSxiCRJEKGiggiosxGnVSTGRof2kh3/E8UvkUshVASPHAmrQIDt+8D/43a1ZnJp0k8JJoO/Ftj/GgOAu0G7a9vexbbdPAP8zcKV3/bUWMPtJerOrxY6AwW3g4rqrKXvA5Q4QfarLhuxIfppCsQi8n9E35YHhW6B/ze2ts4/TByBLXaVvgINDYLxE2ese7w719vbvmU5/PycecohsjayNAAAACXBIWXMAAC4jAAAuIwF4pT92AAAAB3RJTUUH6AQIEQMnlTSSjwAAABl0RVh0Q29tbWVudABDcmVhdGVkIHdpdGggR0lNUFeBDhcAAAASSURBVDjLY2AYBaNgFIyCoQsABMQAAeRw1DoAAAAASUVORK5CYII=)");
+                inputs.push_str(&format!(
+                    "![](data:image/jpeg;base64,{})",
+                    WARMUP_IMAGE_BASE64
```
Narsil, line 141:
```suggestion
                    "![](data:image/jpeg;base64,{WARMUP_IMAGE_BASE64})",
```

Narsil: nit

router/src/validation.rs
```diff
@@ -964,4 +961,76 @@ mod tests {
...
+                    patch_size: 1,
+                },
+                image_grid_pinpoints: vec![],
+            }),
+            Config::Paligemma(Paligemma {
+                text_config: PaliTextConfig {
+                    num_image_tokens: 1,
+                },
+            }),
+        ];
```
Narsil, line 996: I'm never a big fan of parameterized tests.

The forloops tend to obscure the logic (and for me a test is only valuable if it's extremely readable and that the reader can infer quite simply if the test is wrong and needs an update, or if the business logic is wrong and bug needs to be fixed).

Here for instance I don't think we're actually testing anything different, since they all endup in the same codepath no ?

router/client/src/lib.rs
```diff
@@ -44,3 +48,37 @@ impl From<transport::Error> for ClientError {
...
+impl From<Vec<Chunk>> for Input {
+    fn from(chunks: Vec<Chunk>) -> Self {
+        Input {
+            chunks: chunks
+                .into_iter()
+                .map(|c| InputChunk { chunk: Some(c) })
+                .collect(),
+        }
+    }
+}
```
Narsil, line 63: This means an entire realloc of the full vec most of the time.

I think we can handle ergonomics a bit better than havnig to realloc all the time, no ?

router/client/src/client.rs
```diff
@@ -113,18 +117,36 @@ impl Client {
         while n_tokens < max_prefill_tokens {
             let truncate = min(max_input_length, max_prefill_tokens - n_tokens);
 
+            let mut input_chunks = Vec::new();
+            input_chunks.push(Chunk::Text(
+                "_test ".to_string().repeat(max_input_length as usize),
+            ));
```
Narsil, line 123: What about implementing `Chunk` into `InputChunk`, that would help ergonomics and is more likely to be inlined by the compiler no ?

Narsil (commented):
Very nice.

This PR doesn't touch shard side, I'm guessing you're leaving that as a follow-up ?

danieldk:
> This PR doesn't touch shard side, I'm guessing you're leaving that as a follow-up ?

Yeah #1985 will do that.

router/client/src/client.rs
danieldk: It complicates the string-conversion a bit, since we can theoretically encounter `InputChunk { chunk: None }`, but since we are creating the chunks in the router anyway, it's probably not a big issue.

Changed it to `From<InputChunk> for Chunk`.

router/client/src/client.rs
danieldk: Fixed.

router/client/src/lib.rs
```diff
@@ -44,3 +47,34 @@ impl From<transport::Error> for ClientError {
...
+        let mut output = String::new();
+        self.iter().for_each(|c| match &c.chunk {
+            Some(Chunk::Text(text)) => output.push_str(text),
+            Some(Chunk::Image(Image { data, mimetype })) => {
+                let encoded = STANDARD.encode(data);
+                output.push_str(&format!("![](data:{};base64,{})", mimetype, encoded))
+            }
+            // We don't create empty chunks, probably better to be robust
+            // than unreachable!().
+            None => {}
```
danieldk, line 76: Just wanted to make sure you see this. Since we are creating the chunks in the router, this case seems to never occur. But we could also have `unreachable!()` or bubble up an error. I don't have a strong opinion.

router/src/validation.rs
danieldk: Simplifying this now to just test one config, I have prepared a next PR that merges the code paths (but needs this one).

danieldk: The test is not parametrized anymore.

router/client/src/lib.rs
Narsil: Either is fine.
Seems to be an artifact of the protobuf definition.

- `unreachable` makes thing hard crash so easier to debug, but less robust
- do nothing, harder to debug should anything go wrong (if even visible) but more robust.

I don't a strong opinion to be honest. Insinctively I'd make it a `unreachable` or equivalent, at the very least in the silent path I would add a warn log.

Narsil (dismissed):
LGTM

PR #1985, opened 2024-05-31
server: use chunked inputs

danieldk:
# What does this PR do?

The router will now send the input as chunks besides as a single string. This change modifies the server to process chunked input rather than strings. This also allows us to remove the image extraction code from the server.

**Draft note:** mostly checking whether all models pass, also needs to be rebased on #1981 once that PR is merged.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/utils/chunks.py
+++ b/server/text_generation_server/utils/chunks.py
@@ -1,0 +1,27 @@
+from typing import Iterable
+
+from loguru import logger
+
+from text_generation_server.pb import generate_pb2
+
+
+def concat_text_chunks(chunks: Iterable[generate_pb2.InputChunk]) -> str:
+    """
+    Concatenate text in text chunks. Non-text chunks are dropped.
+    """
+    text = None
+    for chunk in chunks:
+        chunk_type = chunk.WhichOneof("chunk")
+        if chunk_type == "text":
+            if text is None:
+                text = chunk.text
+            else:
+                raise NotImplementedError("Request contained more than one text chunk")
+        else:
+            # We cannot reject this, e.g. warmup sends an image chunk.
+            logger.debug(f"Encountered non-text chunk type {chunk_type}")
+
+    if text is None:
+        raise NotImplementedError("Request without a text chunk")
+
+    return text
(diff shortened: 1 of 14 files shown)


server/text_generation_server/utils/chunks.py
```diff
@@ -0,0 +1,20 @@
+from typing import Iterable
+
+from loguru import logger
+
+from text_generation_server.pb import generate_pb2
+
+
+def concat_text_chunks(chunks: Iterable[generate_pb2.InputChunk]) -> str:
```
OlivierDehaene, line 8: Is this method only to be future proof or is there a way today to have multiple text chunks?

server/text_generation_server/utils/chunks.py
```diff
@@ -0,0 +1,20 @@
...
+    """
+    Concatenate text in text chunks. Non-text chunks are dropped.
+    """
+    texts = []
+    for chunk in chunks:
+        chunk_type = chunk.WhichOneof("chunk")
+        if chunk_type == "text":
+            texts = chunk.text
```
drbh, line 16: related to above, maybe we append `chunk.text` to `texts` to handle multiple text chunks?

```suggestion
            texts.append(chunk.text)
```

danieldk: Thanks, that should indeed have been an append, fixed!

server/text_generation_server/utils/chunks.py
danieldk: AFAIK we can currently only have multiple text chunks in a VLM models, so this was indeed only to future proof.

OlivierDehaene: Then mb we should take [0] and crash if len > 1 with an unreachable?

danieldk: Oh, we do need to iterate over chunks, because we are sending image chunks unconditionally during warmup, even for text-only models:

https://github.com/huggingface/text-generation-inference/blob/757223b352896a9b2b9df46c95c0afcaa0cdf9d4/router/client/src/v3/client.rs#L140

The current approach seems more robust? What do you think about logging a warning when `len(texts) > 1`?

danieldk: Updated to:

* Fail when there is more than one text chunk.
* Fail when there is no text chunk.
* Log at debug-level when there is a non-text chunk (only log because e.g. warmup sends an image chunk).

OlivierDehaene (approved):

PR #2005, opened 2024-06-04
Split build workflow for multiple plateforms

fxmarty:
As per title.

We may need other EC2 runners, or stop using them altogether, as it may be meaningless to launch multiple times the same AMI?


Narsil:
Can't we keep a single file and making it parallel using `strategy: matrix` instead ?

fxmarty:
We could indeed keep a single file and just remove the
```
    needs:
      - start-runner
      - build-and-push-image # Wait for the main docker image to be built
      - integration-tests # Wait for the main integration-tests
```
and use e.g.

`runs-on: [self-hosted, tgi-builder]` instead of AWS.

PR #2016, opened 2024-06-05
Allow to set manually paged attention num_blocks

fxmarty:
As per title.

Different devices have different global memory size. In order to compare them apple-to-apple we need to be able to use the same number of blocks in the KV cache buffer independently of the device memory size.

### DIFF
--- a/server/text_generation_server/models/flash_causal_lm.py
+++ b/server/text_generation_server/models/flash_causal_lm.py
@@ -808,12 +808,17 @@ class FlashCausalLM(Model):
 
         free_memory = get_free_memory(self.device, MEMORY_FRACTION)
 
-        num_blocks = (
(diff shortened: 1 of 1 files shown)


OlivierDehaene:
Besides the issue you reported that will be fixed as soon as the CI is green, do you see performance differences when the number of blocks are not the same?

fxmarty:
@OlivierDehaene As you mentioned, the only diff should be `reshape_and_cache`? I did not properly benchmark, but from the timings I got, `CacheManager.allocate` was really the issue

fxmarty:
As discussed, we will rather use `--cuda-memory-fraction`, will still leave the debug log for peace of mind

Narsil:
Can we close this with the new allocate logic ?

PR #2028, opened 2024-06-06
Add FP8 KVCache support

mht-sharma:
# What does this PR do?

This PR introduces support for FP8 KV Cache in Text Generation Inference (TGI), significantly enhancing performance and memory efficiency on both Nvidia and AMD GPUs. By quantizing the KV cache to 8-bit floating point (FP8) formats, we can greatly reduce the memory footprint, leading to faster and more scalable text generation.

### Hardware Compatibility:
* Nvidia GPUs: Supports both FP8E4M3 and FP8E5M2 (TODO: Need VLLM update).
* AMD GPUs: Supports FP8E4M3.

Example Usage:
```
text-generation-launcher --model-id <model_id> --kv-cache-dtype fp8/fp8_e5m2
```

KV cache scaling factors should be included in the FP16 checkpoint for E4M3 format to maintain accuracy. Default scaling factor is set to 1.0 if not provided, which may lead to accuracy loss.

### Checkpoint Structure for KV Scales:
The FP8 KV cache scaling factors are specified through the `.kv_scale` parameter in the attention module

```
model.layers.0.self_attn.kv_scale                < F32
model.layers.1.self_attn.kv_scale                < F32
```

This follows a structure proposed in vllm - https://docs.vllm.ai/en/stable/quantization/fp8.html#fp8-checkpoint-structure-explanation

When providing `.kv_scale` in model, the config should specify proper `kv_cache_torch_dtype` used to generate scales (`float8_e4m3fn` or `float8_e4m3fnuz`).

Currently, users need to extract the KV scales from FP8 checkpoint and add to the FP16 model. A helper script is provided in the PR for the same.

Sample Models with KV scales:  [Models with FP8 KV Cache](https://huggingface.co/collections/mohitsha/fp8-kv-cache-667980b582f924034c635da7)

Todos:
* [x] Documentation
* [ ] Tests
* [ ] Update VLLM for CUDA to support E5M2. @Narsil could you help with this!
* [ ] Only supports LLAMA, will update same for other models in this or other PRs


server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -435,3 +462,33 @@ def forward(
...
+            self.config.num_hidden_layers,
+            self.config.__class__.model_type,
+        ):
+            layer_self_attn = self.model.layers[layer_idx].self_attn
+
+            if SYSTEM == "rocm":
+                # The scaling factor convention we are assuming is
+                # quantized_value * scaling_factor ~= true_value
+                # which is consistent with the practice of setting
+                # scaling_factor = tensor_amax / FPtype_max
```
Narsil, line 483: And why does that imply that rocm needs a factor of 2 ?


If the reason is that much global, shouldn't it be handled directly on load?

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -435,3 +462,33 @@ def forward(
...
+
+            if hasattr(layer_self_attn, "kv_scale"):
+                layer_self_attn.kv_scale = scaling_factor
+                logger.info(
+                    f"Loaded KV cache scaling factor for layer {layer_idx}: {scaling_factor}"
+                )
+            else:
+                raise RuntimeError(
+                    "Self attention has no KV cache scaling " "factor attribute!"
+                )
```
Narsil, line 494: Not needed if done at load time.

server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -692,6 +692,8 @@ def __init__(
         head_size: int,
         dtype: torch.dtype,
         device: torch.device,
+        kv_cache_dtype: str = "auto",
+        quantization_param_path: Optional[str] = None,
```
Narsil, line 696: Fused?

server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -713,6 +715,37 @@ def __init__(
...
+            self.kv_cache_dtype = torch.uint8
+        else:
+            self.kv_cache_dtype = self.dtype
+
+        if kv_cache_dtype == "fp8" and SYSTEM == "rocm":
+            logger.info(f"Using KV cache data type: {kv_cache_dtype}")
+            # Currently scaled KV cache is only enabled on ROCm
+            if quantization_param_path is not None:
+                if callable(getattr(self.model, "load_kv_cache_scales", None)):
+                    self.model.load_kv_cache_scales(quantization_param_path)
```
Narsil, line 728: If done correctly, you don't need any of that because you handle this at load time (and every model will need to be updated to support the new values)

server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -713,6 +715,37 @@ def __init__(
...
+                        f"{self.model.__class__} does not "
+                        "support loading scaling factors."
+                    )
+            else:
+                logger.info(
+                    "Using FP8 KV cache but no scaling factors "
+                    "provided. Defaulting to scaling factors of 1.0. "
+                    "This may lead to less accurate results!"
+                )
+        elif quantization_param_path is not None:
```
Narsil, line 742: This should be handled all the way back in the launcher (again ideally directly by clap, if not possible manually in the rust part). 

And it should be hard error, not a soft one (paramters sent by the user don't make sense, we never silently ignore.)

server/text_generation_server/models/flash_llama.py
```diff
@@ -29,6 +29,8 @@ def __init__(
         quantize: Optional[str] = None,
         speculator: Optional[str] = None,
         dtype: Optional[torch.dtype] = None,
+        kv_cache_dtype: Optional[str] = "auto",
+        quantization_param_path: Optional[str] = None,
```
Narsil, line 33: Fused?

server/text_generation_server/utils/weights_utils.py
```diff
@@ -0,0 +1,48 @@
...
+    except Exception as e:
+        logger.error(f"An error occurred while reading '{filename}': {e}")
+    # This section is reached if and only if any of the excepts are hit
+    # Return an empty iterable (list) => no KV cache scales are loaded
+    # which ultimately defaults to 1.0 scales
+    logger.warning(
+        "Defaulting to KV cache scaling factors = 1.0 "
+        f"for all layers in TP rank {tp_rank} "
+        "as an error occurred during loading."
+    )
```
Narsil, line 47: Everything here should be hard error (I think the standard ones would do fine).

If a user sends invalid information, we shouldn't silently ignore. They should fix it.

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -126,6 +127,16 @@ def __init__(
...
 
+        # This will be overwritten by model initialization if we are using it.
+        # N.B. currently we only support per tensor scalar scaling factors
+        # & only applicable to ROCm (AMD GPU).
+        # The scaling factor convention we are assuming is
+        # quantized_value * scaling_factor ~= true_value
+        # which is consistent with the practice of setting
+        # scaling_factor = tensor_amax / FPtype_max
+        self.kv_scale = 1.0
+        self.kv_cache_dtype = "auto"
```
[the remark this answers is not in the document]
Narsil: We do the same for `quantize` and `speculator` for instance.

Codewise it touches a lot of places, but these are about to be factored out.

Narsil (commented):
Thanks for this PR.

I think a lot has to be changed (to simplify it).

Also I don't see any core logic to actually handle the fp8, are the kernels ready?
Is it possible to test/add tests ?

launcher/src/main.rs
```diff
@@ -184,6 +184,23 @@ struct Args {
...
+
+    /// Path to the JSON file containing the KV cache
+    /// scaling factors. This should generally be supplied, when
+    /// KV cache dtype is FP8. Otherwise, KV cache scaling factors
+    /// default to 1.0, which may cause accuracy issues.
+    /// FP8_E5M2 (without scaling) is only supported on cuda version
+    /// greater than 11.8. On ROCm (AMD GPU), FP8_E4M3 is instead
+    /// supported for common inference criteria.
+    #[clap(long, env)]
+    quantization_param_path: Option<String>,
```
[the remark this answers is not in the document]
Narsil: (I'm trying to avoid adding too many flags, TGI already has too many, and since we don't break, we never remove stuff that was added, that's why if we can read the information from some consistant config in the repo it keeps the interface for the user simpler)

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -152,7 +163,13 @@ def forward(
         self.rotary_emb(query, torch.select(kv, dim=1, index=0), cos, sin)
 
         paged_attention.reshape_and_cache(
-            kv[:, 0], kv[:, 1], kv_cache[0], kv_cache[1], slots
+            kv[:, 0],
+            kv[:, 1],
+            kv_cache[0],
+            kv_cache[1],
+            slots,
+            self.kv_cache_dtype,
```
Narsil, line 171: This one isn't necessary here, it should already be inferrable from `kv_cache[0]`.

Narsil:
Happy to help with the rebase btw.

server/text_generation_server/layers/schema.py
```diff
@@ -0,0 +1,90 @@
...
+possible on ROCm), the model can be optionally augmented with KV cache
+scaling factors.
+"""
+
+from typing import Dict, Optional
+
+from pydantic import BaseModel, ConfigDict, ValidationInfo, model_validator
+
+
+class KVCacheQuantSchema(BaseModel):
```
fxmarty, line 19: I would use builtin dataclasses and post_init here but it is a matter of taste

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
fxmarty: @Narsil I believe it is related to https://onnx.ai/onnx/technical/float8.html and the diff between e4m3fn and e4m3 (different exponent bias). Is that so @mht-sharma?

But shouldn't it be based on this param https://github.com/vllm-project/vllm/blob/319ad7f1d386699e94f629341c9988a926821f24/tests/fp8_kv/llama2-7b-fp8-kv/kv_cache_scales.json#L4 ? Also, we have to kind of decide whether we want to follow the scheme in vllm to store quantization params in a json and keep compatibility for e.g. models on the Hub, or not.

fxmarty (commented):
nothing shocking to me for benchmarking!

mht-sharma:
Thanks for the review @Narsil @fxmarty  I will rebase and address the comments. 

Regarding the format for loading the FP8 scales:

VLLM offers two methods:

* quantization-param-path: This uses a JSON file (kv_cache_scales.json) containing per-tensor scaling factors for each layer. Example can be found [here](https://github.com/vllm-project/vllm/blob/main/tests/fp8_kv/llama2-70b-fp8-kv/kv_cache_scales.json). This file is generated using the Nvidia AMMO quantizer available [here](https://github.com/vllm-project/vllm/tree/main/examples/fp8/quantizer).

* Direct loading from checkpoints: This method has been introduced in one of the recent PRs and is located [here](https://github.com/comaniac/vllm/blob/main/vllm/model_executor/layers/quantization/fp8.py#L293).

VLLM intends to deprecate the `quantization-param-path` method soon, favoring the use of checkpoints for loading scales. Therefore, I would update our approach to also load scales using checkpoints.

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
mht-sharma: Comment from AMD in vllm: 

> We do *2 only for HIP, to deal with the difference in numeric from our chip. after *2 overall effect is identical as without it on NV.

fxmarty: Yes, this is the difference between e4m3fn and e4m3 formats. If using a json like the one liked, there should be a check on the `dtype`.

mht-sharma:
> Thanks for the review @Narsil @fxmarty I will rebase and address the comments.
> 
> Regarding the format for loading the FP8 scales:
> 
> VLLM offers two methods:
> 
> * quantization-param-path: This uses a JSON file (kv_cache_scales.json) containing per-tensor scaling factors for each layer. Example can be found [here](https://github.com/vllm-project/vllm/blob/main/tests/fp8_kv/llama2-70b-fp8-kv/kv_cache_scales.json). This file is generated using the Nvidia AMMO quantizer available [here](https://github.com/vllm-project/vllm/tree/main/examples/fp8/quantizer).
> * Direct loading from checkpoints: This method has been introduced in one of the recent PRs and is located [here](https://github.com/comaniac/vllm/blob/main/vllm/model_executor/layers/quantization/fp8.py#L293).
> 
> VLLM intends to deprecate the `quantization-param-path` method soon, favoring the use of checkpoints for loading scales. Therefore, I would update our approach to also load scales using checkpoints.

Removed the `quantization-param-path` altogether: This method is already deprecated in VLLM, based on discussions here: https://github.com/vllm-project/vllm/issues/4532

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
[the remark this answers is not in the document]
mht-sharma: Done, the params should be created at load time

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -387,6 +406,10 @@ class FlashLlamaForCausalLM(torch.nn.Module):
     def __init__(self, prefix, config, weights):
         super().__init__()
 
+        process_group = weights.process_group
+        self.tp_rank = process_group.rank()
+        self.tp_world_size = process_group.size()
```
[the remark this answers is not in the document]
mht-sharma: Done

server/text_generation_server/models/__init__.py
```diff
@@ -563,6 +568,8 @@ def get_model(
                 quantize=quantize,
                 speculator=speculator,
                 dtype=dtype,
+                kv_cache_dtype=kv_cache_dtype,
```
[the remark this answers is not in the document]
mht-sharma: Done

mht-sharma:
> Also I don't see any core logic to actually handle the fp8, are the kernels ready? Is it possible to test/add tests ?

The core logic for handling FP8 is managed by the Paged Attention kernel in VLLM, with the necessary kernel tests. If you have any specific tests in mind, we can discuss them. VLLM includes tests that compare the output with the expected FP8 output, as seen https://github.com/comaniac/vllm/blob/main/tests/models/test_fp8.py. We can add a similar test if required.

examples/fp8_kvcache/README.md
```diff
@@ -0,0 +1,40 @@
...
+  --save-path SAVE_PATH
+                        Path to save the FP16 model with the kv scales
+```
+
+## Example usage
+To extract KV cache scaling factors from a quantized FP8 model and save them to an unquantized FP16 model, use the following command:
+
+```
+python extract_fp8_kv_scales.py --quantized-model neuralmagic/Meta-Llama-3-8B-Instruct-FP8-KV --model meta-llama/Meta-Llama-3-8B-Instruct --save-path  Meta-Llama-3-8B-Instruct
+```
```
fxmarty, line 40: If we target vllm/tgi intercompatibility, why couldn't we load directly e.g. `neuralmagic/Meta-Llama-3-8B-Instruct-FP8-KV` in TGI? Is it not loadable in vllm?

Also, by `quantized FP8 model`, do you mean a model whose weights are quantized to fp8? How does it relate to FP8 KV cache? To me to obtain the KV cache scales you would simply need to have calibration data passing through the network & collecting stats on the KV cache.

It feels like the the KV cache scales from a quantized model (like `neuralmagic/Meta-Llama-3-8B-Instruct-FP8-KV`), whose KV cache scales may have been obtained after the weights quantization (?), may have an unnecessary bias due to being inferred from calibration on the quantized model, not the unquantized one.

examples/fp8_kvcache/README.md
```diff
@@ -0,0 +1,40 @@
+# FP8 (fp8_e4m3) KV Cache Scaling Factor Extraction Utility
+
+This utility is designed to extract KV cache scaling factors from a quantized `FP8(fp8_e4m3)` Hugging Face (HF) model. The extracted scaling factors are then saved to the corresponding unquantized HF model, which can be used with Text Generation Inference (TGI).
```
fxmarty, line 3: Do you have an example of such model?

docs/source/basic_tutorials/fp8_kv_cache.md
```diff
@@ -0,0 +1,56 @@
+# Accelerating Inference with FP8 KV Cache
```
fxmarty, line 1: I think it would be worth to have some evaluation metrics for an example model

fxmarty (commented):
Great work, that's cool! I left some general comments, @Narsil may have more about the design

docs/source/basic_tutorials/fp8_kv_cache.md
```diff
@@ -0,0 +1,56 @@
...
+```
+
+### Checkpoint structure for KV scales
+The FP8 kv cache scaling factors required in the FP16 checkpoints are specified through the .kv_scale parameter present on the `Attention` module, such as:
+
+```
+model.layers.0.self_attn.kv_scale                < F32
+model.layers.1.self_attn.kv_scale                < F32
+...
+```
```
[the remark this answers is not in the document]
mht-sharma: I guess the current tools Nvidia AMMO and AutoFP8 uses `E4M3FN`. Currently there is no flag to determine the format (unless ofcourse the weight is quantized).  But I could add a check and add a parameter for this in the checkpoint.

examples/fp8_kvcache/README.md
mht-sharma: The model `neuralmagic/Meta-Llama-3-8B-Instruct-FP8-KV` has both the weights quantized along with KV scales. However, I don't think we can load the FP8 weights in TGI yet.

You are right; it may have an additional bias due to calibration after weight quantization. The bias might have been low, so I couldn't find a noticeable difference during inference.

I have tested two other models generated using scales from Nvidia AMMO (VLLM uses this also), which may not have this bias. The quantizer can accept the quantized format as full precision and provide the model with KV scales. We can extract scales from that.

Here is the link to the models: [Llama-2-70b-chat-hf-FP8-KV](https://huggingface.co/mohitsha/Llama-2-70b-chat-hf-FP8-KV/tree/main)

I will add these models and provide an example for this.

examples/fp8_kvcache/README.md
mht-sharma: Updated and added a model to readme.

examples/fp8_kvcache/README.md
mht-sharma: Updated the example with Nvidia AMMO

server/text_generation_server/utils/weights.py
```diff
@@ -762,6 +767,40 @@ def _set_gptq_params(self, model_id, revision):
...
+                    "Only support per-tensor scaling factor for `fp8 (fp8_e4m3)` KV cache"
+                )
+
+            # ROCm uses FP8 format with fp8_e4m3fn, whereas Nvidia GPUs use fp8_e4m3.
+            # The multiplication by 2 compensates for the different numeric representation
+            # between ROCm and Nvidia GPUs, ensuring consistent effective scaling across platforms.
+            #  After this adjustment, the overall effect is equivalent to the scaling applied without
+            # it on Nvidia GPUs.
+            if SYSTEM == "rocm":
+                kv_scale *= 2.0
```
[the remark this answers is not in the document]
mht-sharma: Updated the code to read the corresponding dtype as `kv_cache_torch_dtype` from config.

Added the format in the README.md and utility script to add `kv_cache_torch_dtype` when quantising model

docs/source/basic_tutorials/fp8_kv_cache.md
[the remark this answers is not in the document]
mht-sharma: Updated reading this from config. See: https://github.com/huggingface/text-generation-inference/pull/2028/commits/a7909e6f9409b94cc0f530edce318729bcaffab4

docs/source/basic_tutorials/fp8_kv_cache.md
```diff
@@ -0,0 +1,56 @@
+# Accelerating Inference with FP8 KV Cache
+
+Text Generation Inference (TGI) now supports FP8 KV Cache, enhancing inference speed on both Nvidia and AMD GPUs. This feature significantly boosts performance and memory efficiency, enabling faster and more scalable text generation. By quantizing the KV cache to 8-bit floating point (FP8) formats, we can greatly reduce the memory footprint. This reduction allows for improved throughput in text generation tasks
```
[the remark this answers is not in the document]
mht-sharma: We have a couple of options for numbers and charts
* Max Batch Total Tokens: The logs display the max batch total tokens, which increase when using the FP8 KV cache. We could create a chart showing the max batch total tokens in both cases (with and without the FP8 KV cache).

* Throughput: Currently, I have created a custom script using `AsyncClient` to send 500 requests simultaneously with `asyncio.gather`. This provides a rough estimate of throughput. @Narsil , do you have any suggestions on calculating throughput more precisely?

server/text_generation_server/models/__init__.py
```diff
@@ -246,6 +246,12 @@ class ModelType(enum.Enum):
     }
 
 
+FP8_KVCACHE_SUPPORTED_MODELS = {
+    "llama",
+    "baichun",
```
yao-matrix, line 251: is this a typo of "baichuan"?

server/text_generation_server/utils/weights.py
```diff
@@ -762,6 +767,50 @@ def _set_gptq_params(self, model_id, revision):
...
+                    "Only support per-tensor scaling factor for `fp8 (fp8_e4m3)` KV cache"
+                )
+
+            if kv_cache_torch_dtype not in {"float8_e4m3fn", "float8_e4m3fnuz"}:
+                raise RuntimeError(
+                    f"Found `kv_scale` in the checkpoint, the config must specify the `kv_cache_torch_dtype` "
+                    f"used for generating kv scales. Expected 'float8_e4m3fn' or 'float8_e4m3fnuz', but got '{kv_cache_torch_dtype}'."
+                )
+
+            # ROCm uses FP8 format with fp8_e4m3fn, whereas Nvidia GPUs use fp8_e4m3.
```
yao-matrix, line 802: Is it should be "ROCm uses FP8 format with fp8_e4m3fnuz, whereas NVIDIA GPU uses fp8_e4m3fn"?

mht-sharma: Yes that's correct

server/text_generation_server/models/__init__.py
mht-sharma: Yes. Thanks for pointing it out

Narsil:
Closing this as we added support for FP8 kv cache support in https://github.com/huggingface/text-generation-inference/pull/2603.

More support is coming (for pre-scaled kv-cache fp8)

PR #2033, opened 2024-06-06
ROCm and sliding windows fixes

fxmarty:
Fix models requiring window attention - no need to raise an error at load time in case max_input_tokens < sliding_window.

Also, updates vllm fork commit to use the fixes from https://github.com/ROCm/vllm/pull/28 & fix a few rocm issues

### DIFF
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -476,6 +476,7 @@ fn shard_manager(
     rope_factor: Option<f32>,
     max_total_tokens: usize,
     max_batch_size: Option<usize>,
+    max_batch_prefill_tokens: u32,
     otlp_endpoint: Option<String>,
     log_level: LevelFilter,
     status_sender: mpsc::Sender<ShardStatus>,
@@ -547,6 +548,10 @@ fn shard_manager(
         shard_args.push("--otlp-endpoint".to_string());
         shard_args.push(otlp_endpoint);
     }
+
+    // In case we use sliding window, we may ignore the sliding in flash for some backends depending on the parameter.
+    shard_args.push("--max-batch-prefill-tokens".to_string());
+    shard_args.push(max_batch_prefill_tokens.to_string());
 
     // Copy current process env
     let mut envs: Vec<(OsString, OsString)> = env::vars_os().collect();
@@ -1004,6 +1009,7 @@ fn spawn_shards(
     args: &Args,
     cuda_graphs: Vec<usize>,
     max_total_tokens: usize,
+    max_batch_prefill_tokens: u32,
     max_log_level: LevelFilter,
     shutdown: Arc<AtomicBool>,
     shutdown_receiver: &mpsc::Receiver<()>,
@@ -1061,6 +1067,7 @@ fn spawn_shards(
                 rope_factor,
                 max_total_tokens,
                 max_batch_size,
+                max_batch_prefill_tokens,
                 otlp_endpoint,
                 max_log_level,
                 status_sender,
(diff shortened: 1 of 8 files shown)


Narsil:
> Fix models requiring window attention - no need to raise an error at load time. Only if context > window length at runtime.

context --max-total-tokens, right ? (We need to crash at load time, crashing randomly at runtime is a terrible UX).

server/text_generation_server/models/__init__.py
```diff
@@ -412,9 +412,8 @@ def get_model(
     sliding_window = config_dict.get("sliding_window", -1)
     if sliding_window != -1 and not SUPPORTS_WINDOWING:
         logger.warning(
-            f"Flash attention is available, but doesn't support windowing which is required by model {model_id}"
+            f"Flash attention is available, but doesn't support windowing which is required by model {model_id} for long contexts."
         )
-        FLASH_ATTENTION = False
```
Narsil, line 417: This modifies the behavior of all targets, not just rocm, change things in Rocm if you wnat to change behavior there.

IMHO having load time crash is much better than random runitime one (and checking against max-total-tokens would definitely make sense).


We could change here because WINDOWING is just not needed if `max-total-tokens` < `sliding_window`. (and therefore that logic would be valid for all targets)

fxmarty: that works

fxmarty:
@Narsil I assume paged attention always works with sliding window

server/text_generation_server/models/__init__.py
```diff
@@ -702,7 +707,21 @@ def get_model(
...
     if model_type == MISTRAL:
         sliding_window = config_dict.get("sliding_window", -1)
-        if FLASH_ATTENTION:
+        if (
+            (sliding_window is not None and sliding_window != -1)
+            and not SUPPORTS_WINDOWING
+            and max_batch_prefill_tokens > sliding_window
+        ):
+            raise ValueError(
+                SLIDING_WINDOW_MESSAGE.format(
```
Narsil, line 716:
```suggestion
                f"The backend {SYSTEM} does not support sliding window attention. TGI webserver was started max_batch_prefill_tokens={max_batch_prefill_tokens} larger than sliding_window={sliding_window}. To use this model with the {SYSTEM} backend, please launch TGI with the argument `--max-batch-prefill-tokens` smaller than {sliding_window}."
```

Narsil: I think we can factor this out to have a single location for the error message.


We get sliding window *before* the massive switch.
Then we see if we need windowing.

```python
sliding_window = config_dict.get("sliding_window", -1)
needs_windowing = (sliding_window is not None and sliding_window != -1) and max_input_tokens > sliding_window
if needs_windowing and not SUPPORTS_WINDOWING:
    FLASH_ATTENTION = False
    # or raise 
   'The backend {SYSTEM} ...
```

Wouldn't that simplify massively things? 

Also I wasn't aware those other models also used windowing, I thought only mistral did (not even mixtral)

launcher/src/main.rs
```diff
@@ -548,6 +549,10 @@ fn shard_manager(
         shard_args.push(otlp_endpoint);
     }
 
+    // In case we use sliding window, we may ignore the sliding in flash for some backends depending on the parameter.
+    shard_args.push("--max-batch-prefill-tokens".to_string());
+    shard_args.push(max_batch_prefill_tokens.to_string());
```
Narsil, line 554: Shouldn't it be `--max-input-tokens` actually ? Windowing doesn't apply if seqlen > window but it's multiple queries with each their input lengths < windowing, no?

Narsil (approved):
LGTM

PR #2044, opened 2024-06-10
Adding architecture document

tengomucho:
# What does this PR do?

Added a document (visible [here](https://github.com/huggingface/text-generation-inference/blob/architecture-doc/docs/source/architecture.md)) that aims at explaining the relationship between the TGI components. This can be used to better understand TGI architecture and provide a documentation for someone who might want to implement or improve a solution deployed on a hardware different from the ones that are already supported. It can be also used as base of a reflection on choices for future design changes.

## Who can review?

@OlivierDehaene @Narsil

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,9 +4,9 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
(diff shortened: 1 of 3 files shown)


drbh (approved):
Thank you @tengomucho! LGTM

tengomucho:
@drbh  Thanks for removing the cargo lock file, I added it by mistake.
I saw the rendering is good in github.com, but not in the docs 😐 https://moon-ci-docs.huggingface.co/docs/text-generation-inference/pr_2044/en/architecture
Do you know if something can be done to fix that?

drbh:
no worries at all, thanks for the additional docs. I can see them in production here: https://huggingface.co/docs/text-generation-inference/architecture, can you see them there as well?

I think the link shared above is from `moon-ci` and may just be temporary and may have been deleted with the branch

PR #2062, opened 2024-06-12
fix: set sharded true if WORLD_SIZE is set

drbh:
This is a tiny PR that simply set assumes `sharded` is true if a `WORLD_SIZE` env is set. 

Adding because I wasted time debugging when I just forgot to set `--sharded` along with the `WORLD_SIZE` and `RANK` env vars. This PR allows `--sharded` to be inferred and is for dev convenience

note** we already do this in the launcher https://github.com/huggingface/text-generation-inference/blob/11ea9ce002e796cc59714950b557b4021cbebc58/launcher/src/main.rs#L518-L521 but this avoids the issue if you are running the server/router separately

### DIFF
--- a/server/text_generation_server/cli.py
+++ b/server/text_generation_server/cli.py
@@ -44,6 +44,9 @@ def serve(
     otlp_endpoint: Optional[str] = None,
     max_input_tokens: Optional[int] = None,
 ):
+    # derive sharded from environment variables if not provided
+    sharded = sharded or os.getenv("WORLD_SIZE", None) is not None
+
     if sharded:
         assert (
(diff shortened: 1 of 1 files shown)


Narsil:
Not a huge fan tbh.

All of this is supposed to be handled by the launcher which shouldn't make any mistake.
I understand the inconvenience, but this seems like opening up a new class of bugs for relatively little benefit. (Because WORLD_SIZE could be defined by somehow else than the launcher).

Let's see how the others feel about this.

drbh:
Closing since this change is unneeded and it does not seem like a change that will benefit others. Will revisit if interest arises

PR #2068, opened 2024-06-14
Update the link for qwen2

xianbaoqian:
# What does this PR do?

Fix link for qwen


danieldk:
Thanks! I tacked on a commit to also fix the URL in the model table in the backend. This should also fix the CI failure.

PR #2071, opened 2024-06-16
Fix local installation after Rust 1.79 and transformers 4.41.2

rYoussefAli:
I was trying to install text-generation-inference locally without Docker, and I encountered several problems while doing so. I collected all those problems and made the fixes here so that people do not encounter such tedious problems again.

Fixes # (issue)
1. The rust-toolchain.toml file was overriding Rust version to 1.78.0, making the installation fail as the inline const feature in rust requires 1.79.0.
2. Upgraded the transformers version.
3. Fixed the error: "str" has no attribute "logits" in rw.py when the model starts 

## Before submitting
- [ X] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [ X] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?

## Who can review?

@OlivierDehaene 
@Narsil

### DIFF
--- a/rust-toolchain.toml
+++ b/rust-toolchain.toml
@@ -1,5 +1,5 @@
 [toolchain]
-# Released on: 02 May, 2024
-# https://releases.rs/docs/1.78.0/
-channel = "1.78.0"
+# Released on: 13 June, 2024
+# https://releases.rs/docs/1.79.0/
+channel = "1.79.0"
 components = ["rustfmt", "clippy"]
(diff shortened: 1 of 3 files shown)


Narsil:
The rust-toolchain is already fixed on main (and you don't seem to fix it it seems).
Rw.py should be in it's own PR (and with a reproducer, no need for a test since we don't really maintain that hard non flash models, no custom kernels models anymore).

As for the transformers version you are simply modifying the `benchmark` calls which is an optional dependency, and the lockfile only points to an old version because some package seems to depend on an old version.
Regardless we never manually touch the lock file, only `pyproject.toml` so the solution would need to be in it.

PR #2095, opened 2024-06-20
corrected Pydantic warning.

yukiman76:
On using python client with pydantic 2, it raises a warning on import time: 
this corners of the execution logs

Fixes # (1761)

- [ ] I ran it locally with the patch 

@OlivierDehaene 
@Narsil

### DIFF
--- a/clients/python/text_generation/types.py
+++ b/clients/python/text_generation/types.py
@@ -1,5 +1,5 @@
 from enum import Enum
-from pydantic import BaseModel, field_validator
+from pydantic import BaseModel, field_validator, ConfigDict
 from typing import Optional, List, Union, Any
 
 from text_generation.errors import ValidationError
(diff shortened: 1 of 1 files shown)


clients/python/text_generation/types.py
```diff
@@ -452,5 +452,6 @@ class StreamResponse(BaseModel):
 
 # Inference API currently deployed model
 class DeployedModel(BaseModel):
+    model_config  = ConfigDict(protected_namespaces=())
```
danieldk, line 455:
```suggestion
    # Disable warning for use of `model_` prefix in `model_id`. Be mindful about adding members
    # with model_ prefixes, since this disables guardrails for colliding fields:
    # https://github.com/pydantic/pydantic/issues/9177
    model_config  = ConfigDict(protected_namespaces=())
```

danieldk (commented):
Thanks! I think it's worth adding a comment here to prevent accidents. What do you think?

PR #2102, opened 2024-06-21
Add support for Marlin 2:4 sparsity

### LINKED ISSUE #2098: Sparse Marlin
### Feature request

Sparse-Marlin, an extension of the [Mixed Auto-Regressive Linear (Marlin)](https://github.com/IST-DASLab/marlin/tree/master) dense kernel for 4-bit quantized weights, now with support for 2:4 sparsity.

Specifically, Sparse-Marlin is an extremely optimized GPU kernel performing a matrix multiply and accumulate operation, C=AxB, where A and C are FP16 matrices, and B is a 50% sparse matrix following the 2:4 format, whose values are quantized and stored in INT4 precision. Integrating a 2:4 sparsity scheme on top of the existing 4-bit quantized weight representation offers the opportunity to further enhance the FLOPS/Byte ratio, and push the baseline performance further. Sparse-Marlin boosts the close-to-ideal performance of dense Marlin, and provides near-optimal speedups of around 5.3x, while being more robust to larger batch sizes.

### Motivation

image
image

### Your contribution

https://github.com/IST-DASLab/Sparse-Marlin/blob/main/marlin/marlin_cuda_kernel_nm.cu

Pinging @danieldk since he did the initial marlin PR

I could give it a try if it is wanted, but would also be happy with it if you know

danieldk:
# What does this PR do?

This change adds support for 2:4 sparsity when using Marlin quantization. The 2:4 kernel is used when:

* The quantizer is `marlin`;
* the quantizer checkpoint format is `marlin_24`.

Fixes #2098.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/marlin/marlin_kernels/__init__.pyi
+++ b/server/marlin/marlin_kernels/__init__.pyi
@@ -19,6 +19,23 @@ def gptq_marlin_gemm(
     """
     ...
 
+def gptq_marlin_24_gemm(
+    a: torch.Tensor,
+    b_q_weight: torch.Tensor,
+    b_meta: torch.Tensor,
+    b_scales: torch.Tensor,
+    workspace: torch.Tensor,
+    num_bits: int,
+    size_m: int,
+    size_n: int,
+    size_k: int,
+) -> torch.Tensor:
+    """
+    Matrix multiplication using Marlin kernels. This is an extension of
+    `marlin_gemm` that supports 2:4 sparsity.
+    """
+    ...
+
 def gptq_marlin_repack(
     b_q_weight: torch.Tensor,
     perm: torch.Tensor,
--- a/server/text_generation_server/layers/marlin.py
+++ b/server/text_generation_server/layers/marlin.py
@@ -1,9 +1,8 @@
 from dataclasses import dataclass
-from typing import Optional, Tuple, List
+from typing import List, Optional, Tuple
 
 import torch
 import torch.nn as nn
-
 from text_generation_server.utils.import_utils import SYSTEM
 
 try:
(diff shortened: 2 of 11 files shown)


server/marlin/marlin_kernels/__init__.pyi
```diff
@@ -19,6 +19,23 @@ def gptq_marlin_gemm(
     """
     ...
 
+def gptq_marlin_24_gemm(
```
Narsil, line 22: Are we running a fork of marlin ?
Shouldn't we use the classic makefile + commit approach instead (or already made releases if possible) ?

server/text_generation_server/layers/marlin.py
```diff
@@ -215,6 +214,116 @@ def forward(self, A: torch.Tensor) -> torch.Tensor:
         return C
 
 
+GPTQ_MARLIN_24_MIN_THREAD_N = 128
+GPTQ_MARLIN_24_MIN_THREAD_K = 128
+GPTQ_MARLIN_24_MAX_PARALLEL = 64
+GPTQ_MARLIN_24_SUPPORTED_NUM_BITS = [4, 8]
+GPTQ_MARLIN_24_SUPPORTED_GROUP_SIZES = [-1, 128]
```
Narsil, line 221: This is coming from marlin directly I guess ?

server/text_generation_server/layers/marlin.py
```diff
@@ -215,6 +214,116 @@ def forward(self, A: torch.Tensor) -> torch.Tensor:
...
+
+        assert (
+            in_features % GPTQ_MARLIN_24_MIN_THREAD_K == 0
+        ), f"Number of output features ({out_features}) not divisable by {GPTQ_MARLIN_24_MIN_THREAD_K} threads"
+        if groupsize != -1 and in_features % groupsize != 0:
+            raise ValueError(
+                f"Number of input features ({in_features}) not divisable by group size ({groupsize})"
+            )
+
+        self.register_buffer("B", weight.B)
```
Narsil, line 290: NIT.
Any reason to prefer this over `self.B = weight.B`? 

Register buffer changes the default save behavior of pytorch (which we don't care about).
And potentially adds some callbacks (we don't care about either).

Narsil (approved):
LGTM

server/marlin/marlin_kernels/__init__.pyi
danieldk: These come from VLLM. We could also directly import them from VLLM if we prefer.

server/text_generation_server/layers/marlin.py
danieldk: Yes, they come from upstream.

server/text_generation_server/layers/marlin.py
danieldk: Torch module habits, fixed! Also for the other Marlin layers.

PR #2104, opened 2024-06-21
feat: sort cuda graphs in descending order

drbh:
sorting cuda graphs in descending order uses slightly less memory during initialization 

using the following command on 1 A10G 
```bash
text-generation-launcher \
--model-id meta-llama/Llama-2-7b-chat-hf \
--num-shard 1 \
--cuda-graphs 2,4,8,16,32,64,128,256,512,1024,2048
```

without cuda graphs sorted `22140/23028 MiB` are used and with them sorted `21938/23028 MiB` are used. 

This is a small percent of total memory (`<1%`) however it's 202 MiB saved. Additionally looking at the memory usage overtime, loading in descending order has smaller spike (never goes above its final value, where in ascending it peaks higher than 22222 MiB)

 Below are the two memory recording with [https://github.com/drbh/nvline](https://github.com/drbh/nvline), left is ascending, and right is descending.
 
desc-graphs-nvline

### DIFF
--- a/server/text_generation_server/models/globals.py
+++ b/server/text_generation_server/models/globals.py
@@ -15,6 +15,13 @@ if cuda_graphs is not None:
 else:
     cuda_graphs = None
 
+
+# sorting the cuda graphs in descending order helps reduce the
+# memory impact and results in less memory usage
+if cuda_graphs is not None:
+    cuda_graphs.sort(reverse=True)
+
+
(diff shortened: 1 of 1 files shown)


danieldk (approved):

drbh:
@danieldk thanks for suggesting this and reviewing!

PR #2119, opened 2024-06-25
fix: simplify kserve endpoint and fix imports

drbh:
This PR makes some small changes to the kserve endpoints to fix an import issue and simplify the code.


drbh:
merging as the changes are orthogonal to the core flow and fix issues behind a feature flag

PR #2127, opened 2024-06-26
fix: prefer serde structs over custom functions

drbh:
**updated

This PR prefers an enum that wraps `ChatCompletionChunk` and `ChatCompletion` and uses the serde `tag` macro to set the correct object value.

unfortunately serde does not provide a `rename_all = "dot.case"` https://github.com/serde-rs/serde/pull/2269 so each value has an explicit rename

the following enum is used
```rust
#[derive(Clone, Serialize, ToSchema)]
#[serde(tag = "object")]
enum CompletionType {
    #[serde(rename = "chat.completion.chunk")]
    ChatCompletionChunk(ChatCompletionChunk),
    #[serde(rename = "chat.completion")]
    ChatCompletion(ChatCompletion),
}
```

**update again

This PR has moved to leveraging serde where possible to remove custom deserialization functions and overall make the types more readable and idiomatic


router/src/lib.rs
```diff
@@ -1021,18 +997,17 @@ impl From<Message> for TextMessage {
...
-                })
-                .collect::<Vec<_>>()
-                .join(""),
+            content: match value.content {
+                MessageContent::SingleText(text) => text,
+                MessageContent::MultipleChunks(chunks) => chunks
+                    .into_iter()
+                    .map(|chunk| match chunk {
+                        MessageChunk::Text { text } => text,
+                        MessageChunk::ImageUrl { image_url } => format!("![]({})", image_url.url),
```
Narsil, line 1006: Didn't we remove that ? I see it's equivalent with what was before so it's ok, but I though we removed those.

Narsil (approved):
LGTM ! Thansk for this PR looks much better imho.


There are some caveats wth `#[serde(untagged)]`  we should be super aware of, but at least now it's explicit rather than implicit.

The biggest issue with it is the lack of strictness in the deserialization.

Narsil (approved):

PR #2128, opened 2024-06-26
feat: use model name as adapter id in chat endpoints

drbh:
This PR allows users to specify lora adapter ids as the `model` value in the `/v1/chat/completions` and `/v1/completions` endpoints. 

This feature aligns with other lora implementations and was mentioned here during the initial lora PR https://github.com/huggingface/text-generation-inference/pull/2010#issuecomment-2149032186

```bash
curl localhost:3000/v1/chat/completions -s \
    -H 'Content-Type: application/json'  \
    -X POST \
    -d '{
  "model": "predibase/customer_support",
  "messages": [
    {
      "role": "system",
      "content": "You are a helpful assistant."
    },
    {
      "role": "user",
      "content": "What is deep learning?"
    }
  ],
  "stream": false,
  "max_tokens": 20, "seed": 42
}' | jq .
```

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -606,6 +606,7 @@ async fn completions(
     metrics::increment_counter!("tgi_request_count");
 
     let CompletionRequest {
+        model,
         max_tokens,
         seed,
         stop,
@@ -673,7 +674,7 @@ async fn completions(
                 seed,
                 top_n_tokens: None,
                 grammar: None,
-                ..Default::default()
+                adapter_id: model.as_ref().filter(|m| *m != "tgi").map(String::from),
             },
         })
         .collect();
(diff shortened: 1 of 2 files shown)


router/src/server.rs
```diff
@@ -673,7 +674,7 @@ async fn completions(
                 seed,
                 top_n_tokens: None,
                 grammar: None,
-                ..Default::default()
+                adapter_id: model.as_ref().filter(|m| *m != "tgi").map(String::from),
```
Narsil, line 677: Can we remove this special value ? It should exist in the first place.

if openai *requires* a model id, then we should actually match it against the actual deployed model id, no ?

Narsil (approved):
Merging.

`"tgi"` is used throughout our docs. Just for backward compat we should support it somehow.

Definitely not great though.

PR #2137, opened 2024-06-27
fix: refactor post_processor logic and add test

drbh:
This PR updates the logic for adding a `post_processor` to a tokenizer and should resolve the issues in CI https://github.com/huggingface/text-generation-inference/actions/runs/9698059633/job/26765404563

This PR aims to match the same functionality in `transformers/models/llama/tokenization_llama_fast.py` and fixes an issue where the single and pairs were not correctly constructed and applied. This resulted in a missing `bos_token` and subsequently an issue with accessing slots in `batch.slots[batch.slot_indices]`. Also a test is added for that specific case

### DIFF
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -304,38 +304,23 @@ async fn main() -> Result<(), RouterError> {
         tracing::warn!("Could not find tokenizer config locally and no API specified");
         HubTokenizerConfig::default()
     });
-    let tokenizer: Option<Tokenizer> =
-        tokenizer_filename.and_then(|filename| {
-            let mut tokenizer = Tokenizer::from_file(filename).ok();
-        if let Some(tokenizer) = &mut tokenizer{
-                if let Some(class) = &tokenizer_config.tokenizer_class{
-                    if class == "LlamaTokenizer" || class == "LlamaTokenizerFast"  {
-                        tracing::info!("Overriding LllamaTokenizer with TemplateProcessing to follow python override defined in https://github.com/huggingface/transformers/blob/4aa17d00690b7f82c95bb2949ea57e22c35b4336/src/transformers/models/llama/tokenization_llama_fast.py#L203-L205");
-                        let mut single = vec![];
-                        let mut special_tokens = vec![];
-                        if let Some(true) = &tokenizer_config.add_bos_token{
-                            if let Some(bos_token) = &tokenizer_config.bos_token{
-                                let bos_token_id = tokenizer.token_to_id(&bos_token).expect("Should have found the bos token id");
-                                special_tokens.push((bos_token.clone(), bos_token_id));
-                                single.push(bos_token.to_string());
-                            }
-                        }
-                        single.push("$0".to_string());
-                        if let Some(true) = &tokenizer_config.add_eos_token{
-                            if let Some(eos_token) = &tokenizer_config.eos_token{
-                                let eos_token_id = tokenizer.token_to_id(&eos_token).expect("Should have found the eos token id");
-                                special_tokens.push((eos_token.clone(), eos_token_id));
-                                single.push(eos_token.to_string());
-                            }
-                        }
-                        let post_processor = TemplateProcessing::builder().try_single(single).unwrap().special_tokens(special_tokens).build().unwrap();
+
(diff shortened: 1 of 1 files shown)


router/src/main.rs
```diff
@@ -304,37 +304,21 @@ async fn main() -> Result<(), RouterError> {
...
-                        }
-                        let post_processor = TemplateProcessing::builder().try_single(single).unwrap().special_tokens(special_tokens).build().unwrap();
+
+    let tokenizer: Option<Tokenizer> = tokenizer_filename.and_then(|filename| {
+        let mut tokenizer = Tokenizer::from_file(filename).ok();
+        if let Some(tokenizer) = &mut tokenizer {
+            if let Some(class) = &tokenizer_config.tokenizer_class {
+                if class == "LlamaTokenizer" || class == "LlamaTokenizerFast" {
+                    tracing::info!("Overriding LlamaTokenizer with TemplateProcessing to follow python override defined in https://github.com/huggingface/transformers/blob/4aa17d00690b7f82c95bb2949ea57e22c35b4336/src/transformers/models/llama/tokenization_llama_fast.py#L203-L205");
+                    if let Some(post_processor) = create_post_processor(tokenizer, &tokenizer_config) {
```
Narsil, line 314: We need to override only is the current post_processor is None, forgot that condition.

router/src/main.rs
```diff
@@ -543,6 +527,74 @@ pub async fn get_tokenizer_config(api_repo: &ApiRepo) -> Option<HubTokenizerConf
     Some(tokenizer_config)
 }
 
+/// Create a post_processor for the LlamaTokenizer
+pub fn create_post_processor(
+    tokenizer: &Tokenizer,
+    tokenizer_config: &HubTokenizerConfig,
+) -> Option<TemplateProcessing> {
```
Narsil, line 534: Why not return an Result and remove all those unwrap() into `?`?

router/src/main.rs
```diff
@@ -304,37 +304,21 @@ async fn main() -> Result<(), RouterError> {
...
-                        let post_processor = TemplateProcessing::builder().try_single(single).unwrap().special_tokens(special_tokens).build().unwrap();
+
+    let tokenizer: Option<Tokenizer> = tokenizer_filename.and_then(|filename| {
+        let mut tokenizer = Tokenizer::from_file(filename).ok();
+        if let Some(tokenizer) = &mut tokenizer {
+            if let Some(class) = &tokenizer_config.tokenizer_class {
+                if class == "LlamaTokenizer" || class == "LlamaTokenizerFast" {
+                    tracing::info!("Overriding LlamaTokenizer with TemplateProcessing to follow python override defined in https://github.com/huggingface/transformers/blob/4aa17d00690b7f82c95bb2949ea57e22c35b4336/src/transformers/models/llama/tokenization_llama_fast.py#L203-L205");
+                    if let Some(post_processor) = create_post_processor(tokenizer, &tokenizer_config) {
                         tokenizer.with_post_processor(post_processor);
```
Narsil, line 315: Let's log only when we actually override, here we would log even if the function fails.

router/src/main.rs
```diff
@@ -543,6 +527,74 @@ pub async fn get_tokenizer_config(api_repo: &ApiRepo) -> Option<HubTokenizerConf
...
+        panic!("add_bos_token = true but bos_token is None");
+    }
+
+    if add_eos_token && eos_token.is_none() {
+        panic!("add_eos_token = true but eos_token is None");
+    }
+
+    let mut single = String::new();
+    let mut pair = String::new();
+    let mut special_tokens = Vec::new();
```
Narsil, line 551: Can we keep the actual vec ? `String` is really a clutch.

router/src/main.rs
drbh: updated and added in latest commit

router/src/main.rs
drbh: great point! updated in latest

router/src/main.rs
drbh: updated and moved after `create_post_processor` succeeds

router/src/main.rs
drbh: yep, updated!

Narsil (approved):
LGTM

PR #2140, opened 2024-06-28
feat: download lora adapter weights from launcher

drbh:
This PR updates the launcher to download any lora adapters passed via the env or cli args. 

This is enabled by updating the `download_convert_model` signature to pass all the required values instead of the `Args` struct and reusing the function for each adapter id.


Narsil (approved):
LGTM.

PR #2141, opened 2024-06-28
fix: use weights from base_layer

drbh:
This PR fixes a bug in the `rocm` path for llama. This bug was introduced in the lora adapter pr https://github.com/huggingface/text-generation-inference/pull/2010 and attempts to incorrectly read `.linear.weight` from the `gate_up_proj`.

This throws because `gate_up_proj` is now of type `TensorParallelMultiAdapterLinear` which wraps the type it used to be `TensorParallelColumnLinear`. In order to get the raw weight we need to access it via the `base_layer`


fxmarty (approved):
Thanks there is the same in `flash_mistral_modeling.py`.

To be fair it is not at all your fault, having this controlflow for rocm here is bad anyway & we still don't have the hosted rocm CI to catch that

PR #2148, opened 2024-07-01
fix microsoft/Phi-3-mini-4k-instruct crash in batch.slots[batch.slot_…

sywangyi:
@Narsil

### DIFF
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -309,7 +309,7 @@ async fn main() -> Result<(), RouterError> {
         let mut tokenizer = Tokenizer::from_file(filename).ok();
         if let Some(tokenizer) = &mut tokenizer {
             if let Some(class) = &tokenizer_config.tokenizer_class {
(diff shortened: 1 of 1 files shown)


Narsil:
We can't fix it that way (The initial goal was the fix Phi3 and DeepSeek).

Seems like we just need to override the processor because the one defined on the hub is wrong. https://huggingface.co/microsoft/Phi-3-mini-4k-instruct/discussions/75/files

router/src/main.rs
```diff
@@ -309,7 +309,7 @@ async fn main() -> Result<(), RouterError> {
         let mut tokenizer = Tokenizer::from_file(filename).ok();
         if let Some(tokenizer) = &mut tokenizer {
             if let Some(class) = &tokenizer_config.tokenizer_class {
-                if (class == "LlamaTokenizer" || class == "LlamaTokenizerFast") && tokenizer.get_post_processor().is_none() {
+                if (class == "LlamaTokenizer" || class == "LlamaTokenizerFast") && (tokenizer.get_post_processor().is_none() || tokenizer.get_post_processor().unwrap().added_tokens(false) == 0) {
```
Narsil, line 312:
```suggestion
                if (class == "LlamaTokenizer" || class == "LlamaTokenizerFast"){
```

Narsil (approved):

PR #2165, opened 2024-07-02
Update specification

aymeric-roucher:
# What does this PR do?

Add specification for `response_format` in `openapi.json` to allow for `huggingface_hub.InferenceClient` to use it.

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -903,6 +903,16 @@
             "example": 0.1,
             "nullable": true
           },
+          "response_format": {
+            "type": "array",
+            "items": {
+              "type": "string",
+              "value": "string"
+            },
+            "description": "The constrained response format. 'type' can be 'regex' or 'json_object'.",
+            "example": "null",
+            "nullable": true
+          },
           "seed": {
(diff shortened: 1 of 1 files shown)


Narsil:
This file is normally generated directly from the spec as defined here:

https://github.com/huggingface/text-generation-inference/blob/main/router/src/lib.rs#L825

@drbh Is this normal the spec is out of date ? (If yes, we should have something in CI check it, if not, we should update the Rust code to get that thing in).

Wauplin:
+1 on having `openapi.json` up to date with a CI check. We are now using it to generate `InferenceClient` types and automating this would help getting things on-par! :pray:

drbh:
It shouldn't be out of sync but currently CI doesn't check if it is up to date. 

However, this PR should fix those issues and will ensure that the spec is up to date https://github.com/huggingface/text-generation-inference/pull/2169. It also includes an updated `openapi.json`.

Narsil:
Going ahead an closing this since #2169 takes care of it (Happy to make you a contributor @aymeric-roucher if you're not already)

PR #2169, opened 2024-07-02
feat: improve update_docs for openapi schema

drbh:
This PR aims to help sync the openapi schema ~~via a combination of pre-commit hooks and a new cli arg to dump the spec file~~

This PR now updates `update_doc.py` to have two commands allowing callers to update the launcher docs and or the openapi spec

```bash
python update_doc.py -h
usage: update_doc.py [-h] {openapi,md} ...

Update documentation for text-generation-launcher

positional arguments:
  {openapi,md}
    openapi     Update OpenAPI documentation
    md          Update launcher and supported models

options:
  -h, --help    show this help message and exit
```


.github/workflows/autodocs.yaml
```diff
@@ -14,7 +14,23 @@ jobs:
...
         echo text-generation-launcher --help
-        python update_doc.py --check
+        python update_doc.py md --check
+
+    - name: Install Protoc
+      uses: arduino/setup-protoc@v1
+    - name: Clean unused files
+      run: |
+        sudo rm -rf /usr/local/lib/android # will release about 10 GB if you don't need Android
+        sudo rm -rf /usr/share/dotnet # will release about 20GB if you don't need .NET
```
Narsil, line 28: Where is that coming from ?Feels very random I don't think it' anywhere in our current CI.

docs/openapi.json
```diff
@@ -192,7 +190,6 @@
           "Text Generation Inference"
         ],
         "summary": "Generate a stream of token using Server-Sent Events",
```
Narsil, line 192: Why are the descriptions gone ?

docs/openapi.json
```diff
@@ -756,34 +739,19 @@
...
-            "type": "string",
-            "example": "What is Deep Learning?",
-            "nullable": true
-          },
-          "role": {
-            "type": "string",
-            "example": "user"
+        "oneOf": [
+          {
+            "$ref": "#/components/schemas/TextMessage"
```
Narsil, line 749: Feels the cleaner types also help here, no ?

update_doc.py
```diff
@@ -122,15 +126,155 @@ def check_supported_models(check: bool):
...
+        print("Error: Invalid JSON received from the server")
+        raise SystemExit(1)
+    except Exception as e:
+        print(f"An unexpected error occurred: {e}")
+        raise SystemExit(1)
 
 
 if __name__ == "__main__":
-    main()
+    parser = argparse.ArgumentParser(
```
Narsil, line 253: Can we keep the simpler version that just checks everything by default ?
instead just add `check_openapi` function ?

update_doc.py
```diff
@@ -122,15 +126,155 @@ def check_supported_models(check: bool):
...
 
-def main():
-    parser = argparse.ArgumentParser()
-    parser.add_argument("--check", action="store_true")
+def start_server_and_wait():
+    log_file = open("/tmp/server_log.txt", "w")
 
-    args = parser.parse_args()
+    process = subprocess.Popen(
+        ["text-generation-launcher"],
```
Narsil, line 133: Can we use an extremely small model ?

Narsil: Maybe to simplify we could add a flag in `router` in order to print the openapi directly ?
Would remove the necessity to compile things (except the router).

I know I said internally I didn't like that option, but I also said after that you were right since I didn't have a better idea.

update_doc.py
```diff
@@ -122,15 +126,155 @@ def check_supported_models(check: bool):
...
+    response.raise_for_status()
+    return response.json()
+
+
+def update_openapi_json(new_data, filename="docs/openapi.json"):
+    with open(filename, "w") as f:
+        json.dump(new_data, f, indent=2)
+
+
+def compare_openapi(old_data, new_data):
```
Narsil, line 183: Honestly I wouldn't even ever parse the files and simply reuse some simple diffing.

```
diff = subprocess.run(
                    ["diff", tmp, filename], capture_output=True
                ).stdout.decode("utf-8")
print(diff)
```

Not strong opinion, your code seems cleaner, but given the unreadableness of the original JSON file, I'm not sure it's worth potentially making mistakes in that diff algorithm.

docs/openapi.json
drbh: this is related to an upstream change where utopia added support for a custom description instead of using the same value for both description and summary (related PRs https://github.com/juhaku/utoipa/issues/781 and https://github.com/juhaku/utoipa/pull/948).

update_doc.py
drbh: yep that makes sense, updated in the latest commits

update_doc.py
drbh: updated and added a `text-generation-router print-schema` command that prints the api docs as JSON and exits, if no command is passed it goes through the normal flow

update_doc.py
drbh: good points! I think `diff` is a better long term solution and the pr has been updated to use `diff`

Narsil (approved):

PR #2180, opened 2024-07-03
feat: simple mistral lora integration tests

drbh:
This PR adds integration tests for `mistralai/Mistral-7B-v0.1` with two lora adapters `predibase/dbpedia` and `predibase/customer_support`. These test will be improved in the future, and exist to ensure that basic lora functionality works as expected

### DIFF
--- a/integration-tests/models/test_lora_mistral.py
+++ b/integration-tests/models/test_lora_mistral.py
@@ -1,0 +1,134 @@
+import pytest
+import requests
+
+
+@pytest.fixture(scope="module")
+def lora_mistral_handle(launcher):
+    with launcher(
+        "mistralai/Mistral-7B-v0.1",
+        lora_adapters=[
+            "predibase/dbpedia",
+            "predibase/customer_support",
+        ],
+        cuda_graphs=[0],
+    ) as handle:
+        yield handle
+
+
+@pytest.fixture(scope="module")
+async def lora_mistral(lora_mistral_handle):
+    await lora_mistral_handle.health(300)
+    return lora_mistral_handle.client
+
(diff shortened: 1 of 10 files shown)


integration-tests/models/test_lora_mistral.py
```diff
@@ -0,0 +1,134 @@
...
+
+@pytest.fixture(scope="module")
+def lora_mistral_handle(launcher):
+    with launcher(
+        "mistralai/Mistral-7B-v0.1",
+        lora_adapters=[
+            "predibase/dbpedia",
+            "predibase/customer_support",
+        ],
+        cuda_graphs=[0],
```
Narsil, line 13: Why ?

drbh: currently theres a bug with lora and cuda graphs but wanted to make sure there were basic tests for expected responses. I'll push a update for cuda graphs soon along with some of the other refactoring changes. Happy to keep this open until it's resolved

Narsil: What is the bug ?

Does load with many different loader breaks things ?

drbh: currently the adapter is ignored because now adapter_data is passed to model.forward and is not inlcuded in warmup. I need to update the graph warmup to include adapter data and also replace the values before calling replay

drbh:
this PR has been updated to disable CUDA_GRAPHS if lora adapters are enabled. In the future we'll add support however until support is added we will disable and warn the user

OlivierDehaene (approved):

PR #2193, opened 2024-07-05
fix: refactor adapter weight loading and mapping

drbh:
This PR refactors the loading and mapping of lora adapters and avoids the need for model specific changes for adapters. The goal of this PR is to simplify the loading flow and avoid modify modeling code for loras to work.


drbh:
this PR has been updated to include the ability to load lora adapters from a local directory. 

the path to the lora adapter can be specified in the following way: 
```
LORA_ADAPTERS=adapter_id=/dir/path
```

note it's possible to mix `adapter_ids` with `adapter_id=adapter_path` e.g. 
```
LORA_ADAPTERS=predibase/dbpedia,myadapter=/path/to/dir/
```

server/text_generation_server/utils/adapter.py
```diff
@@ -194,3 +210,70 @@ def load_module_map(
         adapter_weights, weight_names
     )
     return module_map, adapter_config, adapter_weight_names, adapter_tokenizer
+
+
```
OlivierDehaene, line 214: Isn't this super brittle? It only works for llama, no?

drbh: its a bit brittle but should work in most cases, currently this will correctly load in weight for llama, mistral and gemma type models (just added some updates and tests too). The intention of moving this code here is to avoid mapping the weights inside of each models code.. currently this is the best approach I have, however I'm happy to make any changes!

OlivierDehaene (approved):

danieldk (commented):
Looks like a really nice improvement 🎉. Had two questions.

server/text_generation_server/models/__init__.py
```diff
@@ -1056,3 +1066,115 @@ def get_model(
...
+
+    if len(lora_adapters) > 0:
+        target_to_layer = build_layer_weight_lookup(model.model)
+
+        for index, adapter in enumerate(lora_adapters):
+            # currenly we only load one adapter at a time but
+            # this can be extended to merge multiple adapters
+            adapter_parameters = AdapterParameters(
+                adapter_info=[adapter],
+                weights=None,  #  will be set to 1
```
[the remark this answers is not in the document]
drbh: ah yes, the comment isn't very descriptive. I've added more details it in the latest commit

server/text_generation_server/models/__init__.py
```diff
@@ -1056,3 +1066,115 @@ def get_model(
...
+                merge_strategy=0,
+                density=1.0,
+                majority_sign_method=0,
+            )
+
+            adapter_index = index + 1
+            adapter_to_index[adapter.id] = adapter_index
+
+            if adapter_index in model.loaded_adapters:
```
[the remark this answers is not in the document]
drbh: good catch, this shouldn't ever happen, removed.

tensimixt:
> This PR refactors the loading and mapping of lora adapters and avoids the need for model specific changes for adapters. The goal of this PR is to simplify the loading flow and avoid modify modeling code for loras to work.

Does this also apply to lora adapters generated by using the mistral-finetune repo? As the weights in these adapters use keys that are in the consolidated.safetensors file typically found in recent mistral huggingface repos, and these differ from conventional key names. The best example is Mistral 7b instruct v0.3. 

For example have tried to use a lora adapter generated from mistral-finetune but it did not work with TGI. It also did not work with vllm. Also to get it in a state where it might work with vllm had mapped the weights keys to what is conventionally used but found lm_head, embed_tokens and layernorm weight keys that vllm could not handle. Will this PR address these issues also? 

Thank you!

drbh:
Hi @tensimixt, thank you for your question. If `mistral-finetune` loads LoRA with different adapter names than expected, this may not be resolved in this PR. The goal of this PR is to simplify the LoRA logic. Improvements, including support for mistral-finetune, will be explored in future PRs.

server/text_generation_server/utils/adapter.py
danieldk: I guess we could have a mapping table here per model type? But probably better for another PR.

server/text_generation_server/models/__init__.py
```diff
@@ -294,7 +304,7 @@ class ModelType(enum.Enum):
     __GLOBALS[data.name] = data.value["type"]
 
 
-def get_model(
+def _get_model(
```
danieldk, line 307: I think this `_get_model` and `get_model` naming is not really descriptive. Maybe this should still be named `get_method` and the function below `get_model_with_lora_adapters` or something like that?

danieldk (dismissed):
Added a comment about `_get_model`/`get_model`, maybe the naming can be improved?

Looks good to me otherwise.

server/text_generation_server/models/__init__.py
drbh: updated to use `get_model_with_lora_adapters` and renamed `_get_model` back to `get_model`

mhou7712:
Any update about this PR?  We will have this PR included in the next release build?  Thanks.

ErikKaum:
Hi @mhou7712 👋 

So this PR (#2193) has been merged and should be in the next release 👍

mhou7712:
Hi @ErikKaum , cool and thanks a lots!!

PR #2194, opened 2024-07-05
Move quantized weight handling out of the `Weights` class

danieldk:
# What does this PR do?

Quantized weights were loaded in the `Weights` class, but this was getting quite unwieldy, where every higher level method to load weights was a long conditional to cover all the different quantizers.

This change moves loading of quantized weights out of the `Weights` class. This is done by defining a simple `WeightsLoader` interface that is implemented by `Exl2WeightsLoader`, `GPTQWeightsLoader`, and `MarlinWeightsLoader`. These implementations are in the quantizers' respective modules. The `Weights` class provides the low-level load operations (such as loading tensors or sharded tensors), but delegates loads that need quantizer-specific weight processing to a loader. The loaders still use the low-level functionality provided by `Weights`.

I initially tried making a hierarchy where a class like `GPTQWeights` would inherit from `Weights`. But it is not very flexible (e.g. does not work well with the new weight storage mock used in tests) and the implicit indirections made the code harder to follow.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/utils/weights.py
+++ b/server/text_generation_server/utils/weights.py
@@ -1,13 +1,88 @@
-import os
+from abc import ABC, abstractmethod
 from pathlib import Path
 from typing import Dict, List, Optional, Union
-from safetensors import safe_open, SafetensorError
+from safetensors import safe_open
 import torch
-from loguru import logger
-from huggingface_hub import hf_hub_download
-import json
-from text_generation_server.layers.gptq import GPTQParams
-from text_generation_server.utils.log import log_once
+
+
+class WeightsLoader(ABC):
+    """
+    Instances of this type implement higher-level weight loading.
+
+    At a low-level, every weight is stored in the Safetensors format.
+    The interpretation of weights may be different however, for instance
+    could be packed, quantized weights. Loaders are responsible for
+    interpreting the raw tensors, sharding tensors in a manner compatible
+    with the format, etc.
+    """
+
+    @abstractmethod
+    def get_weights_col_packed(
+        self,
+        weights: "Weights",
+        prefix: str,
+        block_sizes: Union[int, List[int]],
+    ):
+        """
+        Get the packed weights at the given prefix with column-splitting for
+        tensor parallelism. This method should be used when multiple different
+        weights are packed into a tensor, for instance, query/key/value
+        weights or a gate/up projection.
+
+        The `block_sizes` determines the proportions of the packed tensors.
+        The columns are split in equally sized blocks when `block_sizes` is an
+        `int`, or in blocks proportional given to the sizes. For instance
+        `[2, 1, 1]` will divide an input with dimensionality `1024` in
+        `[512, 256, 256]`.
+        """
+        ...
+
+    def get_weights_col(self, weights: "Weights", prefix: str):
+        """
+        Get weights at the given prefix and apply column-splitting for tensor
+        paralllism.
+        """
+        return weights.get_multi_weights_col([prefix], 0)
+
+    @abstractmethod
+    def get_multi_weights_col(self, weights: "Weights", prefixes: List[str], dim: int):
+        """
+        Get the weights at the given prefixes, column-split them for tensor
+        parallelim, and then concatenate the weights along the given dimension.
+        """
+        ...
+
+    @abstractmethod
+    def get_weights_row(self, weights: "Weights", prefix: str):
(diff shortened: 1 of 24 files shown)


server/text_generation_server/utils/weights.py
```diff
@@ -1,13 +1,66 @@
...
+        weights: "Weights",
+        prefix: str,
+        block_sizes: Union[int, List[int]],
+    ):
+        return weights.get_packed_sharded(
+            f"{prefix}.weight", dim=0, block_sizes=block_sizes
+        )
+
+    def get_weights_col(self, weights: "Weights", prefix: str):
+        return weights.get_multi_weights_col([prefix], 0)
```
Narsil, line 56: Isn't that always equivalent to `self.get_multi_weights_col([prefix], 0)` (meaning this could be directly into the core ABC interface ? leaving 1 thing less to implement ?.

server/text_generation_server/utils/weights.py
```diff
@@ -37,6 +91,9 @@ def __init__(
         self.dtype = dtype
         self.process_group = process_group
         self.prefix = prefix
+        self.weights_loader = (
+            _DefaultWeightsLoader() if weights_loader is None else weights_loader
```
Narsil, line 95: I think we could make it not optional and force callers to create one.


Then `get_quantizer_loader` could becom `get_loader` and simply return the Regular one instead of `None`. Should save a bit a default logic everywhere,no ?

server/text_generation_server/utils/weights.py
```diff
@@ -1,13 +1,66 @@
...
+        )
+
+    def get_weights_col(self, weights: "Weights", prefix: str):
+        return weights.get_multi_weights_col([prefix], 0)
+
+    def get_multi_weights_col(self, weights: "Weights", prefixes: List[str], dim: int):
+        w = [weights.get_sharded(f"{p}.weight", dim=0) for p in prefixes]
+        return torch.cat(w, dim=dim)
+
+    def get_multi_weights_row(self, weights: "Weights", prefix: str):
```
Narsil, line 62: It feels a bit weird that `multi_row` doesn't  have the same signature as `mutli_weights`.

What is multi about it ?

server/text_generation_server/utils/weights.py
```diff
@@ -1,13 +1,66 @@
...
+
+    At a low-level, every weight is stored in the Safetensors format.
+    The interpretation of weights may be different however, for instance
+    could be packed, quantized weights. Loaders are responsible for
+    interpreting the raw tensors, sharding tensors in a manner compatible
+    with the format, etc.
+    """
+
+    @abstractmethod
+    def get_weights_col_packed(
```
Narsil, line 20: I think all from this interface could use a bit of doc.


imho there's only 3 methods.


col_packed -  (Q,  K, V are concatenated in the model, therefore with TP we need to know that it's the case to slice into that tensor correctly).

col_multi - Q, K, V are split, we need to know each name to slice properly and fuse everything into a single tensor


row - (Row is currently never packed nor multi therefore a simple sharding algorithm, suffice.

Am I missing something ?

Narsil (commented):
Nice change.

I expected a net negative amount of lines from this (or neutral) but I guess the extra interface make it a bit more verbose.

Regardless it's a nice change.

I think we can make the interface slightly simpler and remove the need for default checking in the weights load.

server/text_generation_server/utils/weights.py
danieldk: Except in Exl2, but we could indeed move the standard implementation to the ABC and only override it in the Exl2 loader.

server/text_generation_server/utils/weights.py
danieldk: Sounds good!

server/text_generation_server/utils/weights.py
danieldk: Fixed.

server/text_generation_server/utils/weights.py
danieldk: Fixed.

server/text_generation_server/utils/weights.py
danieldk: Added docs.

Narsil (approved):
LGTM

PR #2204, opened 2024-07-08
fp8 marlin kernel

flozi00:
Adopted from: https://github.com/vllm-project/vllm/commit/47f0954af0a5aefd0db19875f6bdcbe933d055a9

Pinging @danieldk for interest since he added the marlin kernels

### DIFF
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -66,7 +66,6 @@ enum Quantization {
     Gptq,
     /// 4 bit quantization. Requires a specific Marlin quantized model: <https://hf.co/models?search=marlin>.
     Marlin,
-    FP8marlin,
     /// Bitsandbytes 8bit. Can be applied on any model, will cut the memory requirement in half,
     /// but it is known that the model will be much slower to run than the native f16.
     #[deprecated(
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -66,6 +66,7 @@ enum Quantization {
     Gptq,
     /// 4 bit quantization. Requires a specific Marlin quantized model: <https://hf.co/models?search=marlin>.
     Marlin,
+    FP8marlin,
     /// Bitsandbytes 8bit. Can be applied on any model, will cut the memory requirement in half,
     /// but it is known that the model will be much slower to run than the native f16.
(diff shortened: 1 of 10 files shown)


danieldk:
So this would enable FP8 support pre-H100 when using `quantize="fp8"` when wired up?

flozi00:
Yes, exactly

flozi00:
@danieldk could you please take a look for the packing ?
I am not sure how to repack the normal weights into marlin format with your functions.

Please feel free to overtake this PR if it's a fast shot :)

danieldk:
> @danieldk could you please take a look for the packing ? I am not sure how to repack the normal weights into marlin format with your functions.

I was hoping/thinking that this could just hook into

https://github.com/huggingface/text-generation-inference/blob/5c7c9f13903f09636aaf99210710bf07002cdb87/server/text_generation_server/layers/fp8.py#L21

for compute capabilities that don't support FP8 directly, but that are supported by this Marlin kernel? I don't think it needs any special packing?

danieldk:
Ok, so I think we need to:

1. Convert to int32, though we may want to check what torch's representation is and whether it's possible to cast instead.
2. Run `gptq_marlin_repack` (with empty `perm`/`g_idx`).
3. Permute `scales` (see gptq repack).
4. The result the repack can be used as `B` in the marlin FP8 gemm.

flozi00:
Just pushed changes, do you mean that style ? @danieldk

danieldk:
> Just pushed changes, do you mean that style ? @danieldk

Yeah, that looks in the direction that I was thinking. The scales need to be repeated in the output features dimension before permuting.

Let me know if you want to continue working on this or whether I should pick it up. I can probably do it this week between other things.

flozi00:
It would be great if you could take this, since I got some emergencies over night it have to prioritize, so it would be faster if you would finish it I think.

danieldk:
> It would be great if you could take this, since I got some emergencies over night it have to prioritize, so it would be faster if you would finish it I think.

Ok, I'll work on it. Best of luck with the emergencies!

danieldk:
Closing this in favor of #2213 (to run CI).

PR #2212, opened 2024-07-10
[fix] Modifying base in yarn embedding

SeongBeomLEE:
# What does this PR do?

In yarn embedding, the base is not a fixed value.

In yarn embedding, the base should be changeable to the value specified in the config.

Thanks.

## Who can review?
@Narsil


danieldk (approved):
Thanks!

PR #2217, opened 2024-07-11
feat: Add load tests

Hugoch:
# What does this PR do?

This PR adds automated load tests in CI using [Grafana k6](https://k6.io/).

Two tests are performed:
* **Constant virtual users (VUs) load test**: It simulates a fixed pool of users trying to make as many requests as possible on API during 60 seconds.
* **Constant arrival rate load test**: It simulates a constant rate of user requests arrival, independent of the system’s response rate, during 60 seconds.

Both tests were run for 2 different kinds of inputs:
* 5000 [ShareGPT](https://huggingface.co/datasets/anon8231489123/ShareGPT_Vicuna_unfiltered) prompts randomly selected (variable token length)
* 5000 [ShareGPT](https://huggingface.co/datasets/anon8231489123/ShareGPT_Vicuna_unfiltered) prompts truncated to 500 tokens (constant token length). Tokens count uses`llama-tokenizer`

Test compute the following metrics:
* **Inter token latency**: Time to generate a new output token for each user that is querying the system. It translates as the “speed” perceived by the end-user. We aim for at least 300 words per minute (average reading speed) so ITL<150ms
* **Time to First Token**: Time the user has to wait before seeing the first token of its answer. Lower waiting time are essential for real-time interactions, less so for offline workloads.
* **End to End latency**: The overall time the system took to generate the full response to the user.
* **Throughput**: The number of tokens per second the system can generate across all requests
Successful requests: The number of requests the system was able to honor in the benchmark timeframe
* **Error rate**: The percentage of requests that ended up in error, as the system could not process them in time or failed to process them.

At the end of the test, it produces charts with, on the same plot:
* Results for TGI at current commit
* Results for TGI at previous commit (if any)
* Results for TGI at last release tag (if any)

Results are added to https://github.com/huggingface/text-generation-inference/issues/2235

It relies on run workflow artifacts to gather previous results (90 days TTL).

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
      to it if that's the case.

## Who can review?


Hugoch:
ShareGPT constant arrival rate

load_tests/benchmarks/templates/main.js.j2
```diff
@@ -0,0 +1,131 @@
...
+
+    const startTime = Date.now();
+    let firstTokenTime = null;
+    let lastTokenTime = null;
+    let tokensCount = 0;
+    let response = ""
+
+    const res = sse.open(url, params, function (client) {
+        client.on('event', function (event) {
+            // console.log(event.data)
```
ErikKaum, line 61:
```suggestion
```

load_tests/benchmarks/engine.py
```diff
@@ -0,0 +1,151 @@
...
+        if self.process:
+            kill(self.process.pid)
+        if self.thread:
+            self.thread.join()
+
+
+class TGIDockerRunner(InferenceEngineRunner):
+    def __init__(self,
+                 model: str,
+                 image: str = "ghcr.io/huggingface/text-generation-inference:latest",
```
ErikKaum, line 67: are we sure that this runs on the correct docker image? Since it runs on `main` there shouldn't be any contention on concurrent builds 🤔

ErikKaum (approved):
Nice 🚀

One small comment is that this is still quite gnarly for others to use and run on their own machines. And tbh that's also because some things (like k6-sse) don't have very good DX.

I think that's cool for now, but just good to be aware of 👍

load_tests/benchmarks/engine.py
Hugoch: You are right for PR runs that wouldn't work as we need to pinpoint a specific tag (that has to be done in `load_test.py`). For nightly, the current build process ensures that `latest` is a build run on `main`.

Hugoch:
> Nice 🚀
> 
> One small comment is that this is still quite gnarly for others to use and run on their own machines. And tbh that's also because some things (like k6-sse) don't have very good DX.
> 
> I think that's cool for now, but just good to be aware of 👍

Yeah, I agree. That would be easier to have everything containerized. But in that case we would need to mount the Docker socket in the container to be able to spawn TGI Docker from there or do some kind of Docker in Docker.

load_tests/benchmarks/k6.py
```diff
@@ -0,0 +1,233 @@
...
+                        outputs.append(
+                            {"message": conversation["value"], "num_tokens": len(encoding["input_ids"])})
+                    if len(outputs) >= MAX_SAMPLES:  # limit the number of inputs
+                        break
+            with open("inputs_variable_tokens.json", "w") as f:
+                f.write(json.dumps(outputs))
+
+    def run(self):
+        self.k6_config.executor.render()
+        args = f"/tmp/k6-sse run --out json=results.json {self.k6_config.executor.rendered_file}"
```
martinigoyanes, line 170: Hey there, awesome work @Hugoch !

Following up on @ErikKaum comment, why not use vanilla k6 and  gather Time To First Token, Time Per Output Token, etc... from response headers?

This is a snippet from my own TGI load testing tool, maybe it is useful:
```
        // Extract per request stats
        const end_to_end_time = res.timings.duration;
        const generated_tokens = parseInt(res.headers['X-Generated-Tokens']);
        const inference_time = parseInt(res.headers['X-Inference-Time']);
        const prompt_tokens = parseInt(res.headers['X-Prompt-Tokens']);
        const queue_time = parseInt(res.headers['X-Queue-Time']);
        const time_per_output_token = parseInt(res.headers['X-Time-Per-Token']);
        const total_time = parseInt(res.headers['X-Total-Time']);
        const validation_time = parseInt(res.headers['X-Validation-Time']);

        // Compute overhead time from I/0
        const overhead_time = end_to_end_time - total_time
        // Compute time to first token
        const time_to_first_token = inference_time - generated_tokens * time_per_output_token
        ```

Hugoch: Hey @martinigoyanes ! Thx for the feedback!
Headers can be used but do not precisely account for the whole e2e time with HTTP layer. Time to first token cannot be precisely measured: with  `const time_to_first_token = inference_time - generated_tokens * time_per_output_token` as you assume that time per token is the same for all tokens. It is true in average for all n>1 tokens, but you are missing the prefill phase which will delay first token.

Narsil:
My reviews were not sent for 1 month at least, nice :(

load_tests/benchmarks/k6.py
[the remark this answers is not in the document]
Hugoch: I made k6-sse build optional and now uses the local path binary rather that relying on /tmp

Narsil:
Closing as stale (more coming with new benchmarking tools !)

PR #2220, opened 2024-07-11
usage stats and crash reports

ErikKaum:
# What does this PR do?

High level:
- adds usage stats to TGI
- implementation in `router` + docs
- send 2 events: one on start and one on stop (which might potentially be a crash)
- omit sensitive data
- device info from cuda or xpu, tpu deferred for now due to not having similarly a straightforward way of getting device info
- only gather stats when TGI is run inside of docker
- default in and opt-out with `--disable-usage-stats` flag, can also partially opt out with flag `--disable-crash-reports` which only allows sending of usage stats.

## Before submitting
      Pull Request section?
      to it if that's the case.


router/src/main.rs
```diff
@@ -406,8 +455,30 @@ async fn main() -> Result<(), RouterError> {
...
+        }
+        Err(e) => {
+            if let Some(ref ua) = user_agent {
+                if !disable_crash_reports {
+                    let error_event = usage_stats::UsageStatsEvent::new(
+                        ua.clone(),
+                        usage_stats::EventType::Error(e.to_string()),
+                    );
+                    error_event.send().await;
+                }
```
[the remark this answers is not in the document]
ErikKaum: That's a good point, I'll make that modification

router/src/main.rs
```diff
@@ -374,8 +382,49 @@ async fn main() -> Result<(), RouterError> {
         }
     };
 
+    // Only send usage stats when TGI is run in docker
+    let is_docker = option_env!("DOCKER_LABEL").is_some();
```
Hugoch, line 386: I think this needs env variable to be set in container. For a more reliable way of detecting container, you probably need to look at init process groups to check if `cpuacct` is mounted on `/` or not.

router/src/main.rs
```diff
@@ -374,8 +382,49 @@ async fn main() -> Result<(), RouterError> {
...
+
+    let user_agent = if !disable_usage_stats && is_docker {
+        let reducded_args = usage_stats::Args::new(
```
Hugoch, line 389:
```suggestion
        let reduced_args = usage_stats::Args::new(
```

router/src/main.rs
```diff
@@ -374,8 +382,49 @@ async fn main() -> Result<(), RouterError> {
...
+            validation_workers,
+            json_output,
+            ngrok,
+            messages_api_enabled,
+            disable_grammar_support,
+            max_client_batch_size,
+            disable_usage_stats,
+            disable_crash_reports,
+        );
+        Some(usage_stats::UserAgent::new(reducded_args))
```
Hugoch, line 413:
```suggestion
        Some(usage_stats::UserAgent::new(reduced_args))
```

router/src/usage_stats.rs
```diff
@@ -0,0 +1,241 @@
...
+        writeln!(f, "Docker label: {}", self.docker_label)?;
+        writeln!(f, "nvidia-smi:\n{}", self.nvidia_env)?;
+        write!(f, "xpu-smi:\n{}\n", self.xpu_env)?;
+        write!(f, "System:\n{}", self.system_env)?;
+
+        Ok(())
+    }
+}
+
+fn nvidia_smi() -> Option<String> {
```
Hugoch, line 229: I think `nvidia-smi` output is going to be hard to parse. Maybe you can use its csv format to extract useful metrics:
```
nvidia-smi --query-gpu=name,pci.bus_id,driver_version,pstate,pcie.link.gen.max,pcie.link.gen.gpucurrent,temperature.gpu,utilization.gpu,utilization.memory,memory.total,memory.free,memory.used,reset_status.reset_required,reset_status.drain_and_reset_recommended,compute_cap,ecc.errors.corrected.volatile.total,mig.mode.current,power.draw.instant,power.limit --format=csv
```
That query allows to identify GPUs (model/driver), check for ECC errors and temperature (that may explain a crash), check for memory and MIG (that may explain some OOM), check for reset flags (that may explain a crash) and checks for power limits (that may explain low perf).

router/src/usage_stats.rs
```diff
@@ -0,0 +1,241 @@
...
+    let output = Command::new("nvidia-smi").output().ok()?;
+    let nvidia_smi = String::from_utf8(output.stdout).ok()?;
+    let output = nvidia_smi.replace('\n', "\n   ");
+    Some(output.trim().to_string())
+}
+
+fn xpu_smi() -> Option<String> {
+    let output = Command::new("xpu-smi").arg("discovery").output().ok()?;
```
Hugoch, line 237: Probably same parsing problem as `nvidia-smi`, could be worth checking for a csv output?

docs/source/usage_statistics.md
```diff
@@ -0,0 +1,55 @@
...
+      "model_type": "bloom"
+  },
+  "ngrok": false,
+  "revision": null,
+  "tokenizer_class": "BloomTokenizerFast",
+  "validation_workers": 2,
+  "waiting_served_ratio": 1.2,
+  "docker_label": "latest",
+  "git_sha": "245d3de94877d4910ea7876f6bd19b4d7d4a47d9",
+  "nvidia_env": "N/A",
```
Hugoch, line 41: Probably interesting to add an example with nvidia/TPU info?

Hugoch (commented):
Really cool 👏 !!
I think that container detection needs to be improved as I think it only works in specific Docker envs and not with all container engines.

router/src/main.rs
ErikKaum: Thank you!

router/src/main.rs
ErikKaum: Good point! 👍 

After a small read online I found that this was quite a common way. Do you know if there are any downsides/benefits in checking the `cgroup` vs `cpuacct`?

```rust
pub fn is_container() -> io::Result<bool> {
    let path = Path::new("/proc/self/cgroup");
    let file = File::open(&path)?;
    let reader = io::BufReader::new(file);

    for line in reader.lines() {
        let line = line?;
        // Check for common container runtimes
        if line.contains("/docker/") || line.contains("/docker-") ||
           line.contains("/kubepods/") || line.contains("/kubepods-") ||
           line.contains("containerd") || line.contains("crio") ||
           line.contains("podman") {
            return Ok(true);
        }
    }
    Ok(false)
}
```

Hugoch: `cpuacct` was one option to check in `cgroup` but I like your solution better as you check for them all 👍

ErikKaum: 🫡

ErikKaum:
Notes from offline discussion with @OlivierDehaene 

- we decided to defer the implementation of bubbling up the `InferError::Generation`
- reason is that Olivier is writing an overhaul of the scheduler + backend which anyways allows the bubbling of errors

--> better to get this forward now and not to block on things
--> once the rewrite of the backend is in, we can add the `InferError::Generation` to the usage stats collection

router/src/usage_stats.rs
```diff
@@ -0,0 +1,351 @@
...
+            event_type,
+        }
+    }
+    pub async fn send(&self) {
+        let mut headers = HeaderMap::new();
+        headers.insert("Content-Type", "application/json".parse().unwrap());
+        let body = serde_json::to_string(&self).unwrap();
+        let client = reqwest::Client::new();
+        let _ = client
+            .post(TELEMETRY_URL)
```
Hugoch, line 59:
```suggestion
            .post(TELEMETRY_URL)
            .headers(headers)
```
You are missing the header which fails the ingestion on the other end due to parsing error

Hugoch (commented):
Really cool! 👍

router/src/usage_stats.rs
ErikKaum: Thank you for noticing this 🙌

OlivierDehaene (approved):

Hugoch (approved):
LGTM, great job 🥳

Issue #1896, opened 2024-05-14
HF web service streaming response differs from OpenAI, breaking clients


dluc:
### System Info

Attempting to reuse an existing OpenAI client to stream responses from HF endpoint doesn't work due to a couple of differences. In my case the differences break the .NET client in Azure AI SDK, though I suspect it might affect other clients too.

Differences found:

1. When streaming response tokens, OpenAI terminates the stream with a final `[DONE]` string, while HF simply stops sending tokens. Clients expecting `[DONE]` get stuck waiting either for another token of for the termination string.
2. OpenAI supports '0.0 <= top_p <= 1.0', while HF supports only '0.0 < top_p < 1.0'
3. When sending top_p = 0 to HF endpoint, the service replies `200 OK` with an error `{"error":"Input validation error: `top_p` must be > 0.0 and < 1.0","error_type":"validation"}` and no final `[DONE]`. Given the status code and the lack of a termination, the error is parsed as data and causes a client to hang, waiting for the next token.

### Information

- [ ] Docker
- [ ] The CLI directly

### Tasks

- [ ] An officially supported command
- [ ] My own modifications

### Reproduction

# Example 1: error with top_p = 0

Request:
```
curl -v -X POST https://api-inference.huggingface.co/v1/chat/completions \
    -H "Authorization: Bearer ${HF_KEY}" \
    -H "Content-Type: application/json" \
    -d '{"messages":[{"content":"how much is 1+1","role":"system"}],
      "max_tokens":50,
      "temperature":0,
      "top_p":0.0,
      "presence_penalty":0,
      "frequency_penalty":0,
      "stream":true,
      "model":"NousResearch/Nous-Hermes-2-Mixtral-8x7B-DPO"}'
```

Response:
```
< HTTP/2 200
< date: Tue, 14 May 2024 22:33:46 GMT
< content-type: text/event-stream
< x-compute-type: 2-a100
< x-request-id: ...
< cache-control: no-cache
< access-control-allow-credentials: true
< vary: origin, Origin, Access-Control-Request-Method, Access-Control-Request-Headers
< x-accel-buffering: no
< access-control-allow-origin: *
< x-compute-characters: 67
< x-sha: ...
<
data:{"error":"Input validation error: `top_p` must be > 0.0 and < 1.0","error_type":"validation"}
```

OpenAI returns a response instead (see next).

# Example 2: OpenAI response includes '[DONE]`

Request:
```
[...]

PR #2221, opened 2024-07-11
fix: append DONE message to chat stream

drbh:
This PR adds the missing `[DONE]` message to the end of the `/v1/completions` and `/v1/chat/completions` endpoints.

This should resolve: https://github.com/huggingface/text-generation-inference/issues/1896 and align with the openai spec https://platform.openai.com/docs/api-reference/chat/create#chat-create-stream

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -812,6 +812,10 @@ async fn completions(
             }
         };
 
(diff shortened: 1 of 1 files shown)


OlivierDehaene (approved):
LGTM

PR #2226, opened 2024-07-12
fix custom cache dir 

### LINKED ISSUE #2147: loading fast tokenizer implementation for cached model with `HF_HUB_OFFLINE=1` fails
### System Info

TGI version: >=2.0.4 affected,  2.0.1 not affected
model: irrelevant, tested e.g. with llama-3-8b-instruct.
hardware: 1x NVIDIA H100

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [X] An officially supported command
- [ ] My own modifications

### Reproduction

1. Download model, e.g. with `huggingface-cli download meta-llama/Meta-Llama-3-8B-Instruct`
2. Start TGI container v2.1.0 with `HF_HUB_OFFLINE=1` and `--model-id "meta-llama/Meta-Llama-3-8B-Instruct"`
3. Observe that TGI starts with warnings 
```
Could not find a fast tokenizer implementation for meta-llama/Meta-Llama-3-8B-Instruct
```
and also, requests at /v1/chat/completions fail with `template not found`, although a chat template is present in the downloaded tokenizer_config.json. 

In contrast, if I explicitly link to the cached model and start the TGI container v.2.1.0 with `HF_HUB_OFFLINE=1` and `--model-id "/path/to/cache/huggingface/hub/models--meta-llama--Meta-Llama-3-8B-Instruct/snapshots/<sha>/"`, the startup works fine without that warning. Also, if I start with `HF_HUB_OFFLINE=0` and provide an authentication token, startup works well, too. 

This bug did not appear in TGI v2.0.1. I haven't tested thoroughly, but I believe this bug already appeared in TGI v2.0.4

### Expected behavior

In my understanding, downloading the model with huggingface-cli and starting with `HF_HUB_OFFLINE=1` should always work without problems. In particular, it should be possible to start TGI with a cached model and still use fast tokenizer implementation.

ErikKaum:
# What does this PR do?
- Fixes #2147

### How to reproduce the bug
- env vars `HUGGINGFACE_HUB_CACHE=/some/other/dir` and `HF_HUB_OFFLINE=1` have to be set
- launch with e.g. this tokenizer `text-generation-router --tokenizer-name meta-llama/Meta-Llama-3-8B-Instruct`
- result, the `tokenizer_file_name` is `None` and the config as well 

```
2024-07-12T12:40:29.724257Z  WARN text_generation_router: router/src/main.rs:328: Could not find tokenizer config locally and no API specified
2024-07-12T12:40:29.724292Z  INFO text_generation_router: router/src/main.rs:353: Using config None
2024-07-12T12:40:29.724307Z  WARN text_generation_router: router/src/main.rs:355: Could not find a fast tokenizer implementation for meta-llama/Meta-Llama-3-8B-Instruct
```

### Fix
- check if HUGGINGFACE_HUB_CACHE is set and use that
- if not --> then use the `Cache::default()`

This was most likely to only come up when running TGI in docker since there the `ENV HUGGINGFACE_HUB_CACHE=/data \` is set to something besides the default.

### DIFF
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -295,9 +295,6 @@ async fn main() -> Result<(), RouterError> {
             )
         }
     };
-
-
-
     let config: Option<Config> = config_filename.and_then(|filename| {
         std::fs::read_to_string(filename)
             .ok()
@@ -354,7 +351,6 @@ async fn main() -> Result<(), RouterError> {
         tracing::warn!("Could not find a fast tokenizer implementation for {tokenizer_name}");
         tracing::warn!("Rust input length validation and truncation is disabled");
     }
-
 
     // if pipeline-tag == text-generation we default to return_full_text = true
     let compat_return_full_text = match &model_info.pipeline_tag {
--- a/router/src/main.rs
+++ b/router/src/main.rs
@@ -210,7 +210,11 @@ async fn main() -> Result<(), RouterError> {
     }
     let api = if use_api {
         if std::env::var("HF_HUB_OFFLINE") == Ok("1".to_string()) {
-            let cache = Cache::default();
+            let cache = std::env::var("HUGGINGFACE_HUB_CACHE")
+                .map_err(|_| ())
+                .map(|cache_dir| Cache::new(cache_dir.into()))
+                .unwrap_or_else(|_| Cache::default());
+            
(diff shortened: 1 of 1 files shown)


OlivierDehaene (approved):
Thanks

PR #2244, opened 2024-07-18
fix: adjust default tool choice

drbh:
This PR fixes a regression with the default and auto tool_choice. These changes will ensure that `auto` tools work, and to use `OneOf` in the case tools are provided without a tool_choice

### DIFF
--- a/docs/openapi.json
+++ b/docs/openapi.json
@@ -909,7 +909,7 @@
           "tool_choice": {
             "allOf": [
               {
-                "$ref": "#/components/schemas/ToolType"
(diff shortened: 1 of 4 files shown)


drbh:
This PR has been updated to improve the syntax of the default tool choice. These change make use of the existing `ToolChoice` struct and add an option to explicitly set `NoTool` via `tool_choice="none"`. 

Additionally this PR has been updated to include refactors to the `ToolGrammar` impl that make it more readable and improvements to the error messages if a function call fails

danieldk (approved):
Looks good!

PR #2248, opened 2024-07-18
feat(fp8): use fbgemm kernels and load fp8 weights directly

OlivierDehaene:
@danieldk, since you were the one that reworked the weights logic, do you think there is a better to plug the new fp8 weights in Transformers?


server/text_generation_server/layers/fp8.py
```diff
@@ -40,16 +54,53 @@ def fp8_quantize(weight, qdtype=torch.float8_e4m3fn):
...
-        self.qweight, self.scale = fp8_quantize(weight)
+        self.dtype = dtype
+        self.qweight = qweight
+        self.scale = scale
+        self.scale_upper_bound = scale_upper_bound
 
         self.bias = bias if bias is not None else None
 
+    @classmethod
+    def from_unquant(cls, weight, bias, dtype):
```
danieldk, line 72: Do we need to pass through `scale_upper_bound` here?

server/text_generation_server/layers/fp8.py
```diff
@@ -20,12 +26,20 @@ def get_fp8_linear() -> torch.nn.Module:
...
 
 
-def fp8_quantize(weight, qdtype=torch.float8_e4m3fn):
-    device = weight.device
+def fp8_quantize(weight, scale_upper_bound=None, qdtype=torch.float8_e4m3fn):
+    if HAS_FBGEMM:
+        if scale_upper_bound.device != weight.device:
+            scale_upper_bound = scale_upper_bound.to(weight.device)
+
+        qweight, scale = torch.ops.fbgemm.quantize_fp8_per_row(
```
danieldk, line 34: Does the quantization kernel work on H100? I guess it does, but if it doesn't, we also have to make this on the CUDA capability.

server/text_generation_server/utils/weights.py
```diff
@@ -163,8 +200,13 @@ def get_tensor(self, tensor_name: str, to_device=True):
         tensor = f.get_tensor(tensor_name)
         # Special case for gptq which shouldn't convert
         # u4 which are disguised as int32. Exl2 uses int16
-        # as well.
-        if tensor.dtype not in [torch.int16, torch.int32, torch.int64]:
+        # as well. FP8 uses torch.float8_e4m3fn
+        if tensor.dtype not in [
+            torch.float8_e4m3fn,
```
danieldk, line 205: This would only be needed when we want to support FP8 checkpoints I think? Same below.

server/text_generation_server/utils/weights.py
```diff
@@ -62,6 +64,20 @@ def get_weights_row(self, weights: "Weights", prefix: str):
         ...
 
 
+@dataclass
```
danieldk, line 67: The changes to this file are not needed anymore after #2250.

server/text_generation_server/layers/linear.py
```diff
@@ -92,7 +92,11 @@ def forward(self, inp: torch.Tensor) -> torch.Tensor:
 
 def get_linear(weight, bias, quantize):
     if quantize is None:
-        if SYSTEM == "rocm":
+        if isinstance(weight, FP8Weight):
```
danieldk, line 95: Not needed anymore after #2250.

server/text_generation_server/layers/fp8.py
OlivierDehaene: I don't know from where we would get it from though.

server/text_generation_server/layers/fp8.py
OlivierDehaene: Added check on compute capabilities.

server/text_generation_server/utils/weights.py
OlivierDehaene: Yes that's one of the goal of this PR too.

server/text_generation_server/models/__init__.py
```diff
@@ -302,6 +302,9 @@ def get_model(
         if quantize in ["awq", "exl2", "gptq", "marlin"]:
             # These quantizers only work with float16 params.
             dtype = torch.float16
+        elif quantize == "fp8":
+            # gemm kernels are fp8xfp8->bf16
+            dtype = torch.bfloat16
```
OlivierDehaene, line 307: @danieldk is this compatible with the marlin kernels?

server/text_generation_server/utils/weights.py
```diff
@@ -110,21 +111,92 @@ def get_weights_col_packed(
...
-            weights.get_packed_sharded(
-                f"{prefix}.weight", dim=0, block_sizes=block_sizes
-            ),
+        w = weights.get_packed_sharded(
+            f"{prefix}.weight", dim=0, block_sizes=block_sizes
         )
 
+        if w.dtype == torch.float8_e4m3fn:
+            # FIXME: here to avoid circular import
+            from text_generation_server.layers.fp8 import Fp8Weight
```
OlivierDehaene, line 120: Not happy about this.

server/text_generation_server/utils/weights.py
```diff
@@ -198,20 +270,29 @@ def _has_tensor(self, tensor_name: str):
     def get_shape(self, tensor_name: str):
         return self._get_slice(tensor_name).get_shape()
 
-    def get_tensor(self, tensor_name: str, to_device=True):
+    def get_tensor(self, tensor_name: str, to_device=True, cast=True):
```
OlivierDehaene, line 273: Same not happy about this but this is required to not cast the fp8 scales to bfloat16 and keep them in fp32.

server/text_generation_server/utils/weights.py
```diff
@@ -110,21 +111,92 @@ def get_weights_col_packed(
         prefix: str,
         block_sizes: Union[int, List[int]],
     ):
-
-        return self.weight_class(
-            weights.get_packed_sharded(
-                f"{prefix}.weight", dim=0, block_sizes=block_sizes
-            ),
+        w = weights.get_packed_sharded(
```
OlivierDehaene, line 114: My logic here is that even though some of the weights are in fp8, this should still be manageable by the default loader.
I want to avoid adding `--quantize fp8` for this type of checkpoints but this is a bit of a hack.

server/text_generation_server/layers/fp8.py
```diff
@@ -39,26 +55,70 @@ def fp8_quantize(weight, qdtype=torch.float8_e4m3fn):
 
 
 @dataclass
-class Fp8Weight(Weight):
+class Fp8Weight:
```
danieldk, line 58: Would be nice if this could still inherit `Weight` as an interface.

server/text_generation_server/layers/fp8.py
```diff
@@ -39,26 +55,70 @@ def fp8_quantize(weight, qdtype=torch.float8_e4m3fn):
...
+    weight_scale: Optional[torch.Tensor] = None
+    input_scale: Optional[torch.Tensor] = None
 
     def get_linear(self, bias: torch.Tensor):
-        return get_fp8_linear()(self.weight, bias)
+        if self.weight_scale is None:
+            return get_fp8_linear().from_unquant(self.weight, bias, self.dtype)
+        return get_fp8_linear().from_fp8(
+            self.weight, self.weight_scale, self.input_scale, bias, self.dtype
+        )
```
danieldk, line 69: Nice! Looks like a pattern we could reuse in the future for quantization that is pre-quantized or on the fly.

server/text_generation_server/models/__init__.py
danieldk: Supports both `float16` and `bfloat16`.

However do we want to set the default to this, since most models are `float16`? Is it needed for the fbgemm quantization kernel?

server/text_generation_server/utils/weights.py
danieldk: Hmmm, this breaks the abstraction quite a bit. This class is also used by other (explicit) quantizers like eetq.

I think it would make more sense to put this implementation in something like a `HybridFP8FP16Loader` in the `fp8` module. Then we could add some logic to the [`get_loader`](https://github.com/huggingface/text-generation-inference/blob/e52be9bba2a11f6b2145fcd21be0ba860b977a4d/server/text_generation_server/utils/quantization.py#L95) function, along the lines of: when `quantizer==None` is set, enumerate over the tensors (should be cheap I think for just getting the dtypes?) and then if an FP8 weight is encountered,

server/text_generation_server/utils/weights.py
danieldk: Yeah, I had a discarded PR that did something similar. Maybe we can call it `to_dtype` for consistency with `to_device`?

server/text_generation_server/utils/weights.py
```diff
@@ -110,21 +111,92 @@ def get_weights_col_packed(
...
+            scale = weights.get_packed_sharded(
+                f"{prefix}.weight_scale", dim=0, block_sizes=block_sizes, cast=False
+            )
+            input_scale = weights.get_tensor(f"{prefix}.input_scale", cast=False)
+            return Fp8Weight(
+                weight=w,
+                weight_scale=scale,
+                input_scale=input_scale,
+                dtype=weights.dtype,
+            )
```
SunMarc, line 139: I changed `input_scale` to `input_scale_ub` which is less ambiguous.

server/text_generation_server/models/__init__.py
OlivierDehaene: Yes it's required for fbgemm.
I can add a method in `layers/fp8.py` to check whether we will use fbgemm and set the default appropriately.

OlivierDehaene: Added a check.

server/text_generation_server/utils/weights.py
OlivierDehaene: Agreed, modified the code to add another loader.

danieldk (approved):
Looks great!

PR #2250, opened 2024-07-18
Improve the handling of quantized weights

danieldk:
# What does this PR do?

Handling of quantized weights was split between two mechanisms:

- For quantized checkpoints, we used the new weight loader infrastructure.
- For quantization while loading (EETQ, FP8, bitsandbytes) we instead relied on conditional in `get_linear`.

Weight loaders support context managers to selectively load particular layers with different weight loaders, which is useful for models like Idefics2 AWQ, which uses a quantized text model, but unquantized vision and connector models. However, the context manager would be overrided by `get_linear`, which string-checks `quantizer`. Also, the context manager would not work with EETQ, FP8, and bitsandbytes.

This change migrates all quantizers to the weight loader infrastructure. This has several benefits:

- We can use context managers with all quantizers.
- All the implementation details move down to the quantizer layers, `get_linear` does not need to know how to handle quantizer linear layers.
- All quantizer weights are strongly typed, we don't pass around raw tensors.
- We don't have to pass around the `quantizer` string everywhere.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/layers/fp8.py
+++ b/server/text_generation_server/layers/fp8.py
@@ -1,7 +1,8 @@
-from enum import Enum, auto
+from dataclasses import dataclass
 
 import torch
 from text_generation_server.utils.import_utils import SYSTEM
+from text_generation_server.utils.weights import Weight
 
 
 def get_fp8_linear() -> torch.nn.Module:
@@ -35,6 +36,14 @@ def fp8_quantize(weight, qdtype=torch.float8_e4m3fn):
     qweight = qweight.to(qdtype)
     scale = scale.float().reciprocal()
     return qweight, scale
+
+
+@dataclass
+class Fp8Weight(Weight):
+    weight: torch.Tensor
+
+    def get_linear(self, bias: torch.Tensor):
+        return get_fp8_linear()(self.weight, bias)
 
 
 class Fp8Linear(torch.nn.Module):
--- a/server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
@@ -18,6 +18,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from contextlib import contextmanager
 from typing import List, Optional, Tuple
 
 import torch
@@ -25,7 +26,6 @@ import torch.distributed
 
 from torch import nn
 from transformers.activations import ACT2FN
-from typing import Optional, List, Tuple
 
 from text_generation_server.utils.import_utils import SYSTEM
 from text_generation_server.layers.attention import (
@@ -42,9 +42,15 @@ from text_generation_server.layers import (
     TensorParallelMultiAdapterLinear,
     TensorParallelAdapterRowLinear,
 )
+from text_generation_server.layers.fp8 import Fp8Weight
 from text_generation_server.layers.rotary import PositionRotaryEmbedding
 from text_generation_server.layers.layernorm import (
     FastRMSNorm,
+)
+from text_generation_server.utils.weights import (
+    DefaultWeightsLoader,
+    UnquantizedWeight,
+    Weights,
 )
 
 if SYSTEM == "rocm":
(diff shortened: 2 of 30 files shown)


integration-tests/models/__snapshots__/test_chat_llama/test_flash_llama_simple.json
```diff
@@ -5,19 +5,19 @@
       "index": 0,
       "logprobs": null,
       "message": {
-        "content": "As of your last question, the weather in Brooklyn, New York, is typically hot and humid throughout the year. The suburbs around New York City are jealously sheltered, and at least in the Lower Bronx, there are very few outdoor environments to explore in the middle of urban confines. In fact, typical times for humidity levels in Brooklyn include:\n\n- Early morning: 80-85% humidity, with occas",
+        "content": "As of your last question, the weather in Brooklyn, New York, is typically moderate to warm year-round. The suburban areas around the borough are jealously sheltered from the Northeastern United States' harsh wind and rain systems. In fact, Brooklyn vs the urban confines of Manhattan or Staten Island is energized by an idyllic summer that often sees crisp East Coast air embraced by the mild atmosphere, year after",
```
OlivierDehaene, line 8: Do you know why this changed?

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -336,9 +338,21 @@ def __init__(self, index, prefix, config, weights):
...
-            prefix=f"{prefix}.mlp", config=config, weights=weights, index=index
-        )
+
+        # Only use FP8 when we are not using another quantizer. We should probably
+        # have nicer way of querying this, but for now it'll do.
+        weights_loader = weights.weights_loader
+        if (
+            can_use_fp8()
+            and isinstance(weights_loader, DefaultWeightsLoader)
+            and weights_loader.weight_class is UnquantizedWeight
```
OlivierDehaene, line 348: Doesn't this mean that we will always use fp8 for the MLP? I don't think we want that.
In my opinion it should be the other way around; wrap all the other linear layers with a context manager using UnquantizedWeight if we are in FP8 but that's a bit cumbersome.

I guess that's why the outputs changed.

OlivierDehaene (dismissed):

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
```diff
@@ -470,12 +492,15 @@ class FlashLlamaForCausalLM(torch.nn.Module):
     def __init__(self, prefix: str, config, weights):
         super().__init__()
 
-        self.embed_tokens = TensorParallelEmbedding(
-            prefix=(
-                "model.embed_tokens" if not prefix else f"{prefix}.model.embed_tokens"
-            ),
-            weights=weights,
-        )
+        with no_fp8(weights):
```
OlivierDehaene, line 495: I think you are also missing the `lm_head`? or is not necessary?

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
danieldk: Changed around as you suggested (do nothing, unless `--quantize=fp8`, then exclude attention and piece embeddings).

server/text_generation_server/layers/fp8.py
```diff
@@ -1,7 +1,17 @@
-from enum import Enum, auto
+from dataclasses import dataclass
 
 import torch
 from text_generation_server.utils.import_utils import SYSTEM
+from text_generation_server.utils.weights import Weight
+
+
+def can_use_fp8():
```
OlivierDehaene, line 8: I don't think this is used anymore

server/text_generation_server/models/custom_modeling/flash_llama_modeling.py
danieldk: Added.

server/text_generation_server/layers/fp8.py
danieldk: Indeed, removed.

OlivierDehaene (approved):
Nice!

PR #2262, opened 2024-07-19
feat: add ruff and resolve issue

drbh:
This PR adds [ruff](https://docs.astral.sh/ruff/) linting to TGI and adds ruff to the pre-commit hook


Narsil:
I'm fine moving to ruff. But seeing integration tests failure means something very wrong has happened.

server/text_generation_server/layers/fp8.py
```diff
@@ -12,17 +12,26 @@
...
 
 FBGEMM_MM_AVAILABLE = False
 FBGEMM_DYN_AVAILABLE = False
-try:
-    import fbgemm_gpu.experimental.gen_ai
 
+
+def is_fbgemm_gpu_available():
+    try:
+        return importlib.util.find_spec("fbgemm_gpu.experimental.gen_ai") is not None
```
Narsil, line 24: What's the rationale behind this change ?

I find it must less clear in intent. If ruff complains there must be a good reason behind it.

server/text_generation_server/layers/tensor_parallel.py
```diff
@@ -69,12 +69,12 @@ def load(config, prefix: str, weights):
 
         # GPTQ,AWQ,EETQ don't quantize heads (nor embeddings)
         if config.quantize in ["gptq", "awq", "eetq", "marlin"]:
-            quantize = None
+            pass
```
Narsil, line 72: This feels wrong. We really want to set quantize=None in each of these I think.

server/text_generation_server/models/causal_lm.py
```diff
@@ -233,7 +233,7 @@ def filter(self, request_ids: List[int]) -> Optional["CausalLMBatch"]:
         ]
 
         # Ensure that past_key_values tensors can be updated in-place
-        if type(self.past_key_values[0]) == tuple:
+        if isinstance(self.past_key_values[0], tuple):
```
Narsil, line 236: No. 
type(xxX) == tuple  was definitely intended here, ti's not a rookie mistake.
It's because xxx can be an instance of tuple, yet not a tuple.

I don't remember specificially the issue at hand, but I'd rather trust our past selves than the linter here.

server/text_generation_server/models/custom_modeling/flash_deepseek_v2_modeling.py
```diff
@@ -39,6 +39,12 @@
 from transformers.activations import ACT2FN
 from transformers.configuration_utils import PretrainedConfig
 
+if SYSTEM == "rocm":
+    try:
+        from vllm import _custom_C
```
Narsil, line 44: Where is this coming from ?

This doesn'tseem to be linked to ruff.

server/text_generation_server/models/custom_modeling/t5_modeling.py
```diff
@@ -45,6 +45,15 @@
     SpeculativeHead,
 )
 
+# copied from https://github.com/huggingface/transformers/blob/cd4584e3c809bb9e1392ccd3fe38b40daba5519a/src/transformers/models/t5/modeling_t5.py#L1316
+# Warning message for FutureWarning: head_mask was separated into two input args - head_mask, decoder_head_mask
+__HEAD_MASK_WARNING_MSG = """
```
Narsil, line 50: ???? Why do we need this ?

Narsil (commented):
Thanks a lot.

A few places I'm confused by the changes asked by ruff.
We could merge as-is if necessary, mostly nits.

server/text_generation_server/layers/gptq/utils.py
```diff
@@ -0,0 +1,56 @@
+import torch
```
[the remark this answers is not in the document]
drbh: this resolve a Pyflakes issue of `undefined-name` for the `torch_snr_error` function used in `server/text_generation_server/layers/gptq/quantize.py`. This file adds `torch_snr_error` as copied from `openppl-public/ppq`

server/text_generation_server/layers/tensor_parallel.py
drbh: this is a lint for `unused-variable` as `quantize` is set but never used in this function.

reverted the change but i'm not sure if this check is needed..

@danieldk is there a case we need the `quantize` variable?

server/text_generation_server/layers/rotary.py
```diff
@@ -2,12 +2,13 @@
 import math
 import torch
 from torch import nn
-from loguru import logger
+
+# Inverse dim formula to find dim based on number of rotations
```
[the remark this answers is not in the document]
drbh: ooo stray comment.  removed, thanks!

server/text_generation_server/models/causal_lm.py
drbh: ahhh great catch!!

It seems that a variable can be an instance of a `tuple` via inheritance which would return True, even when its not a base `tuple`

example repro below
```python
class CustomTuple(tuple):
    pass

ct = CustomTuple([1, 2, 3])

print(type(ct) == tuple)  # False
print(isinstance(ct, tuple))  # True
``` 

reverted `isinstance` changes and opted to replace the `==` with `is` to make the linter happy. 

**This should be equivalent, `is` checks for object identity where `==` checks for value, and in the case of built-in types this will be the same.

server/text_generation_server/models/custom_modeling/flash_deepseek_v2_modeling.py
drbh: this is related to a `undefined-name` lint, `_custom_C` is used in the `DeepseekV2MLP.forward` call but was not imported/defined

server/text_generation_server/models/custom_modeling/t5_modeling.py
drbh: this is also a `undefined-name` lint, `__HEAD_MASK_WARNING_MSG` is called as a warning in `T5ForConditionalGeneration.forward` however I'm not sure we need this warning... leaving for now but happy to remove

PR #2284, opened 2024-07-23
feat(router): drop permit after batching

OlivierDehaene:


Narsil:
Closing for now.

We can revisit after chunking has landed

PR #2232, opened 2024-07-15
Feature: During Generation add support for no_repeat_ngram_size

njbrake:
# What does this PR do?

Fixes #2219 

## Before submitting
- [ ] This PR fixes a typo or improves the docs (you can dismiss the other checks if that's the case).
- [x ] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case.
- [ x] Did you make sure to update the documentation with your changes? Here are the
      [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and
      [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).
- [ x] Did you write any new necessary tests?

## Who can review?

Anyone in the community is free to review the PR once the tests have passed. Feel free to tag
members/contributors who may be interested in your PR.
@ErikKaum

Issue #2219, opened 2024-07-11
Support for no_repeat_ngram_size generation parameter


njbrake:
### Feature request

The Transformers library supports the no_repeat_ngram_size parameter for generation. https://huggingface.co/docs/transformers/v4.18.0/en/main_classes/text_generation#transformers.generation_utils.GenerationMixin.generate.no_repeat_ngram_size 
Could this be supported in the TGI library?

### Motivation

This parameter helps to cut down on repetition and helps generate more concise responses.

### Your contribution

I would be happy to help implement this if I can get some pointers about what code needs to be adjusted to accommodate the change? It's unclear to me whether the change needs to be implemented in the Rust layer, the Python layer, or both.

I'd like the parameter to be supported for Llama type models (Llama, Mistral, Phi, etc).

Thanks!

ErikKaum:
Hi @njbrake 👋 

I think it would be a welcomed contribution if you have the bandwidth 👍 

First would be to implement the args parsing in the `text-generation-launcher`: https://github.com/huggingface/text-generation-inference/blob/dbb23fbfa868ad8f961c60896e346fad3d2ab440/launcher/src/main.rs#L170

That should ripple down when starting the `text-generation-server`: https://github.com/huggingface/text-generation-inference/blob/dbb23fbfa868ad8f961c60896e346fad3d2ab440/launcher/src/main.rs#L679

And then in the `text-generation-server` (which is python) actually have the implementation: https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/utils/logits_process.py

Will this help you getting started?

njbrake:
Thanks! Yes that helps me get started. The first two steps you reference (In main.rs) appear to refer to a command to pass when launching the server. `no_repeat_ngram_size` is a generation configuration parameter that should be passed by the client when running an inference request. Do you have any advice about where I need to look for that? It seems like the text-generation-launcher isn't the right place. For example, I would expect to be able to do something like: 

```
from huggingface_hub import InferenceClient
client = InferenceClient()
output = client.chat_completion(
    messages="Finish this sentence....",
    max_tokens=512,
    temperature=0.1,
    no_repeat_ngram_size=7
)
```

ErikKaum:
Ah ofc, sorry I was too fast with that 👍 

So then this is where the server request is parsed (depending on the endpoint):
[...]

PR #2308, opened 2024-07-25
no repeat ngram size ci

ErikKaum:
# What does this PR do?
- To be able to run CI 
- original PR here: https://github.com/huggingface/text-generation-inference/pull/2232


server/text_generation_server/utils/logits_process.py
```diff
@@ -30,6 +30,7 @@ def __init__(
         top_k=None,
         top_p=None,
         typical_p=None,
+        no_repeat_ngram_size=None,
```
drbh, line 33: `no_repeat_ngram_size` is now passed to `StaticWarper` but it doesn't seem like any new warpers that make use of the new argument. Do we need  `NoRepeatNGramLogitsProcessor` appended to `self.warpers`?

benchmark/src/main.rs
```diff
@@ -89,6 +89,11 @@ struct Args {
     #[clap(long, env)]
     frequency_penalty: Option<f32>,
 
+    /// Generation parameter in case you want to specifically test/debug particular
+    /// decoding strategies, for full doc refer to the `text-generation-server`
+    #[clap(long, env)]
+    no_repeat_ngram_size: Option<u32>,
```
drbh, line 95: this comment may need to be updated

drbh (changes requested):
Thanks for adding @ErikKaum looks good but may need a couple changes. 

Additionally it would be nice to add a new `integration-test` that uses the new generation parameter.

server/text_generation_server/utils/logits_process.py
ErikKaum: ah good catch!
Actually there's one commit missing here from the original PR, the `NoRepeatNGramLogitsProcessor` shouldn't be in the warpers in the first place.

ErikKaum: now there shouldn't be things that changed the `StaticWarper` 👍  only stylistic changes

benchmark/src/main.rs
ErikKaum: Should we update all of them of in that case 🤔

drbh: oh haha good point, I didnt realize all the args have a similar comment

server/text_generation_server/utils/logits_process.py
```diff
@@ -20,17 +20,13 @@
     TypicalLogitsWarper,
 )
 
+from transformers.generation.logits_process import _calc_banned_ngram_tokens
+
```
drbh, line 24: I believe ruff is erroring because `_calc_banned_ngram_tokens` is unused. Removing it should make the linter happy

```suggestion
```

drbh (changes requested):

ErikKaum:
Closed, more details in: https://github.com/huggingface/text-generation-inference/issues/2219

PR #2282, opened 2024-07-23
patch-error-on-invalid-grammar

ErikKaum:
# What does this PR do?
- fixes #2280

# Implementation
Try + except the `schema = build_regex_from_schema(schema)` and return `None` from `_cached_compile_fsm` if grammar compilation fails.

There might be a better place to do the error handling. This solution in a way creates two errors, like here:

```bash
2024-07-23T09:06:27.017813Z  INFO text_generation_router::server: router/src/server.rs:1651: Setting max batch total tokens to 16000
2024-07-23T09:06:27.285053Z  INFO text_generation_router::server: router/src/server.rs:1889: Connected
2024-07-23T09:06:35.377377Z ERROR text_generation_launcher: Error compiling FSM: Could not translate the instance {'abc': 'def'} to a
    regular expression. Make sure it is valid to the JSON Schema specification. If
    it is, please open an issue on the Outlines repository
2024-07-23T09:06:35.489362Z ERROR text_generation_launcher: Method Prefill encountered an error.
Traceback (most recent call last):
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/bin/text-generation-server", line 8, in <module>
    sys.exit(app())
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/lib/python3.11/site-packages/typer/main.py", line 311, in __call__
    return get_command(self)(*args, **kwargs)
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/lib/python3.11/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/lib/python3.11/site-packages/typer/core.py", line 778, in main
    return _main(
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/lib/python3.11/site-packages/typer/core.py", line 216, in _main
    rv = self.invoke(ctx)
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/lib/python3.11/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/lib/python3.11/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/Users/erikkaum/miniconda3/envs/text-generation-inference/lib/python3.11/site-packages/click/core.py", line 783, in invoke
[...]

Issue #2280, opened 2024-07-22
Server crashes when provided with invalid JSON schema


npt:
### System Info

text-generation-inference 2.1.1
docker command line:
```
docker run --runtime=nvidia --gpus 4 --shm-size 1g -e HUGGING_FACE_HUB_TOKEN=$TOKEN -e CUDA_VISIBLE_DEVICES=0,1,2,3 -p 8080:80 -v $VOLUME:/data ghcr.io/huggingface/text-generation-inference:2.1.1 --model-id meta-llama/Meta-Llama-3-70B-Instruct --dtype bfloat16 --max-input-tokens 8191 --max-total-tokens 8192
```
model info:
```json
{
  "model_id": "meta-llama/Meta-Llama-3-70B-Instruct",
  "model_sha": "7129260dd854a80eb10ace5f61c20324b472b31c",
  "model_dtype": "torch.bfloat16",
  "model_device_type": "cuda",
  "model_pipeline_tag": "text-generation",
  "max_concurrent_requests": 128,
  "max_best_of": 2,
  "max_stop_sequences": 4,
  "max_input_tokens": 8191,
  "max_total_tokens": 8192,
  "waiting_served_ratio": 0.3,
  "max_batch_total_tokens": 425744,
  "max_waiting_tokens": 20,
  "max_batch_size": null,
  "validation_workers": 2,
  "max_client_batch_size": 4,
  "router": "text-generation-router",
  "version": "2.1.1",
  "sha": "4dfdb481fb1f9cf31561c056061d693f38ba4168",
  "docker_label": "sha-4dfdb48"
}
```

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [X] An officially supported command
- [ ] My own modifications

### Reproduction

1. POST this request body to /generate:
```json
{
  "inputs": "My name is Olivier and I",
  "parameters": {
    "grammar": {"type": "json", "value": {"abc":"def"}}
  }
}
```

2. The server crashes and exits, with these logs:
```
2024-07-22T19:56:58.315280Z ERROR text_generation_launcher: Method Prefill encountered an error.
Traceback (most recent call last):
  File "/opt/conda/bin/text-generation-server", line 8, in <module>
    sys.exit(app())
  File "/opt/conda/lib/python3.10/site-packages/typer/main.py", line 311, in __call__
    return get_command(self)(*args, **kwargs)
[...]

PR #2309, opened 2024-07-25
fix: reject grammars without properties

drbh:
This PR is adds a new check to grammar validation that ensures that the grammar contains `properties` 

For a grammar to be valid it needs to 1. be a valid `JSON Schema`, and 2. have `properties` that are used to constrain the output.

This PR is related to 
pr: https://github.com/huggingface/text-generation-inference/pull/2282 
issue: https://github.com/huggingface/text-generation-inference/issues/2280

### req/res example

The following request would crash TGI due to outlines failing to parse the grammar, now an error is returned because `value` does not include `properties`

```bash
curl localhost:3000/generate \
    -X POST \
    -d '{
        "inputs": "My name is Olivier and I",
        "parameters": {
            "grammar": {"type": "json", "value": {"abc":"def"}}
            }
        }'
```

now returns 

```json
{
    "error": "Input validation error: grammar is not valid: Grammar must have a 'properties' field",
    "error_type": "validation"
}
```

### DIFF
--- a/router/src/validation.rs
+++ b/router/src/validation.rs
@@ -353,6 +353,14 @@ impl Validation {
                             .compile(&json)
                             .map_err(|e| ValidationError::InvalidGrammar(e.to_string()))?;
 
+                        // The schema can be valid but lack properties.
+                        // We need properties for the grammar to be successfully parsed in Python.
+                        // Therefore, we must check and throw an error if properties are missing.
+                        json.get("properties")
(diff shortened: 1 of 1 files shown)


ErikKaum:
Nice 👍 this is the 80% catch that I wasn't able to find!

ErikKaum (approved):
lgtm 👍

Issue #1977, opened 2024-05-29
[Feature]: Additional metrics to enable better autoscaling / load balancing of TGI servers in Kubernetes


EandrewJones:
### Feature request

TGI provides some valuable metrics on model performance and load today. However, there are still a number of missing metrics, the absence of which poses a challenge for orchestration and autoscaling in Kubernetes.

Here is a list of the metrics that we (K8s serving WG, see below) have identified for inclusion TGI:
| Metric Name                                              | Type      | Unit         | Implemented by TGI Already                   |
| -------------------------------------------------------- | --------- | ------------ | -------------------------------------------- |
| model_load_time                                          | Counter   | Seconds      |                                              |
| time_per_output_token_per_batch_size                     | Histogram | Milliseconds |                                              |
| request_wait_time (total time - time spent on inference) | Hisogram  | Milliseconds |                                              |
| request_queue_time                                       | Histogram | Milliseconds | ? (tgi_request_queue_duration)   |
| max_token_capacity                                       | Counter   | Tokens       |                                              |
| time_per_prefill_token                                   | Histogram | Milliseconds |                                              |
| total_tokens_in_current_batch                            | Gauge     | Tokens       |                                              |
| time_to_first_token                                      | Histogram | Milliseconds |                                              |
| estimated_max_prefill_tokens_per_second                  | Gauge     | Tokens       |                                              |
| estimated_max_batch_before_compute_saturation            | Gauge     | Tokens       |                                              |
| request_input_length                                     | Histogram | Tokens       | $\checkmark$ (tgi_request_input_length)      |
| request_output_length                                    | Histogram | Tokens       | $\checkmark$  (tgi_request_generated_tokens) |
| request_with_evicted_tokens                              | Counter   | Count        |                                              |
| total_evicted_tokens                                     | Counter   | Tokens       |                                              |

### Additional Context
I believe TGI already uses OTel. OTel is in the process of adding support for LLM metrics which TGI may be able to piggyback off for some of the above. For reference, see OTel's [LLM Semantic Convention WG](https://docs.google.com/document/d/1EKIeDgBGXQPGehUigIRLwAUpRGa7-1kXB736EaYuJ2M/edit?pli=1) (Please request access if you are not able to view it).

cc @Narsil @drbh

### Motivation

[...]

PR #2311, opened 2024-07-26
Add model_load_time metric

Edwinhr716:
# What does this PR do?
Adds metric that measures the time spent in downloading the model, loading into GPU memory, and time it takes for the server to be ready to receive a request. 

Because the router is the component that emits the metrics, but the launcher is the one who downloads the model and thus is where the download time is tracked, it is necessary to pass the duration from the launcher to the router. I considered passing it as a CLI argument, but to minimize the number of changes required opted to use an environment variable to do it instead. Open to suggestions as to how to better pass the value to the router. 

To make it easier to work with Rust's Instant type I opted to measure two different values and then add them together: the time it takes to download to the model to launching the router; and the time it takes from launching the router to the router being ready to receive requests. 

This PR is part of the metrics standardization effort. 

Fixes # (issue)

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests),
      Pull Request section?
- [X] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link
      to it if that's the case. #1977 

## Who can review?

### DIFF
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -1162,7 +1162,7 @@ fn spawn_webserver(
     max_input_tokens: usize,
     max_total_tokens: usize,
     max_batch_prefill_tokens: u32,
-    startup_time: u64,
+    download_time: u64,
     shutdown: Arc<AtomicBool>,
     shutdown_receiver: &mpsc::Receiver<()>,
 ) -> Result<Child, LauncherError> {
@@ -1200,8 +1200,6 @@ fn spawn_webserver(
         format!("{}-0", args.shard_uds_path),
         "--tokenizer-name".to_string(),
         args.model_id,
-        "--startup-time".to_string(),
-        startup_time.to_string(),
     ];
 
     // Grammar support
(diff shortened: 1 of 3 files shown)


Narsil:
I'm still confused what's the interest to measure this one huge metric that encompasses so many different things:

- Model downloads, from where
- Potential model conversion
- Actual model load time which can be coming from disk or CPU RAM. 

As a model runner, getting ready times can be interesting, but without more context it's quite useless, no ?
What are users supposed to do with that metric ? It's also never being modified during the server's lifetime, so it's not really probing the system to do monitoring no ?
In our monitoring systems, we only care about the logs showing insights as to what's happening, which contain every step, why they are occurring and how long each step takes.

I have nothing against the PR itself, but adding code without clear reasons for clear benefits is always a bit strange.

Edwinhr716:
> What are users supposed to do with that metric ? 

because it is essentially the startup latency of the model, it is useful in determining the pod autoscaling threshold and frequency. So if model-load-time is above say 40 seconds, the user might not want to decrease the number of pods too often as it takes long to create a new pod once demand rises again.

cc @achandrasekar

PR #2335, opened 2024-07-30
fix: attempt forward on flash attn2 to check hardware support

drbh:
This PR attempt to execute a forward pass to validate that flash attention 2 works on the current hardware. Prior to this change `import flash_attn_2_cuda` could load but `flash_attn_2_cuda.varlen_fwd` fail with `FlashAttention only supports Ampere GPUs or newer`.

This change causes the runtime error when the library is loaded which will set `V2` to False and avoid using flash attn 2 in the forward pass

### DIFF
--- a/server/text_generation_server/layers/attention/cuda.py
+++ b/server/text_generation_server/layers/attention/cuda.py
@@ -173,6 +173,41 @@ def paged_attention(
 try:
     import flash_attn_2_cuda
 
+    # try forwarding to see if it works with all dummy inputs
+    batch_size = 1
+    num_heads = 1
+    head_dim = 1
+    seqlen = 1
+
+    try:
+        flash_attn_2_cuda.varlen_fwd(
+            torch.zeros(batch_size, num_heads, seqlen, head_dim),  # q
+            torch.zeros(batch_size, num_heads, seqlen, head_dim),  # k
+            torch.zeros(batch_size, num_heads, seqlen, head_dim),  # v
+            None,  # out (optional)
+            torch.zeros(batch_size + 1, dtype=torch.int32),  # cu_seqlens_q
+            torch.zeros(batch_size + 1, dtype=torch.int32),  # cu_seqlens_k
+            None,  # alibi_slopes (optional)
+            None,  # q_padded (optional)
+            None,  # k_padded (optional)
+            None,  # v_padded (optional)
+            seqlen,  # max_seqlen_q
+            seqlen,  # max_seqlen_k
+            1.0,  # softmax_scale
+            0.0,  # softmax_lse (default value)
+            False,  # is_causal
+            True,  # return_softmax
+            -1,  # window_size_left
+            -1,  # window_size_right
+            0.0,  # softmax_softcap
+            False,  # deterministic
+            None,  # rng_state (optional)
+        )
+    except RuntimeError as e:
+        raise ImportError(
+            "Flash Attention V2 is not supported on this machine. " f"Error: {e}"
+        ) from e
+
     V2 = True
 except ImportError:
     try:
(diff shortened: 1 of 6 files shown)


server/text_generation_server/layers/attention/cuda.py
```diff
@@ -173,6 +174,41 @@ def paged_attention(
 try:
     import flash_attn_2_cuda
 
+    # try forwarding to see if it works with all dummy inputs
```
danieldk, line 177: Wouldn't it be easier to require the minimum needed CUDA capability? We could even skip the import altogether if the hardware doesn't have the right capability.

Narsil: I agree, it would be simpler.

server/text_generation_server/layers/attention/cuda.py
```diff
@@ -254,9 +290,11 @@ def attention(
         softcap=None,
     ):
         if window_size_left != -1:
-            raise NotImplementedError(
-                "window_size_left is only available with flash attn v2"
+            warnings.warn(
```
Narsil, line 293: Can you also not change that.

It's important to hard crash and not silently ignored.

server/text_generation_server/layers/attention/cuda.py
drbh: yea I agree thats a better method. Updated to use `is_ampere_or_newer = major >= 8 and minor >= 0` in the latest commits and avoid trying the forward pass

server/text_generation_server/layers/attention/cuda.py
drbh: updated in the latest commit to still throw when an invalid value is passed

server/text_generation_server/models/__init__.py
```diff
@@ -484,10 +484,14 @@ def get_model(
...
 
+    is_max_input_within_sliding_window = (
+        max_input_tokens <= sliding_window if max_input_tokens is not None else False
+    )
+
     if (
         (sliding_window is not None and sliding_window != -1)
         and not SUPPORTS_WINDOWING
-        and max_input_tokens > sliding_window
+        and is_max_input_within_sliding_window
```
danieldk, line 494: Maybe I'm misreading this, but shouldn't the exception be raised when the max input is *not* inside the window size?

server/text_generation_server/models/custom_modeling/flash_mistral_modeling.py
```diff
@@ -487,10 +486,10 @@ def __init__(self, prefix: str, config, weights, name=None):
             ),
             weights=weights,
         )
-        self.max_past = config.sliding_window
+        self.max_past = config.sliding_window if SUPPORTS_WINDOWING else None
         self.max_past_tensor = (
-            torch.tensor(config.sliding_window, device=weights.device)
-            if self.max_past is not None
+            torch.tensor(self.max_past, device=weights.device)
```
danieldk, line 491: Doesn't this try to create a tensor with the `None` value if windowing is not supported? Same below.

server/text_generation_server/models/__init__.py
Narsil: Yes this line was correct before.

The only thing that needs to happen is that if `max_input_tokens <= sliding_window` we can set sliding_window to -1 and forget about the windowing.

The code changes need to be much less.

drbh: updated to set `sliding_window` to -1 if `max_input_tokens <= sliding_window` in the latest commit

server/text_generation_server/models/custom_modeling/flash_mistral_modeling.py
drbh: ahh yes you are right 😅, fixed in the latest changes to avoid this all together

danieldk (approved):

PR #2339, opened 2024-07-31
refactor usage stats

ErikKaum:
# What does this PR do?
- refactors the usage stats options to be only one flag with three options instead of two different flags
- small semantic fixes


backends/v3/src/main.rs
```diff
@@ -68,10 +68,8 @@ struct Args {
     disable_grammar_support: bool,
     #[clap(default_value = "4", long, env)]
     max_client_batch_size: usize,
-    #[clap(long, env, default_value_t)]
-    disable_usage_stats: bool,
-    #[clap(long, env, default_value_t)]
-    disable_crash_reports: bool,
+    #[clap(default_value = "on", long, env)]
+    usage_stats: Option<String>,
```
Narsil, line 72: Make it a real enum instead ? Saves you writing the actual parsing logic.

docs/source/usage_statistics.md
```diff
@@ -70,4 +70,6 @@ As of release 2.1.2 this is an example of the data collected:
 
 ## How to opt-out
 
-You can easily opt out by passing the `--disable-usage-stats` to the text-generation-launcher command. This will disable all usage statistics. You can also pass `--disable-crash-reports` which disables sending specific crash reports, but allows anonymous usage statistics.
+By passing the `--usage-stats` to the text-generation-launcher you can control how much usage statistics are being collected.
+`--usage-stats=no-stack` will emit the stack traces from errors and the error types, but will continue to send start and stop events
```
Narsil, line 74:
```suggestion
`--usage-stats=no-stack` will not emit the stack traces from errors and the error types, but will continue to send start and stop events
```

ErikKaum: Thank you!

router/src/server.rs
```diff
@@ -1698,33 +1696,35 @@ pub async fn run(
...
-            disable_usage_stats,
-            disable_crash_reports,
-        );
-        Some(usage_stats::UserAgent::new(reduced_args))
-    } else {
-        None
+    let user_agent = match usage_stats_level {
+        usage_stats::UsageStatsLevel::On | usage_stats::UsageStatsLevel::NoStack
+            if is_container =>
+        {
```
Narsil, line 1702:
```suggestion
    let user_agent = match (usage_stats_level, is_container) {
        (usage_stats::UsageStatsLevel::On | usage_stats::UsageStatsLevel::NoStack, true)  =>
        {
        _ => None
```

Not necessarily better but I tend to try and limit the number of branches whenever possible, so codepaths are easier to reason about.

ErikKaum: yeah, this makes sense to me 👍 thank you!

backends/v3/src/main.rs
ErikKaum: Ah yeah ofc, I missed the derive ValueEnum 👍

backends/v3/src/main.rs
```diff
@@ -68,10 +68,8 @@ struct Args {
     disable_grammar_support: bool,
     #[clap(default_value = "4", long, env)]
     max_client_batch_size: usize,
-    #[clap(long, env, default_value_t)]
-    disable_usage_stats: bool,
-    #[clap(long, env, default_value_t)]
-    disable_crash_reports: bool,
+    #[clap(default_value = "on", long, env)]
+    usage_stats: Option<usage_stats::UsageStatsLevel>,
```
Narsil, line 72: Since you're defining a default (And you can define it on the enum) you can skip the `Option` here.

Narsil: With the enum default I think you can even skip the value in clap `#[clap(default, long env)]`

ErikKaum: True, I'll delete the option from `launcher/src/main.rs` as well.

With `#[clap(default, long env)]` I got a compiler error, `attribute `default` requires a value`.
So seems like it's not supported.

PR #2346, opened 2024-08-01
fix EleutherAI/gpt-neox-20b does not work in tgi

sywangyi:
@OlivierDehaene OR @Narsil  @danieldk please help review.
found gpt-neox-20b does not work in tgi. there's shape issue in rotary_emb
https://github.com/huggingface/transformers/blob/main/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L221-L233
add similar logic in tgi could fix the crash.


sywangyi:
@drbh

drbh:
Hi @sywangyi, thanks for the contribution! Just tested with `EleutherAI/gpt-neox-20b` locally and it loads and generates as expected 🙏

PR #2357, opened 2024-08-02
[TENSORRT-LLM] - Implement new looper thread based backend

mfuntowicz:
Current backend implementation relies on locking mecanism to access, within each tokio's requests context thread, the executor on the C++ side. 

This locking results in a heavy contention for all clients except the one which already acquired the lock in write mode. 
This is inherent to the tokio's RwLock implementation which is giving priority to writers.

This backend implement a new, background thread based, TensorRT-LLM backend. 
Everything related to the executor happens within this sole background thread and thus doesn't need any `Sync` nor `RwLock` to be implemented to make it work. 

The only required trait is `Send` to be able to move the backend from the creator thread to the background thread in order to catch backend creation issue.


Narsil (dismissed):
LGTM.

Narsil (approved):

PR #2359, opened 2024-08-05
feat: include local lora adapter loading docs

drbh:
This PR update the lora docs to add instructions for loading loras from a local path


drbh:
merging as this only updates docs

PR #2365, opened 2024-08-06
fix: prefer original layernorm names for 180B

drbh:
This PR prefers the `ln_attn` layernorm weight prefix as used in Falcon 180B. Other falcon models, both V2 and V2 include use `input_layernorm` in other model sizes


drbh:
merging as this should resolve the issue in the inference API, however we should revisit with a better solution (not based on number of layers)

Issue #2363, opened 2024-08-06
Build Intel CPU optimized image automatically


Feelas:
Hello, we are looking for the best way for deploying TGI on Xeons. 

I understand that container images tagged with `x.y.z-intel` are the XPU builds, while `Dockerfile_intel` defines both XPU and CPU paths, XPU being the default. I have successfully ran manual builds for CPU and those works great.

By using the default `x.y.z-intel` tag, one launches the XPU-optimized variant. On Xeons, this causes `Target function add_rms_norm on cpu haven't implemented yet.` errors which I don't encounter when building the CPU image manually (via modified `PLATFORM` argument).

@sywangyi @Narsil would it be possible to automatically build the CPU-optimized variant w/ IPEX & upload as part of CI? 
It boils down to adding additional image build, which runs `Dockerfile_intel` with build_arg `PLATFORM` set to `cpu` instead of default `xpu` and pushing it to a separate tag (e.g. `x.y.z-intel-cpu`).

Additionally, is there any place where all of the available image tags are listed?

Thanks for all of the great work!

ErikKaum:
Thanks for taking this on @sywangyi 🙌

PR #2367, opened 2024-08-07
Cpu dockerimage

sywangyi:
@Narsil could you please review if ci could generate cpu docker image correctly with the PR? And once the dockerimage is generated by ci. I will verify in my side.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -50,6 +50,7 @@ jobs:
                 export label_extension=""
                 export docker_devices=""
                 export runs_on="aws-g6-12xlarge-plus-priv"
+                export platform=""
                 ;;
             rocm)
                 export dockerfile="Dockerfile_amd"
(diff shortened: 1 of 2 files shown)


sywangyi:
@yao-matrix

sywangyi:
try to resolve https://github.com/huggingface/text-generation-inference/issues/2363

Narsil:
This introduces a split in the intel builds, is that desired ?

Also this will break naming because `-intel` would not be used, shouldn't we at least keep using it for either `-cpu` or `-xpu` (whichever makes more sense) ?

sywangyi:
yes, we want to a split in the intel builds, currently intel-cpu and intel xpu use difference torch and ipex.
I use label as intel-cpu, and intel-xpu, WDYT?

Narsil:
I think keeping `intel` for `intel-cpu` makes more sense to me (just to keep backward compatibilty).

I'll go ahead and merge this in the meantime.

Narsil (approved):

sywangyi:
thanks, I see [docker pull ghcr.io/huggingface/text-generation-inference:latest-intel-cp](https://github.com/huggingface/text-generation-inference/pkgs/container/text-generation-inference/256993947?tag=latest-intel-cpu) release, and it can work correctly.

PR #2391, opened 2024-08-09
feat: add guideline to chat request and template

drbh:
This PR adds support for a `guideline` input to chat request. This is required for some models who contain a `guideline` in their template like https://huggingface.co/google/shieldgemma-9b


Narsil (commented):
Shouldn't we add a test for that ? (Unit test in the router ?)

ErikKaum (approved):
this should do the job, thank you 🙌

PR #2392, opened 2024-08-09
Add support for prefix caching to the v3 router

danieldk:
# What does this PR do?

This change adds support for prefix caching to the v3 router. This is broken up from the backend support to ease reviewing.

For now prefix caching is only enabled with `USE_PREFIX_CACHING=1` in this case, the router will switch to `RadixAllocator`. This allocator uses a radix trie to keep track of prefills that were seen prior. If a new prefill is a prefix of a previously-seen prefil, the router will send a request with `prefix_len>0`, which can be used by the backend to decide to reuse KV blocks from the cache, rather than recomputing them.

Even though backend support is not added in this PR, the backend will still work with prefix caching enabled. The prefix lengths are just ignored and not used.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4045,6 +4045,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "slotmap",
  "text-generation-router",
  "thiserror",
  "tokenizers",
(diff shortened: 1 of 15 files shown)


Narsil (approved):
LGTM. I have some remarks for some types, mostly regarding `Arc` vs `References` but I think I get the current choices.

Let's get it merged so we can move forward.

PR #2403, opened 2024-08-12
feat: validate template variables before apply and improve sliding wi…

drbh:
This PR improves the template to throw a better error the `guideline` is included in a template, but not provided in the request


Narsil (approved):
LGTM !

PR #2407, opened 2024-08-12
fix: include create_exllama_buffers and set_device for exllama

drbh:
This PR reimports `set_device` and `set_device` which are later used in `Warmup` https://github.com/huggingface/text-generation-inference/blob/9a7830bd287fb1ca2a3e94f208acd2eb881eb311/server/text_generation_server/server.py#L104 and the buffer is used here https://github.com/huggingface/text-generation-inference/blob/9a7830bd287fb1ca2a3e94f208acd2eb881eb311/server/text_generation_server/layers/gptq/exllamav2.py#L214

This should resolve the ex2 test cases that are receiving a None instead of the pointer to the buffer


drbh:
merging as this fixes imports that were accidentally removed in #2262 and adds missing imports

Issue #2388, opened 2024-08-09
[BUG] Running FP8 quantized model fails on NVIDIA L4 (repack_fp8_for_marlin)


DrNochi:
### System Info

- **Hardware**: AWS g6.12xlarge (us-east-2) / 4x NVIDIA L4 GPU
- **OS**: Ubuntu 24.04 LTS (Noble Numbat)
- **NVIDIA Driver**: nvidia-open 560.28.03
- **CUDA**: 12.6
- **Docker**: Docker version 27.1.1, build 6312585
- **NVIDIA Container Toolkit**: 1.16.1
- **TGI**: ghcr.io/huggingface/text-generation-inference:latest (4b44be4c038f)

```
+-----------------------------------------------------------------------------------------+
| NVIDIA-SMI 560.28.03              Driver Version: 560.28.03      CUDA Version: 12.6     |
|-----------------------------------------+------------------------+----------------------+
| GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
| Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
|                                         |                        |               MIG M. |
|=========================================+========================+======================|
|   0  NVIDIA L4                      Off |   00000000:38:00.0 Off |                    0 |
| N/A   41C    P8             16W /   72W |       1MiB /  23034MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   1  NVIDIA L4                      Off |   00000000:3A:00.0 Off |                    0 |
| N/A   42C    P8             17W /   72W |       1MiB /  23034MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   2  NVIDIA L4                      Off |   00000000:3C:00.0 Off |                    0 |
| N/A   41C    P8             17W /   72W |       1MiB /  23034MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
|   3  NVIDIA L4                      Off |   00000000:3E:00.0 Off |                    0 |
| N/A   38C    P8             16W /   72W |       1MiB /  23034MiB |      0%      Default |
|                                         |                        |                  N/A |
+-----------------------------------------+------------------------+----------------------+
```

### Information

- [X] Docker
- [ ] The CLI directly

### Tasks

- [X] An officially supported command
[...]

PR #2414, opened 2024-08-13
fix: repack for marlin when single scale is provided

drbh:
This PR adjust the conditional for repacking fp8 for marlin to run when a single scale is provided. This avoids a `IndexError` in the case that scales only contain a single value. 

not ~~related to: https://github.com/huggingface/text-generation-inference/issues/2388~~

### DIFF
--- a/server/text_generation_server/layers/marlin/fp8.py
+++ b/server/text_generation_server/layers/marlin/fp8.py
@@ -38,9 +38,12 @@ class GPTQMarlinFP8Linear(nn.Module):
 
         log_once(logger.info, "GPU does not support FP8, using Marlin FP8 kernel")
 
+        # if scales is a scalar (0D tensor), convert it to a 1D tensor
+        if scales.dim() == 0:
+            scales = scales.unsqueeze(0)
+
         scales = scales.unsqueeze(0)
-        # repack weights for Marlin if a single scale is provided
-        if scales.size(0) == 1:
+        if scales.shape[1] == 1:
             out_features, in_features = qweight.shape
             scales = scales.repeat(1, out_features)
         qweight, scales = repack_fp8_for_marlin(qweight, scales)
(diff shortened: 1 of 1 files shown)


server/text_generation_server/layers/marlin/fp8.py
```diff
@@ -39,7 +39,8 @@ def __init__(
         log_once(logger.info, "GPU does not support FP8, using Marlin FP8 kernel")
 
         scales = scales.unsqueeze(0)
-        if scales.shape[1] == 1:
+        # repack weights for Marlin if a single scale is provided
+        if scales.size(0) == 1:
```
Narsil, line 43:
```suggestion
        if scales.shape[0] == 1:
```

Can you explain where this change is coming from ?

drbh: yes, apologies for not including an example above. Currently if attempting to quantize an unquantized model with marlin fp8 the line above will throw when attempting to use marlin to repack.

```
text-generation-launcher --model-id meta-llama/Meta-Llama-3-8B --quantize fp8
```

with the change above this model loads and generates as expected

```bash
curl 127.0.0.1:3000/generate -X POST -d '{"inputs":"What is Deep Learning?","parameters":{"max_new_tokens":20}}' -H 'Content-Type: application/json'
# {"generated_text":" Deep learning is a subset of machine learning that is inspired by the structure and function of the human brain"}
```

Narsil:
I'm confused this still doesn't fix neural magic.

```
text-generation-launcher --model-id meta-llama/Meta-Llama-3-8B --quantize fp8
```
is currently working on `main`. This might have been fixed by something else  ?

Can we introduce a failing test before fixing this ?

Narsil:
Closing as stale, feel free to reopen.

PR #2419, opened 2024-08-14
Fixing exl2 and other quanize tests again.

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


launcher/src/main.rs
```diff
@@ -72,17 +82,17 @@ enum Quantization {
...
     /// Bitsandbytes 8bit. Can be applied on any model, will cut the memory requirement in half,
     /// but it is known that the model will be much slower to run than the native f16.
-    #[deprecated(
-        since = "1.1.0",
-        note = "Use `eetq` instead, which provides better latencies overall and is drop-in in most cases"
-    )]
+    // #[deprecated(
+    //     since = "1.1.0",
+    //     note = "Use `eetq` instead, which provides better latencies overall and is drop-in in most cases"
+    // )]
```
Narsil, line 88: We cna't keep that easily because `Deserialize` macro will use that field, and there's  no way for us to insert `#[allow(deprecated)]` in the generated code.

launcher/src/main.rs
```diff
@@ -72,17 +82,17 @@ enum Quantization {
...
     Bitsandbytes,
     /// Bitsandbytes 4bit. Can be applied on any model, will cut the memory requirement by 4x,
     /// but it is known that the model will be much slower to run than the native f16.
-    BitsandbytesNF4,
+    BitsandbytesNf4,
```
Narsil, line 92: This is the same for `ValueEnum` and make `serde` values transparent with `kebab-case`.

launcher/src/main.rs
```diff
@@ -1544,20 +1557,26 @@ fn main() -> Result<(), LauncherError> {
...
-                | Quantization::BitsandbytesFP4,
+                | Quantization::BitsandbytesNf4
+                | Quantization::BitsandbytesFp4,
             ),
         ) => {
             tracing::info!("Bitsandbytes doesn't work with cuda graphs, deactivating them");
             vec![]
         }
+        (None, Some(Quantization::Exl2)) => {
+            tracing::info!("Exl2 doesn't work with cuda graphs, deactivating them");
```
Narsil, line 1577: New inserted logic, we remove cuda graphs when the model is exl2.

server/pyproject.toml
```diff
@@ -46,6 +46,7 @@ marlin-kernels = [
   { url = "https://github.com/danieldk/marlin-kernels/releases/download/v0.2.0/marlin_kernels-0.2.0+cu123torch2.4-cp311-cp311-linux_x86_64.whl", python = "~3.11", optional = true },
   { url = "https://github.com/danieldk/marlin-kernels/releases/download/v0.2.0/marlin_kernels-0.2.0+cu123torch2.4-cp312-cp312-linux_x86_64.whl", python = "~3.12", optional = true },
 ]
+rich = "^13.7.1"
```
Narsil, line 49: This is a new dep of exl2 (which we don't care that much about, but makes evolving versions easier since we don't have to maintain a fork).

launcher/src/main.rs
```diff
@@ -1544,20 +1557,25 @@ fn main() -> Result<(), LauncherError> {
...
```
OlivierDehaene, line 1576:
```suggestion
            tracing::warn!("Exl2 doesn't work with cuda graphs, deactivating them");
```

launcher/src/main.rs
OlivierDehaene: Would it be possible to add the deprecation in the Display impl instead? I think only Clap uses Display.

Issue #2369, opened 2024-08-07
Add /v1/models API endpoint to be compatible with OpenAI APIs


ktrapeznikov:
### Feature request

Add `/v1/models/` endpoint to list model information. This will basically same info as `/info` endpoint just compliant with OpenAI APIs

### Motivation

Lot of other tools expect `/v1/models` to list the available models.

### Your contribution

 I can try submitting a PR

ErikKaum:
Hi @ktrapeznikov 👋 

To me this makes sense. If you have time making a PR please feel free to go ahead 🙌

Simon-Stone:
This would be very helpful to my use case, too!

ErikKaum:
Awesome 🙌 we're quite constrained on bandwidth at the moment. But once we get to this it should be a quick thing to implement.

Also, if you have the time please feel free to submit a PR 👍

PR #2433, opened 2024-08-19
feat: add /v1/models endpoint

drbh:
This PR adds a new endpoitns `/v1/models` that returns data similar to `/info`. This aims to align with openai's models endpoint, however in the TGI case we'll always return a single model

```bash
curl 127.0.0.1:3000/v1/models 
```

```json
{
    "object": "list",
    "data": [
        {
            "id": "meta-llama/Meta-Llama-3-8B ",
            "object": "model",
            "created": 0,
            "owned_by": "meta-llama/Meta-Llama-3-8B "
        }
    ]
}
```

https://platform.openai.com/docs/api-reference/models/list

resolves: https://github.com/huggingface/text-generation-inference/issues/2369


Narsil (approved):
LGTM

drbh:
Tagging @Wauplin for visibility since this adds a new endpoint/docs

PR #2449, opened 2024-08-22
Only 1 allocator2

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


backends/v3/src/backend.rs
```diff
@@ -35,27 +35,15 @@ impl BackendV3 {
...
-            Attention::Paged
-        };
-        let block_size = if attention == Attention::FlashDecoding {
-            256
-        } else if attention == Attention::FlashInfer {
-            1
-        } else {
-            16
-        };
+        let prefix_caching =
```
OlivierDehaene, line 38: In the future, this could be done by calling `Info` from the gRPC client.

backends/v3/src/radix.rs
```diff
@@ -16,26 +14,35 @@ pub struct RadixAllocator {
...
-            unimplemented!("Window size not supported in the prefix-caching block allocator yet");
-        }
+        // assert_eq!(
+        //     block_size, 1,
+        //     "Radix tree allocator only works with block_size=1, was: {}",
+        //     block_size
+        // );
+        // if window_size.is_some() {
+        //     unimplemented!("Window size not supported in the prefix-caching block allocator yet");
+        // }
```
OlivierDehaene, line 35:
```suggestion
```

launcher/src/main.rs
```diff
@@ -18,30 +18,131 @@ use std::sync::{mpsc, Arc};
...
+use std::{
+    fs, io,
+    io::{Read, Write},
+};
 use thiserror::Error;
 use tracing_subscriber::{filter::LevelFilter, EnvFilter};
 
 mod env_runtime;
 
+fn get_config(
```
OlivierDehaene, line 30: Same, this could be done through `Info` in the future.

router/src/lib.rs
```diff
@@ -1072,6 +1082,16 @@ pub(crate) struct GenerateRequest {
     pub inputs: String,
     #[serde(default = "default_parameters")]
     pub parameters: GenerateParameters,
+
+    /// This is used internally because some requests
+    /// already contain the templated input therefore
+    /// we shouldn't add the special tokens.
+    #[serde(default = "default_true", skip)]
+    pub add_special_tokens: bool,
```
OlivierDehaene, line 1090: Shouldn't we add this to the internal `tokenize`, `infer`... router functions instead?

server/tests/conftest.py
```diff
@@ -1,7 +1,10 @@
 import pytest
-
+import os
 from text_generation_server.pb import generate_pb2
 
+os.environ["USE_PREFIX_CACHING"] = "1"
+os.environ["ATTENTION"] = "flashinfer"
```
OlivierDehaene, line 6: ?

server/text_generation_server/layers/attention/common.py
```diff
@@ -9,26 +9,48 @@
...
+                    device=device,
+                    dtype=torch.int32,
+                )
+                max_q = 1
+            else:
+                assert max_q is not None
+            assert max_k is not None
             cu_seqlen_k = torch.zeros(shape[-1] + 1, device=device, dtype=torch.int32)
+            # if max_q is not None and max_q < 1000 and max_q > 1:
+            #     import ipdb;ipdb.set_trace()
```
OlivierDehaene, line 42:
```suggestion
```

server/text_generation_server/layers/attention/common.py
```diff
@@ -9,26 +9,48 @@
...
             self.input_lengths = input_lengths
+            self.prefix_lengths = prefix_lengths
             device = self.input_lengths.device
             shape = self.input_lengths.shape
-            cu_seqlen_q = torch.arange(
-                shape[0] + 1,
-                device=device,
-                dtype=torch.int32,
-            )
+            if cu_seqlen_q is None:
```
OlivierDehaene, line 30: When can it be None, and shouldn't it take into account speculative tokens into the arange?

server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -257,22 +262,15 @@ def from_tokenized(
...
-
             orig_input_length = len(tokenized_input)
 
-            if PREFIX_CACHING:
-                prefix_len = r.prefix_len
-                if prefix_len == orig_input_length:
-                    assert prefix_len > 0
-                    prefix_len -= 1
-            else:
-                prefix_len = 0
```
OlivierDehaene, line 275: Do we correctly default to 0 in the router?

router/src/lib.rs
Narsil: This was way faster to do it this way (as we're passing GenerateRequest everywhere).
Using `[serde(skip)]` means it's not settable through public API nor is it visible with utoipa.

My first attempt I used a separate variable and found it was just adding lots of noise for nothing (we *need* that information on *every* generateRequest).

server/tests/conftest.py
Narsil: This is for the server tests, those are run in isolation (without the router/launcher).

Therefore we need those variables to be set (those variables are enforced to exist in the python code, in order to minize issues in the resolution of their values.

server/text_generation_server/layers/attention/common.py
Narsil: This is already done beforehand currently.

Althrough this should be fixed once speculative is correctly implemented with `cu_seqlen_q` and `cu_speculated_ids` or somethign (to get the offsets of the input_ids that are speculated for instance).

server/text_generation_server/models/flash_causal_lm.py
Narsil: Yes, we do not however correctly account for leaving 1 token spare for the prefill to actually occur normally.

launcher/src/main.rs
Narsil: I think I prefer doing it super early on, so that dependencies between flags can be determined early, and the regular code can just use 1 strong value.

But I see the point.

OlivierDehaene (approved):

PR #2455, opened 2024-08-25
Fix: don't apply post layernorm in SiglipVisionTransformer

tgaddair:
This fixes a bug with LLaVA Next when using Siglip as the vision model. LLaVA Next expects the output of the vision model to be the encoder outputs before layernorm (see original transformers implementation here: https://github.com/huggingface/transformers/blob/main/src/transformers/models/llava_next/modeling_llava_next.py#L813).

This also makes Siglip consistent with the existing Clip implementation:

https://github.com/huggingface/text-generation-inference/blob/main/server/text_generation_server/models/custom_modeling/clip.py#L613


drbh:
closing as https://github.com/huggingface/text-generation-inference/pull/2459 was merged with these commits included. Thanks again!

Issue #2460, opened 2024-08-27
Error With Function calling & Messages API post Chat Template FIx 


arshabdi:
### System Info

Hey all,

Sharing a issue im facing as off the new merge a couple of hours ago in regards to function calling & the messages API. 

Using the latest docker img w/config (Llama 3.1 8b Instruct);

`   sudo docker run --gpus all --shm-size 64g -e HF_TOKEN=$token -e HF_HUB_VERBOSITY=debug -p 8080:80 -v $volume:/data \
    ghcr.io/huggingface/text-generation-inference:sha-cfa73b5 \
    --model-id $model --quantize bitsandbytes-fp4 `

when i try to run the same test code that i used with the previous build i now receive the following error on the client side: 

`openai.UnprocessableEntityError: Error code: 422 - {'error': 'Template error: invalid operation: cannot convert map to bool (in <string>:60)', 'error_type': 'template_error'}`

from within my docker container i can see the follow error aswell:
`2024-08-27T03:31:26.384967Z ERROR chat_completions:apply_chat_template: text_generation_router::infer: router/src/infer/mod.rs:151: Template error: invalid operation: cannot convert map to bool (in <string>:60)`

showing that the most recent changes (specifically in router/src/infer/mod.rs) has caused some sort of issue with tool use.

below is a copy of the client-side code i am using to test:

`from openai import OpenAI
import openai
from dotenv import load_dotenv
load_dotenv()
tools = [
        {
            "type": "function",
            "function": {
                "name": "get_current_weather",
                "description": "Get the current weather",
                "parameters": {
                    "type": "object",
                    "properties": {
                        "location": {
                            "type": "string",
                            "description": "The city and state, e.g. San Francisco, CA"
                        },
                        "format": {
                            "type": "string",
                            "enum": ["celsius", "fahrenheit"],
                            "description": "The temperature unit to use. Infer this from the users location."
                        }
                    },
                    "required": ["location", "format"]
                }
            }
        }
]

chat_completion = client.chat.completions.create(
    model="tgi",
    messages = [
        {"role": "user", "content": "What is the weather like in New York?"}
    ],
    tools=tools, 
    tool_choice="auto", 
)

for message in chat_completion:
    print(message)`

Once again would like to mention that i was not having this problem prior to the most recent merge. 

[...]

PR #2463, opened 2024-08-27
fix: bump minijinja version and add test for llama 3.1 tools

drbh:
This PR adds `tojson` support into chat templates, tests for meta-llama/Meta-Llama-3.1-8B-Instruct's template and some logic to avoid an indexing issue with messages. 

Resolves: https://github.com/huggingface/text-generation-inference/issues/2460

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2169,9 +2169,9 @@ dependencies = [
 
 [[package]]
 name = "minijinja"
-version = "2.1.0"
+version = "2.2.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "45f7e8e35b6c7b169bf40b0176d2c79291ab8ee53290b84e0668ab21d841aa9d"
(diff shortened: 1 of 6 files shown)


arshabdi:
Thanks @drbh youre the man! Hoping to have this merged and testing soon ;)

drbh:
merging as the failing test in CI is unrelated to these changes (`test_flash_llama_exl2_load` only calls `generate` and never uses a template or interacts with minijinja) additional this PR does not contain any logic changes and only bumps minijinja and adds a new test.

drbh:
@arshabdi thank you for reporting the issue and the kind words 🙏

It appears that to support `meta-llama/Meta-Llama-3.1-8B-Instruct` we just needed to bump minijinja. minijinja's newer version adds support for `tojson` and `set messages = messages[1:]` that are used within the template. I've added a test for the specific template and tested generations locally. Please let me know if you still have any issues!

PR #2471, opened 2024-08-29
Small fixes for supported models

osanseviero:
* Remove supported hardware from titles
* Link to non-core model serving doc + remove unrelated


docs/source/supported_models.md
```diff
@@ -36,17 +34,4 @@ Text Generation Inference enables serving optimized models on specific hardware
...
-# or, for text-to-text generation models
-AutoModelForSeq2SeqLM.from_pretrained(<model>, device_map="auto")
-```
-
-If you wish to serve a supported model that already exists on a local folder, just point to the local folder.
-
-```bash
-text-generation-launcher --model-id <PATH-TO-LOCAL-BLOOM>
-```
+If the above list lacks the model you would like to serve, depending on the model's pipeline type, you can try to initialize and serve the model anyways to see how well it performs, but performance isn't guaranteed for non-optimized models. Read more about [Non-core Model Serving](../basic_tutorials/non_core_models).
```
Vaibhavs10, line 37:
```suggestion
If the model you would like to serve is not on the above list, depending on the model's pipeline type, you can try to initialize and serve the model anyway to see how well it performs, but performance isn't guaranteed for non-optimized models. 

```python
 # for causal LMs/text-generation models
 AutoModelForCausalLM.from_pretrained(<model>, device_map="auto")`
 # or, for text-to-text generation models
 AutoModelForSeq2SeqLM.from_pretrained(<model>, device_map="auto")
 ```

 If you wish to serve a supported model that already exists on a local folder, just point to the local folder.

 ```bash
 text-generation-launcher --model-id <PATH-TO-LOCAL-BLOOM>
 ```

Read more about [Non-core Model Serving](../basic_tutorials/non_core_models).
```

I think keeping both would make more sense.

docs/source/_toctree.yml
```diff
@@ -16,7 +16,7 @@
   - local: installation
     title: Installation from source
   - local: supported_models
-    title: Supported Models and Hardware
+    title: Supported Models
```
Vaibhavs10, line 19: We also wanted to move this up in the section.

Vaibhavs10 (commented):
Thanks for the updates, left a minor comment.

docs/source/supported_models.md
Vaibhavs10: Ah GH messed up the suggestion.

docs/source/_toctree.yml
osanseviero: WDYT?

docs/source/supported_models.md
osanseviero: I was removing the python snippet in the PR. Do you want to re-add it?

docs/source/_toctree.yml
Vaibhavs10: Yes! just after Quick tour is lovely! ty!

docs/source/supported_models.md
Vaibhavs10: Yeah - because it ties up nicely with the message that even if the model is not supported you can still use it with TGI, here's a small snippet to test and you can go read it more about it here.

(I don't feel too strongly about it, so feel free to call the shot as you deem fit)

Vaibhavs10 (commented):
Oh also, can you run `python update_doc.py` that'll fix the failing test.

You might need to run `npm i -g @redocly/cli@latest` before hand.

(I don't have approve rights on this repo, but everything else LGTM.)

docs/source/supported_models.md
osanseviero: What about the Python snippet?

Vaibhavs10: I think it'd be good to have the Python snippet. but don't feel too strongly. your call! 🤗

Narsil (approved):

PR #2482, opened 2024-08-30
feat: support lora revisions and qkv_proj weights

drbh:
This PR resolves two issues with lora. This PR adds support for `@revisions` on lora adapters. This will fetch or use lora adapters with a specific revision/branch/commit. Additionally this PR adds lora support for `qkv_proj` weights, which are used in the phi3 model. Lastly this PR has some other small refactors and improved error messages for lora


drbh:
merging as these changes hotfix and issue with loading loras on phi3

PR #2498, opened 2024-09-06
Fixing more correctly the invalid drop of the batch.

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/backends/v3/src/queue.rs
+++ b/backends/v3/src/queue.rs
@@ -252,17 +252,14 @@ impl State {
         let next_batch_span = info_span!(parent: None, "batch", batch_size = tracing::field::Empty);
         next_batch_span.follows_from(Span::current());
 
-        let mut batch_requests = Vec::with_capacity(self.entries.len());
-        let mut batch_entries =
-            IntMap::with_capacity_and_hasher(self.entries.len(), BuildNoHashHasher::default());
-
+        let mut batch = Vec::with_capacity(self.entries.len());
         let mut max_input_length = 0;
         let mut prefill_tokens: u32 = 0;
         let mut decode_tokens: u32 = 0;
         let mut max_blocks = 0;
 
         // Pop entries starting from the front of the queue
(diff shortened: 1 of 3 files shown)


backends/v3/src/radix.rs
```diff
@@ -89,6 +89,8 @@ impl Allocator for RadixAllocator {
 
         let suffix_blocks = (suffix_len + self.block_size - 1) / self.block_size;
 
+        tracing::info!("Prefix {prefix_len} - Suffix {suffix_len}");
```
danieldk, line 92: Probably not needed anymore?

danieldk (approved):
Looks good, though one log I don't think we need anymore?

backends/v3/src/queue.rs
```diff
@@ -324,23 +321,59 @@ impl State {
...
+        for (id, mut entry, block_allocation) in batch {
+            let block_allocation = if let (Some((tokens, input_ids)), Some(block_allocator)) =
+                (block_allocation, &self.block_allocator)
+            {
+                match block_allocator.allocate(tokens, input_ids).await {
+                    None => {
+                        // Entry is over budget
+                        // Add it back to the front
+                        tracing::debug!("Over budget: not enough free blocks");
+                        self.entries.push_front((id, entry));
```
drbh, line 365: not sure if this is possible, but if the first entry always fails (maybe size or resource related) is it possible this will loop forever since it keeps pushing the value to the front and attempted in the next batch? I guess in that case we'd always report a over budget warning which seems like a reasonable response

Narsil: If that happens we break (next line).
So we escape this and simply return an empty list, no ?

drbh: ah yea I likely misread. My understanding is that `self.entries[0]` would now be the entry that failed to allocated in the last batch. If for some read that specific entry was causing the failure I think it would be retried on the next loop and could continue that cycle

PR #2500, opened 2024-09-06
Add links to Adyen blogpost

martinigoyanes:
I feel this setup makes more sense since the blogpost was not really talking about Streaming or any of the concepts you already have inside the "Conceptual Guides". Instead, it gives more of a holistic view of the architecture and the interplay between server and router. 

I created this new section "External" for cases like this, ideally in the future the list of articles in "External" will grow as TGI adoption grows.

Eager to hear your feedback @Narsil


Narsil (dismissed):
LGTM, thansk for this contribution !

Narsil (approved):

PR #2516, opened 2024-09-12
Adding a test for FD.

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


backends/v3/src/queue.rs
```diff
@@ -364,7 +364,7 @@ impl State {
                         // Add it back to the front
                         tracing::debug!("Over budget: not enough free blocks");
                         self.entries.push_front((id, entry));
-                        break;
+                        continue;
```
Narsil, line 367: Important so we either allocate the rest or push them back into entries (otherwise they get discarded without being run)

backends/v3/src/queue.rs
```diff
@@ -436,6 +436,12 @@ impl State {
             batch_entries.insert(id, entry);
         }
 
+        // Empty batch
+        if batch_requests.is_empty() {
+            tracing::debug!("Filterered out all entries");
+            return None;
```
Narsil, line 442: This is important because sending empty prefill is not really nice.

backends/v3/src/radix.rs
```diff
@@ -1,10 +1,17 @@
 use crate::block_allocator::{Allocator, BlockAllocation};
 use slotmap::{DefaultKey, SlotMap};
+use std::hash::{Hash, Hasher};
 use std::{
     collections::{BTreeSet, HashMap},
     sync::Arc,
 };
 
+fn hash_(slice: &[u32]) -> u32 {
```
Narsil, line 9: The crux of the change. Instead of matching on a single token (input_ids) we match on the entire block content.

backends/v3/src/radix.rs
```diff
@@ -1,10 +1,17 @@
 use crate::block_allocator::{Allocator, BlockAllocation};
 use slotmap::{DefaultKey, SlotMap};
+use std::hash::{Hash, Hasher};
 use std::{
     collections::{BTreeSet, HashMap},
     sync::Arc,
 };
 
+fn hash_(slice: &[u32]) -> u64 {
```
Narsil, line 9: The crux of the change. Instead of matching on a single token (input_ids) we match on the entire block content.

danieldk, line 9: For slice length 1, we can return `slice[0] as u64`. Maybe also an assertion that the slice length is at least 1?

Narsil: Done

backends/v3/src/queue.rs
OlivierDehaene: I missed the PR where this was moved.
Since we check the prefill token budget early, if all entries in `batch`end up being in the radix tree, can't we end up running at very low prefill utilization then? 
If I understand all the components correctly, the block allocator is the only one to truly know the length in tokens that each request will have.

flake.lock
```diff
@@ -978,11 +978,11 @@
...
-        "narHash": "sha256-nJHA1SvIQbXySpL2ueNbzQOhnkQASa5tOLz/kdW0PWA=",
+        "lastModified": 1726229792,
+        "narHash": "sha256-9xsLmjc9nr7a4PTddKv2DOi82ompTtJNyjO6R67y5tE=",
         "owner": "danieldk",
         "repo": "tgi-nix",
-        "rev": "d40f3c22e9bcc5e16c94d4605cf6a7d74dd07f46",
+        "rev": "1a902f4818e94c3f8d95f6000db17bc3fadd0ce7",
         "type": "github"
       },
       "original": {
```
danieldk, line 988: Fine with updating, but the lockfile changes seem unrelated?

danieldk (approved):
The radix tree changes look good to me!

flake.lock
Narsil: I just did `nix flake update` at some point.

Any issue with that?

PR #2527, opened 2024-09-17
We can have a tokenizer anywhere.

Narsil:
# What does this PR do?

We can check that it works with `google/byt5-small` for instance (And also deadlocks if multiple validation workers are called since they all fight for the GIL as I've owned the gil in each thread.)

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


router/src/server.rs
```diff
@@ -208,40 +207,30 @@ async fn get_chat_tokenize(
...
-                    text,
-                    start,
-                    stop,
-                }
-            })
-            .collect();
+    let tokens: Vec<SimpleToken> = encoding
+        .get_ids()
+        .iter()
+        .zip(encoding.get_offsets())
```
OlivierDehaene, line 213: This zip will fail for Python.

router/src/server.rs
```diff
@@ -1560,35 +1549,25 @@ async fn tokenize(
...
-            Json(ErrorResponse {
-                error: "No fast tokenizer or tokenizer.json for this model".to_string(),
-                error_type: "no fast tokenizer".to_string(),
-            }),
-        ))
-    }
+    let tokens: Vec<SimpleToken> = encoding
+        .get_ids()
+        .iter()
+        .zip(encoding.get_offsets())
```
OlivierDehaene, line 1555: Same.

OlivierDehaene (commented):
Very nice!

router/src/server.rs
Narsil: I "fixed" it by ignoring offsets if not present.

Python non rust will definitely exist in forms where offsets are not available (ByT5) so I'll assume we cannot have them.
In that case we'll send the raw ids and totally ignore the "text" part (since it doesn't have a good definition without offsets in the original query)

OlivierDehaene (approved):

server/text_generation_server/models/custom_modeling/mamba_modeling.py
```diff
@@ -196,7 +196,10 @@ class MambaModel(nn.Module):
     def __init__(self, config, weights):
         super().__init__()
         prefix = "backbone"
-        self.embed_tokens = TensorParallelEmbedding(f"{prefix}.embedding", weights)
+        try:
+            self.embed_tokens = TensorParallelEmbedding(f"{prefix}.embeddings", weights)
+        except RuntimeError:
+            self.embed_tokens = TensorParallelEmbedding(f"{prefix}.embedding", weights)
```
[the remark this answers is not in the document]
Narsil: No `s`. There's a difference between mamba-hf and mamba ssh here.

PR #2510, opened 2024-09-11
fix: pass missing revision arg for lora adapter when loading multiple…

drbh:
This PR correctly includes the missing `adapter.revision` param in `load_module_map` which should resolve a bug when loading multiple lora adapters and at least on has a revision

PR #2529, opened 2024-09-17
bump hf hub to 0.25.0

nbroad1881:
a customer wants to use lora adapters with revision syntax which is only possible in the latest hf hub client.

e.g. Name/model-name@revision

@Narsil


nbroad1881:
Maybe related to https://github.com/huggingface/text-generation-inference/pull/2510

PR #2552, opened 2024-09-23
Fix `/vertex` payload parsing when `MESSAGES_API_ENABLED`

alvarobartt:
# What does this PR do?

This PR fixes the input payload parsing on Vertex AI i.e. when compiled with the `google` feature and deployed in a Vertex AI environment (replicable with the following environment variables `AIP_MODE=PREDICTION`, `AIP_HTTP_PORT=80`, `AIP_PREDICT_ROUTE=/predict`, and `AIP_HEALTH_ROUTE=/health`); when the environment variable `MESSAGES_API_ENABLED` is set to true, meaning that the `/vertex` endpoint is a proxy for `/v1/chat/completions` endpoint instead of the default `/generate`.

> [!NOTE]
> As already discussed with @Narsil this is most likely not the best approach, here are some things that I tried:
> - Implementing the From trait to go from the input Vertex AI payload into `ChatRequest`, but that still requires us to have a "duplicated" struct holding the generation kwargs, which is not ideal
> - The above could be solved by moving the generation kwargs out of `ChatRequest` whilst just leaving `messages` there, and using `#[serde(flatten)]` over the generation kwargs; but that's not ideal either
> Any feedback is appreciated!

## Who can review?

@Narsil as of our previous conversation today, feel free to take over the PR, or even close it in favour of a potentially better PR! Thanks in advance Nicolas! 🤗


alvarobartt:
Closed in favour of #2553, thank you @Narsil!

PR #2566, opened 2024-09-25
Fix build with `--features google`

alvarobartt:
# What does this PR do?

As of recent changes within the Vertex AI integration via the `google` feature, the build via `cargo build --features google` was not working fine; this PR fixes that.

## Who can review?

@Narsil


Narsil:
Can we actually make sure this doesn't happen anymore ?

By adding `cargo test --features google` in a new CI step in `.github/workflows/tests.yaml` For instance ?
(We should also probably add this in the nix tests but I'll take care of it)

PR #2567, opened 2024-09-25
Add LoRA adapters support for Gemma2

alvarobartt:
# What does this PR do?

This PR adds support for LoRA adapters for Gemma2.

It has been tested with both `text-generation-launcher` and from a pre-built container.

```bash
export HF_TOKEN=$(cat ~/.cache/huggingface/token)
text-generation-launcher \
    --model-id=google/gemma-2-9b-it \
    --lora-adapters=google-cloud-partnership/gemma-2-9b-it-lora-magicoder \
    --num-shard=4 \
    --max-input-length=4000 \
    --max-total-tokens=4096
```

And requests sent via cURL as:

```bash
curl localhost:8080/v1/chat/completions \
    -X POST \
    -d '{"messages": [{"role": "user", "content": "Write a function to combine a base path and relative path into an absolute path."}], "max_tokens": 256, "model": "google-cloud-partnership/gemma-2-9b-it-lora-magicoder"}}'
```

cc'ing @Jofthomas for visibility

## Who can review?

@Narsil

### DIFF
--- a/server/text_generation_server/models/custom_modeling/flash_gemma2_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_gemma2_modeling.py
@@ -163,7 +163,9 @@ def _load_gqa(config, prefix: str, weights):
 
 
 class FlashGemma2Attention(torch.nn.Module):
-    def __init__(self, prefix: str, config, weights, layer_id, causal: bool, is_sliding: bool):
+    def __init__(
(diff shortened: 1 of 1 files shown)


Narsil (dismissed):
LGTM.

We really need to abstract away that thing.

Narsil:
Unrelated failures.

PR #2575, opened 2024-09-26
 Improve support for GPUs with capability < 8

danieldk:
# What does this PR do?

Improve support for GPUs with capability < 8

- For models that cannot use flashinfer, use flash-attn v1 + paged attention for models with a compute capability older than 8.
- Disable prefix caching when using paged attention.
- When using flash-attn v1, pass the key/value, rather than the cache, since v1 cannot use block tables.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/launcher/src/gpu.rs
+++ b/launcher/src/gpu.rs
@@ -1,9 +1,9 @@
 use std::sync::LazyLock;
 
-pub static COMPUTE_CAPABILITY: LazyLock<Option<(isize, isize)>> =
+pub static COMPUTE_CAPABILITY: LazyLock<Option<(usize, usize)>> =
     LazyLock::new(get_cuda_capability);
 
-fn get_cuda_capability() -> Option<(isize, isize)> {
+fn get_cuda_capability() -> Option<(usize, usize)> {
     use pyo3::prelude::*;
 
     let py_get_capability = |py: Python| -> PyResult<(isize, isize)> {
@@ -13,7 +13,11 @@ fn get_cuda_capability() -> Option<(isize, isize)> {
     };
 
     match pyo3::Python::with_gil(py_get_capability) {
-        Ok(capability) => Some(capability),
+        Ok((major, minor)) if major < 0 || minor < 0 => {
+            tracing::warn!("Ignoring negative GPU compute capabilities: {major}.{minor}");
+            None
+        }
+        Ok((major, minor)) => Some((major as usize, minor as usize)),
         Err(err) => {
             tracing::warn!("Cannot determine GPU compute capability: {}", err);
             None
--- a/launcher/src/gpu.rs
+++ b/launcher/src/gpu.rs
@@ -1,0 +1,22 @@
+use std::sync::LazyLock;
+
+pub static COMPUTE_CAPABILITY: LazyLock<Option<(isize, isize)>> =
+    LazyLock::new(get_cuda_capability);
+
+fn get_cuda_capability() -> Option<(isize, isize)> {
+    use pyo3::prelude::*;
+
+    let py_get_capability = |py: Python| -> PyResult<(isize, isize)> {
+        let torch = py.import_bound("torch.cuda")?;
+        let get_device_capability = torch.getattr("get_device_capability")?;
+        get_device_capability.call0()?.extract()
+    };
+
+    match pyo3::Python::with_gil(py_get_capability) {
+        Ok(capability) => Some(capability),
+        Err(err) => {
+            tracing::warn!("Cannot determine GPU compute capability: {}", err);
+            None
+        }
+    }
+}
--- a/launcher/src/main.rs
+++ b/launcher/src/main.rs
@@ -103,6 +103,10 @@ fn resolve_attention(config: &Option<Config>, lora_adapters: &Option<String>) ->
                             );
                             attention = Some(fallback_attention.to_string());
                         }
+                        if fallback_attention == "paged" && prefix_caching.is_none() {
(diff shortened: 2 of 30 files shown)


launcher/src/main.rs
```diff
@@ -65,6 +66,7 @@ fn get_config(
 }
 
 fn resolve_attention(config: &Option<Config>, lora_adapters: &Option<String>) -> (String, String) {
+    let compute_capability = *gpu::COMPUTE_CAPABILITY;
```
Narsil, line 69: Why not just `get_cuda_capability()` ?

launcher/src/gpu.rs
```diff
@@ -0,0 +1,22 @@
+use std::sync::LazyLock;
+
+pub static COMPUTE_CAPABILITY: LazyLock<Option<(isize, isize)>> =
+    LazyLock::new(get_cuda_capability);
+
+fn get_cuda_capability() -> Option<(isize, isize)> {
+    use pyo3::prelude::*;
+
+    let py_get_capability = |py: Python| -> PyResult<(isize, isize)> {
```
Narsil, line 9: Really `isize` I'm curious why not `usize`.

launcher/src/main.rs
Narsil: No need to start a slow GIL if it's not needed.

launcher/src/main.rs
```diff
@@ -101,17 +110,26 @@ fn resolve_attention(config: &Option<Config>, lora_adapters: &Option<String>) ->
...
     let attention = attention.unwrap_or("flashinfer".to_string());
+    let prefix_caching = if attention == "paged"
+        && prefix_caching.is_none()
+        && compute_capability.is_some()
+    {
+        tracing::info!("Disabling prefix caching because it is not supported with 'flashinfer'");
+        "false".to_string()
+    } else {
+        prefix_caching.unwrap_or("true".to_string())
+    };
```
Narsil, line 131: Can't we move upstairs ? We're not allowed to change anything if the user sets it directly so this line must always be `unwraps_or`.
And in the previous code every modification must be protected to check that the thing is not None (there might even be a clearer method for it on `Option`.

server/text_generation_server/models/custom_modeling/flash_cohere_modeling.py
```diff
@@ -297,8 +298,8 @@ def forward(
             # flash attention
             attn_output = attention(
                 query,
-                kv_cache[0] if SYSTEM != "ipex" else key,
-                kv_cache[1] if SYSTEM != "ipex" else value,
+                kv_cache[0] if PREFILL_IN_KV_CACHE else key,
+                kv_cache[1] if PREFILL_IN_KV_CACHE else value,
```
Narsil, line 302: Another option would be to send both to every implementation and let V2/flashinfer use the cache directly with block tables, and let V1 and other backends use the raw values.

Let's keep this for now, but maybe food for thought if this logic complexifies.

launcher/src/main.rs
danieldk: I am not sure I understand. This will only evaluate the compute capability once, it's just a cheap Rust lock. If we'd call `get_cuda_capability`, we'd call into Python land every time.

launcher/src/gpu.rs
danieldk: I was mostly respecting the Python and CUDA types:

https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__DEVICE.html#group__CUDART__DEVICE_1g1bf9d625a931d657e08db2b4391170f0

server/text_generation_server/models/custom_modeling/flash_cohere_modeling.py
danieldk: Yeah, that crossed my mind as well, certainly worth considering in the future.

launcher/src/main.rs
danieldk: Do we need an `unwrap_or`? The first branch of the conditional only fires when `prefix_caching` is `None`, so the user has set something. Don't know why I added the `compute_capability.is_some()` though.

danieldk: I'll reread the preceding code to see if we check this everywhere.

Narsil: I don't know how to make it the most obvious tbh.

But what I was really aiming for was:

- User signal takes precedence over everything else.
- If not user specifed resolve attention (prefx + attention) as simply as possible:
  - When some conditions are met, we use non defaults.
  - If nothing else happens we use the defaults (so defaults are more easily seen imho)

I wasn't super happy with the design but that's the reason for the big ugly block on top.This is where I hide 'non defauls resolutions' 

So the spirit of my comment is just to try and keep the same rational, have the messy thing only in 1 place. (With the hidden agenda that this mess should disappear, whenever flashinfer supports all the features we would hope it does)

launcher/src/gpu.rs
Narsil: Fine by me, just feels very odd, I wouldn't know what to do with version -10. Having a parsing error there wouldn't shock me tbh.

danieldk: I'll add some validation and then convert it to usize.

danieldk: Added validation and changed types to `usize`.

launcher/src/main.rs
Narsil: Yes yes I understand, but since we're only calling once there's no need, but it's really totally not important.

Narsil (approved):
LGTM

PR #2578, opened 2024-09-26
feat: propagate max_concurrent_requests to queue state entries instead of hardcoded value in v2 and v3 backends

Venkat2811:
# What does this PR do?

`max_concurrent_requests` is currently used only in router's semaphore.  This change will propagate `max_concurrent_requests` (default val 128) to queue state entries instead of hardcoded value (128) in v2 and v3 backends.


Venkat2811:
@Narsil @OlivierDehaene pls review

Narsil:
@Venkat2811 I think you are confused about what `with_capacity` means.

`with_capacity` is simply a default max size for a vector which reduces the number of reallocations when a vector grows.
It has nothing to do with the concurrent requests size actually (the VecDeque will grow/reduce in size as necessary).

I'm going to close this for now. Happy to reopen if I missed something.

PR #2588, opened 2024-09-30
Fp8 e4m3_fnuz support for rocm

mht-sharma:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/layers/fp8.py
+++ b/server/text_generation_server/layers/fp8.py
@@ -1,7 +1,7 @@
 import torch
 
 from dataclasses import dataclass
-from typing import Optional, Union, List
+from typing import Optional, Tuple, Union, List
 from loguru import logger
 
 from text_generation_server.utils.import_utils import SYSTEM
@@ -51,8 +51,32 @@ def get_fp8_linear() -> torch.nn.Module:
     return Fp8Linear
 
 
+def normalize_e4m3fn_to_e4m3fnuz(
+    weight: torch.Tensor,
+    weight_scale: torch.Tensor,
+    input_scale: Optional[torch.Tensor] = None,
+) -> Tuple[torch.Tensor, torch.Tensor, Optional[torch.Tensor]]:
+    assert weight.dtype == torch.float8_e4m3fn
+    # The bits pattern 10000000(-128) represents zero in e4m3fn
+    # but NaN in e4m3fnuz. So here we set it to 0.
+    # https://onnx.ai/onnx/technical/float8.html
+    weight_as_int8 = weight.view(torch.int8)
+    ROCM_FP8_NAN_AS_INT = -128
+    weight_as_int8[weight_as_int8 == ROCM_FP8_NAN_AS_INT] = 0
+    weight = weight_as_int8.view(torch.float8_e4m3fnuz)
+
+    # For the same bits representation, e4m3fnuz value is half of
+    # the e4m3fn value, so we should double the scaling factor to
+    # get the same dequantized value.
+    # https://onnx.ai/onnx/technical/float8.html
+    weight_scale = weight_scale * 2.0
+    if input_scale is not None:
+        input_scale = input_scale * 2.0
+    return weight, weight_scale, input_scale
+
+
 def fp8_quantize(
-    weight, scale_upper_bound=None, qdtype=torch.float8_e4m3fn, scalar=False
+    weight, scale=None, scale_upper_bound=None, qdtype=torch.float8_e4m3fn, scalar=False
 ):
     if FBGEMM_DYN_AVAILABLE and not scalar:
         qweight, scale = torch.ops.fbgemm.quantize_fp8_per_row(
@@ -62,8 +86,11 @@ def fp8_quantize(
 
     # weight, scale = quant_weights(weight, torch.int8, False)
     finfo = torch.finfo(qdtype)
-    # Calculate the scale as dtype max divided by absmax
-    scale = finfo.max / weight.abs().max().clamp(min=1e-12, max=scale_upper_bound)
+
+    if scale is None:
+        # Calculate the scale as dtype max divided by absmax
+        scale = finfo.max / weight.abs().max().clamp(min=1e-12, max=scale_upper_bound)
+
     # scale and clamp the tensor to bring it to
     # the representative range of float8 data type
     # (as default cast is unsaturated)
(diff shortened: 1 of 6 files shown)


server/text_generation_server/layers/fp8.py
```diff
@@ -200,56 +272,89 @@ def get_linear(self, bias: torch.Tensor):
...
+            self.weight,
+            self.weight_scale,
+            self.input_scale,
+            self.activation_scale_ub,
+            bias,
+            self.dtype,
         )
 
 
 class Fp8Linear(torch.nn.Module):
```
mht-sharma, line 284: Would it be cleaner to have a separate Fp8LinearRocm?

server/text_generation_server/layers/fp8.py
```diff
@@ -92,9 +123,17 @@ def get_weights(self, weights: "Weights", prefix: str):
                 .reshape(-1)
                 .expand(w.shape[0])
             )
+            try:
+                input_scale = weights.get_tensor(
```
danieldk, line 127: `Weights` also has `_has_tensor` maybe we should make it public and use it here?

server/text_generation_server/layers/fp8.py
```diff
@@ -72,6 +99,10 @@ def fp8_quantize(
     # as both required as inputs to torch._scaled_mm
     qweight = qweight.to(qdtype)
     scale = scale.float().reciprocal()
+
+    if SYSTEM == "rocm":
+        qweight, scale, _ = normalize_e4m3fn_to_e4m3fnuz(qweight, scale)
```
danieldk, line 104: We should wire up `scale` at some point for CUDA as well.

server/text_generation_server/layers/fp8.py
danieldk: Same for `try: [...]get_tensor` below.

server/text_generation_server/layers/fp8.py
```diff
@@ -266,15 +371,49 @@ def forward(self, input: torch.Tensor) -> torch.Tensor:
...
+            output = torch._scaled_mm(
+                qinput,
+                self.qweight.t(),
+                out_dtype=self.dtype,
+                scale_a=scale,
+                scale_b=self.scale,
+                bias=self.bias,
+            )
+
+            if type(output) is tuple and len(output) == 2:
```
danieldk, line 394: Did this change between torch versions or is output for AMD different?

```suggestion
            if isinstance(output, tuple) and len(output) == 2:
```

server/text_generation_server/layers/fp8.py
danieldk: Maybe, it depends a bit on how much conditional code we end up with. We did separate FP8 Marlin for this reason.

server/text_generation_server/layers/marlin/fp8.py
```diff
@@ -62,7 +62,7 @@ def from_unquant(cls, weight, bias, dtype):
         return cls(qweight=qweight, scales=scales.to(dtype), bias=bias)
 
     @classmethod
-    def from_fp8(cls, weight, scale, _input_scale, bias, dtype):
+    def from_fp8(cls, weight, scale, _input_scale, _scale_upper_bound, bias, dtype):
```
danieldk, line 65: Type.

These arguments get a bit messy. It's easy to mix up a tensor or a float (which was already happening here?). Maybe we should switch these to kwargs-only so that the call sites need to be explicit (+ type annotations).

server/text_generation_server/layers/fp8.py
mht-sharma: Updated to use the `has_tensor`

server/text_generation_server/models/__init__.py
```diff
@@ -342,22 +342,19 @@ def get_model(
     model_type = config_dict.get("model_type", None)
 
     quantization_config = config_dict.get("quantization_config", None)
-    compression_config = config_dict.get("compression_config", None)
```
mht-sharma, line 345: @danieldk config renamed to quantisation config.

server/text_generation_server/layers/marlin/fp8.py
mht-sharma: Converted them to kwargs and added type hints.

server/text_generation_server/layers/fp8.py
mht-sharma: Done

server/text_generation_server/layers/fp8.py
mht-sharma: Updated to use the `has_tensor`

server/text_generation_server/layers/fp8.py
mht-sharma: This is a common change for torch 2.5. https://github.com/pytorch/pytorch/blob/main/aten/src/ATen/native/cuda/Blas.cpp#L1175

server/text_generation_server/layers/fp8.py
```diff
@@ -125,9 +164,24 @@ def get_weights_col_packed(
                 )
             scale = scale.reshape(-1).expand(w.shape[0])
 
+            input_scale = None
+            if weights.get_tensor(f"{prefix}.input_scale"):
```
danieldk, line 168:
```suggestion
            if weights.has_tensor(f"{prefix}.input_scale"):
```

?

server/text_generation_server/layers/fp8.py
```diff
@@ -154,9 +208,21 @@ def get_multi_weights_col(self, weights: "Weights", prefixes: List[str], dim: in
             ]
             scale = torch.cat(scale, dim=0).reshape(-1)
 
+            input_scale = [
+                _load_scalar_or_matrix_scale(weights, f"{p}.input_scale", shape)
+                for p, shape in zip(prefixes, shapes)
+                if weights.has_tensor(f"{p}.input_scale")
```
danieldk, line 214: Given this conditional, we probably need an assertion like

```
assert len(input_scale) == 0 or len(input_scale) == length(prefixes)
```

server/text_generation_server/models/__init__.py
danieldk: I think we should check for both keys, at least for the time being. Some customers/users may have checkpoints that still have `compression_config`. Maybe with a comment that `compression_config` is for backwards compatibility?

danieldk (approved):

PR #2597, opened 2024-10-01
Unroll notify error into generate response

drbh:
This PR unrolls the message from `notify_error` in order to give the model a way to not choose a tool and respond with a non constrained generation


Narsil (dismissed):
LGTM.

PR #2605, opened 2024-10-03
Revert "Unroll notify error into generate response"

drbh:
Reverts huggingface/text-generation-inference#2597 as this solution does not account for streaming response since it depends on parsing the completed generation


drbh:
unrolling logic will be superseeded by changes to overall tool implementation https://github.com/huggingface/text-generation-inference/pull/2606

PR #2609, opened 2024-10-04
Simplify the `attention` function

danieldk:
# What does this PR do?

- Use one definition rather than multiple (will make it easier to do shared things once, such as calculating the FP8 KV cache reciprocal).
- Add `key`/`value` arguments, so that we don't need the `PREFILL_IN_KV_CACHE` constant.
- Make it kwargs-only (to avoid mixing up the various `Tensor` args).

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/layers/attention/cuda.py
+++ b/server/text_generation_server/layers/attention/cuda.py
@@ -252,7 +252,9 @@ def attention(
             window_left=window_size_left,
         )
 
-    elif ATTENTION == "flashdecoding":
+    # If we are using flashdecoding or paged, we always use flash-attn for
+    # the prefill. We have to branch on whether we use flash-attn v1 or v2.
+    elif V2:
         out = torch.empty_like(query)
         if window_size_left <= 0 and window_size_left != -1:
             raise ValueError("`window_size_left` must be > 0 or -1")
@@ -262,14 +264,15 @@ def attention(
 
         return flash_attn_2_cuda.varlen_fwd(
             query,
-            kv_cache.key,
-            kv_cache.value,
+            # flashdecoding: pass the KV caches, paged: pass the KV.
+            kv_cache.key if ATTENTION == "flashdecoding" else key,
+            kv_cache.value if ATTENTION == "flashdecoding" else value,
             out,
             seqlen.cu_seqlen_q,
             seqlen.cu_seqlen_k,
             None,
             None,
-            block_tables,
+            block_tables if ATTENTION == "flashdecoding" else None,
             None,
             seqlen.max_q,
             seqlen.max_k,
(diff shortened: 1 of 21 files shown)


server/text_generation_server/layers/attention/rocm.py
```diff
@@ -227,29 +225,32 @@ def paged_attention(
...
-        block_tables: torch.Tensor,
-        softmax_scale: float,
-        window_size_left: int = -1,
-        causal: bool = True,
-        softcap: float = 0.0,
-    ):
+
+
+def attention(
+    *query,
```
Narsil, line 231: Isn't that supposed to be `*, query, ` ? 

Also it doesn't matter as much since we force kwargs, but key, value, key_cache, value_cache, somehow makes more sense to my brain (same tensor shapes, come from the same location).

Nit really

danieldk: > Isn't that supposed to be `*, query,`  ?

Indeed, fixed!

danieldk: > Also it doesn't matter as much since we force kwargs, but key, value, key_cache, value_cache, somehow makes more sense to my brain (same tensor shapes, come from the same location).

I think now that the KV cache is a separate data type, we should just pass that in the future.

danieldk: Also done.

drbh (changes requested):
LGTM

- just the small nits about function signatures

drbh (dismissed):
nice! looks good to me ✨

server/text_generation_server/layers/attention/cuda.py
```diff
@@ -216,263 +216,130 @@ def paged_attention(
...
-                0,
-                None,
-            )
-            return out
+    elif ATTENTION == "flashdecoding":
+        out = torch.empty_like(query)
+        if window_size_left <= 0 and window_size_left != -1:
+            raise ValueError("`window_size_left` must be > 0 or -1")
 
-elif ATTENTION == "paged":
```
Narsil, line 356: You're breaking paged here.

ATTENTION="paged" text-generation-launcher ...

shows the issue.

PAGED still uses v2, not v1 (unless sm is too low)

Narsil: paged attention is not V1 vs V2, those are separate concerns.

Narsil: And we cannot use the block_tables implementation for paged + v2, because that requires BLOCK_SIZE=256, where paged attention uses block_size = 16.

danieldk: Should be fixed now, tested Llama & Mistral with `paged`, `flashattention` and `flashinfer`.

Narsil (approved):
LGTM

Narsil (approved):

PR #2614, opened 2024-10-07
feat: allow tool calling to respond without a tool

drbh:
This PR is an improved version of #2597 and handles processing streaming responses as well as non streaming generations. This PR unrolls tool responses that do not select one of the user provided tools.

This implementation parses the first few tokens of generations requests with user provided tools to determine if a tool was selected. If no tool was selected the response is parsed and returned as a normal `content` value. 

This PR also supersedes draft PR  #2606 as this approach has more benefits and less downsides


router/src/server.rs
```diff
@@ -452,23 +453,39 @@ async fn generate_stream(
...
+        while let Some(raw_event) = response_stream.next().await {
+            match raw_event {
+                Ok(stream_token) => {
+                    let event = Event::default();
+                    let event = event.json_data(stream_token).unwrap();
+                    yield Ok(event);
+                }
+                Err(_err) => {
+                     let event = Event::default();
+                    yield Ok(event);
```
Narsil, line 470: Why ? You're changing functionality here.

drbh: great catch, Thanks. I've updated the logic to emit an event in both cases of a token or an error (to match the existing functionality).

Narsil (approved):
LGTM.

Some fixes in `_err`.
Remove send_function_name logic.
Add `stream` test variant.

integration-tests/models/test_tools_llama.py
```diff
@@ -244,3 +253,44 @@ async def test_flash_llama_grammar_tools_insufficient_information(
...
+        count += 1
+        content_generated += response.choices[0].delta.content
+        last_response = response
+        assert response.choices[0].delta.tool_calls is None
+
+    assert count == 11
+    print(content_generated)
+    assert (
+        content_generated
+        == "There is no weather related function available to answer your prompt"
```
Narsil, line 294: Odd answer to be honest.

drbh: yea I agree, I think its because the model is technically responding to the `error` field in the `notify_error` tool, so the text is more of an error text than an actual response. Going to make a couple changes to see if I can minimize the issue

Narsil (approved):
LGTM

PR #2623, opened 2024-10-08
feat: add test for positional rotary embeddings

drbh:
This PR adds a simple set of tests for `PositionRotaryEmbedding.static` and simple usage of the rotary embed on dummy tensors. The goal is to add more test to avoid any regressions and provide a easy place to debug/test more advance feature of PositionRotaryEmbedding in the future.


drbh:
closing as tests for now, tests can be revisited later if needed

PR #2628, opened 2024-10-09
Add support for FP8 KV cache scales

danieldk:
# What does this PR do?

Since FP8 only has limited dynamic range, we can scale keys/values before storing them into the cache (and unscale them in attention). To avoid rescaling the cache as the absmax values change, good scales are usually determined per layer using calibration calibration data and stored in the checkpoint.

This change adds support for for using key-value scales and loading them
from checkpoints in the two most common formats:

- Separate per-layer `k_scale` and `v_scale` scalars.
- Per-layer `kv_scale` scalar (older format).

Currently, scales are only used with an `float8_e4m3fn` cache.

Besides adding support for key/value scales, the `fp8_quantize` function is also extended to support quantization with a kernel vendored from vLLM. This is slightly faster than the PyTorch implementation, but also scales in FP32, potentially improving accuracy.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/layers/attention/kv_cache.py
+++ b/server/text_generation_server/layers/attention/kv_cache.py
@@ -110,7 +110,11 @@ class KVCache:
         """Check if the cache can be scaled by the given scales."""
         if kv_scales.key_scale_cpu == 1.0 and kv_scales.value_scale_cpu == 1.0:
             return False
-        elif self.dtype == torch.float8_e4m3fn and SYSTEM == "cuda":
+        elif (
+            self.dtype == torch.float8_e4m3fn
+            and ATTENTION == "flashinfer"
+            and SYSTEM == "cuda"
+        ):
             log_once(
                 logger.info,
                 "Using FP8 KV cache scales",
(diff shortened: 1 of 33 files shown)


server/text_generation_server/layers/attention/kv_cache.py
```diff
@@ -94,17 +147,33 @@ def store(
...
+                    scalar=True,
+                )[0]
+            if kv_scales.value_scale_cpu != 1.0:
+                value = fp8_quantize(
+                    value.float(),
+                    scale=kv_scales.value_scale,
+                    qdtype=self.dtype,
+                    scalar=True,
+                )[0]
+
```
mht-sharma, line 172: Should this be only done when ATTENTION is `flashinfer`, since scales are not passed to flash decoding and paged attention yet?

mht-sharma (commented):
LGTM! Thanks for the PR @danieldk. This will help me enable FP8 KV cache on ROCm next.

server/text_generation_server/layers/attention/kv_cache.py
danieldk: Ah yes, good catch! I added an additional condition to `can_scale` to check that `ATTENTION=="flashinfer"`.

mht-sharma (approved):
LGTM thanks!

PR #2640, opened 2024-10-13
update ipex to fix incorrect output of mllama in cpu

sywangyi:
@OlivierDehaene OR @Narsil @danieldk please help review and merge. thanks


Narsil (approved):
LGTM. Any pointers to the origin of the issue maybe ?

sywangyi:
ipex.llm.functional.varlen_attention has bug if  causal = False. fixed in the updated ipex

PR #2645, opened 2024-10-14
PR 2634 CI - Fix the tool_choice format for named choice by adapting OpenAIs scheme

drbh:
This PR re opens #2634 to run CI


drbh:
@linusbierhoff thank you for the contribution! 🙏 I've opened this PR to run CI, and add some other tool calling improvements along with your changes.

This PR now removes the old `ToolType` in favor of a more simple `ChatCompletionToolChoiceOption`. This new struct includes the addition of `TypedChoice` and correctly allows a user to specify the following tools as strings or as an object. Additionally this PR includes support for the `required` tool_choice, which forces the model to use one of the user provided tools. Lastly the new struct also improves the openapi spec and makes the tool choice a `oneOf` which improves the spec readability

router/src/lib.rs
```diff
@@ -956,48 +966,72 @@ pub fn default_tool_prompt() -> String {
...
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema)]
+pub struct FunctionName {
+    pub name: String,
+}
+
+#[derive(Debug, Clone, PartialEq, Serialize, Deserialize, ToSchema, Default)]
+#[serde(from = "ToolTypeDeserializer")]
+pub enum ChatCompletionToolChoiceOption {
```
Narsil, line 983: Can we avoid java naming please ?

router/src/lib.rs
```diff
@@ -956,48 +966,72 @@ pub fn default_tool_prompt() -> String {
...
     /// Means the model can pick between generating a message or calling one or more tools.
     #[schema(rename = "auto")]
-    OneOf,
+    Auto,
     /// Means the model will not call any tool and instead generates a message.
     #[schema(rename = "none")]
+    #[default]
     NoTool,
+    /// Means the model must call one or more tools.
+    #[schema(rename = "required")]
```
Narsil, line 992: `schema rename` is not what you want, you want serde rename most likely.

router/src/lib.rs
```diff
@@ -956,48 +966,72 @@ pub fn default_tool_prompt() -> String {
...
+    Required,
     /// Forces the model to call a specific tool.
     #[schema(rename = "function")]
+    #[serde(alias = "function")]
```
Narsil, line 996: Can we have link in rustdoc to where all these attributes are explained, possibly in the original openai spec (or anthropic)

router/src/lib.rs
```diff
@@ -1604,4 +1638,40 @@ mod tests {
...
+            ChatCompletionToolChoiceOption::Required
+        );
+
+        let ref_choice = ChatCompletionToolChoiceOption::Function(FunctionName {
+            name: "myfn".to_string(),
+        });
+
+        let named = r#"{"tool_choice":"myfn"}"#;
+        let de_named: TestRequest = serde_json::from_str(named).unwrap();
+        assert_eq!(de_named.tool_choice, ref_choice);
```
Narsil, line 1670:
```suggestion
        assert_eq!(de_named.tool_choice,ChatCompletionToolChoiceOption::Function(FunctionName {
            name: "myfn".to_string(),
        }) );
```

Please avoid indirection in tests. Tests are about being as dumb as possible, so a reader can easily assess if the test is correct, or outdated. Every indirection makes that harder to reason about.

router/src/infer/tool_grammar.rs
```diff
@@ -20,44 +20,47 @@ impl ToolGrammar {
...
-                        }
-                    },
-                    "required": ["content"]
-                }),
-            },
-        };
-        tools.push(no_tool);
+        // add the no_tool function to the tools as long as we are not required to use a specific tool
+        if tool_choice != ChatCompletionToolChoiceOption::Required {
+            let no_tool = Tool {
```
Narsil, line 34: Isn't this only ever used in the `auto` branch ? Wouldn't putting it in the match branch make more sense ?

router/src/infer/tool_grammar.rs
```diff
@@ -20,44 +20,47 @@ impl ToolGrammar {
...
         let tools_to_use = match tool_choice {
-            ToolType::Function(function) => {
+            ChatCompletionToolChoiceOption::Function(function) => {
                 vec![Self::find_tool_by_name(&tools, &function.name)?]
             }
-            ToolType::OneOf => tools.clone(),
-            ToolType::NoTool => return Ok((tools, None)),
+            ChatCompletionToolChoiceOption::Required => tools.clone(),
+            ChatCompletionToolChoiceOption::Auto => tools.clone(),
+            ChatCompletionToolChoiceOption::NoTool => return Ok((tools, None)),
```
Narsil, line 63: Why return some tools, if we don't want to use them ? Wouldn't discarding them be better ?

router/src/infer/tool_grammar.rs
```diff
@@ -20,44 +20,47 @@ impl ToolGrammar {
...
         // if tools are provided and no tool_choice we default to the OneOf
         let tools_to_use = match tool_choice {
-            ToolType::Function(function) => {
+            ChatCompletionToolChoiceOption::Function(function) => {
                 vec![Self::find_tool_by_name(&tools, &function.name)?]
             }
-            ToolType::OneOf => tools.clone(),
-            ToolType::NoTool => return Ok((tools, None)),
+            ChatCompletionToolChoiceOption::Required => tools.clone(),
+            ChatCompletionToolChoiceOption::Auto => tools.clone(),
```
Narsil, line 62: Why is this `.clone` needed ? I know it was there before, but since we're upgrading this, couldn't we get rid of it ?

router/src/infer/tool_grammar.rs
Narsil, line 63: Also I think early returns in `match` branches are bug nests, it's easy to overlook those branches, returning an empty slice and letting the rest roll seems slightly better (code reading wise)

router/src/lib.rs
```diff
@@ -849,8 +849,8 @@ pub(crate) struct ChatRequest {
 
     /// A specific tool to use. If not provided, the model will default to use any of the tools provided in the tools parameter.
     #[serde(default)]
-    #[schema(nullable = true, example = "null")]
-    pub tool_choice: ToolChoice,
+    #[schema(nullable = true, default = "null", example = "null")]
+    pub tool_choice: Option<ChatCompletionToolChoiceOption>,
```
Narsil, line 853: Why the change to `Option` you have the `NoTool` choice, there shouldn't be a need to `Option` wrapping is there ?

router/src/lib.rs
```diff
@@ -906,8 +906,18 @@ impl ChatRequest {
         let (inputs, grammar, using_tools) = prepare_chat_input(
             infer,
             response_format,
-            tools,
-            tool_choice,
+            tools.clone(),
```
Narsil, line 909: No clone.

router/src/lib.rs
```diff
@@ -906,8 +906,18 @@ impl ChatRequest {
...
+            // unwrap or default (use "auto" if tools are present, and "none" if not)
+            tool_choice.map_or_else(
```
Narsil, line 911: That should be already done in the deserialization, there are about 3 different function calls to `preprare_chat_input`.

Narsil (changes requested):

Narsil:
`required` is that an OpenAI spec ?

Can you point to the original references for the things we're adding.
Also dropping the old code is a pretty bad breaking change, it should be in big bold red !.

Why are we dropping support for the old way, if it was there, is it documented anywhere ? Should we update the old doc ? 

This PR is also seriously lacking integration tests for all of these. The deserialization tests are nice, but not enough.

router/src/lib.rs
drbh: yea I can revert the name change if thats better.

The name change is to align with the naming that OpenAI is using in their spec. https://github.com/openai/openai-openapi/blob/d033c364c6574068ee89f3d5f845a4830bddd503/openapi.yaml#L9581, I aligned the names because I thought it may be helpful in automatically checking the spec, however it is a very long name..

router/src/lib.rs
drbh: oh I actually included it to improve the generated schema

this macro changes the `enum` on the schema to show the expected lowercase text instead of `Required`
```json
{
  "type": "string",
  "description": "Means the model must call one or more tools.",
  "enum": [
    "required"
  ]
},
```

router/src/lib.rs
drbh: yep, just added a useful link above the enum that explains all 4 of the options in OpenAI's docs this link will also be included in the openapi schema for the `ChatCompletionToolChoiceOption` description

router/src/lib.rs
drbh: great points, indirection removed in the latest commit

router/src/infer/tool_grammar.rs
drbh: yep good point. Just moved into the match branch and removed the `let mut tools` and only mutate tools if the Auto option is chosen. thanks

router/src/infer/tool_grammar.rs
drbh: yes agreed they should be removed. Just removed all of the tool clones above and improved the match in general

router/src/lib.rs
drbh: `Option` is needed so the user can send a http request without a `tool_choice` param included. If `tool_choice` is `None` (user did not include a tool_choice in the request), `ChatCompletionToolChoiceOption` defaults to `NoTool`

router/src/lib.rs
drbh: removed, it was previously required because of a check of the `tools` vec that was below this line (moved it up and removed the clone). thanks

drbh:
> required is that an OpenAI spec ?

yea `required` is supported by OpenAI, see docs here: https://platform.openai.com/docs/guides/function-calling/configuring-function-calling-behavior-using-the-tool_choice-parameter. Note** OpenAI added this parameter after TGI's initial tool implementation (hence why it was missing), additionally prior to https://github.com/huggingface/text-generation-inference/pull/2614 `auto` and `required` had the same functionality in TGI, however now its more pressing/great to have the api's aligned.

> Also dropping the old code is a pretty bad breaking change, it should be in big bold red !.

yea regarding the removed code - as the original PR noted, the `function` tool choice was expecting the incorrect format. I believe I confused some of the function calling formats and ended up adding one that was not aligned with any major LLM api. 

regarding flagging the breaking changes, would it be best to include the old API but throw a helpful error if it is used? Currently I've removed the path in favor of the corrected one. 

> Why are we dropping support for the old way, if it was there, is it documented anywhere ? Should we update the old doc ?

we're dropping support since it was incorrectly implemented, we do not have documentation specific to this tool choice, however I'm going to push another commit with updated docs for the supported tool choice types

> This PR is also seriously lacking integration tests for all of these. The deserialization tests are nice, but not enough.

just added a test for `required` and one for a function object in `test_tools_llama.py`

router/src/infer/tool_grammar.rs
drbh: good point, updated in the latest to commit to discard the tools and continue executing the function normally

drbh: this was originally added to allow passing tools to the chat template without enforcing a grammar, which was introduced in this PR https://github.com/huggingface/text-generation-inference/pull/2454, but I believe should be removed in favor of discarding the tools. The latest commits discard tools when the choice is `NoTool`

router/src/lib.rs
Narsil: > I thought it may be helpful in automatically checking the spec

Then let's wait for it to be actually relevant, it's not usually a good thing to do things "because we might need it". Either you need it or you don't.

Personally I don't think it has any relevance to use their names.

router/src/lib.rs
Narsil: Doesn't a default take care of that ? And shouldn't the default be "auto"?

integration-tests/models/test_tools_llama.py
```diff
@@ -327,3 +329,102 @@ async def test_flash_llama_grammar_tools_sea_creatures_stream(
...
+                if line == "[DONE]":
+                    break
+                response = json.loads(line)
+                tool_calls_generated += response["choices"][0]["delta"]["tool_calls"][
+                    "function"
+                ]["arguments"]
+                last_response = response
+
+    assert count == 30
+    print(tool_calls_generated)
```
Narsil, line 425: Oops

router/src/lib.rs
```diff
@@ -903,15 +903,47 @@ impl ChatRequest {
...
-            guideline,
-            messages,
-        )?;
+        // unwrap or default (use "auto" if tools are present, and "none" if not)
+        let tool_choice = tool_choice.unwrap_or_else(|| {
+            if tools.is_some() {
+                ChatCompletionToolChoiceOption::Auto
+            } else {
+                ChatCompletionToolChoiceOption::NoTool
+            }
```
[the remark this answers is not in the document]
Narsil: What happens if tool_choice is "required" and no tools are sent ?

("auto", tools = []) we should just ignore the grammar right ?

router/src/lib.rs
Narsil: Yes, and serde(renamed) actually modifies what the input is from the user.

Narsil: Actually a better serde(rename) on the whole struct should fix every field (both schema and actual JSON parsing)

router/src/lib.rs
drbh: got it that make sense, I've changed the name to `ToolChoice` in the latest commit. (its much cleaner 👍)

router/src/lib.rs
drbh: oh yea thats much better, updated in the latest commit

integration-tests/models/test_tools_llama.py
```diff
@@ -327,3 +329,102 @@ async def test_flash_llama_grammar_tools_sea_creatures_stream(
...
+                },
+                {
+                    "role": "user",
+                    "content": "Tell me a story about 3 sea creatures",
+                },
+            ],
+            "tools": tools,
+            "tool_choice": {
+                "type": "function",
+                "function": {"name": "get_current_weather"},
```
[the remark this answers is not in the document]
drbh: agreed, just updated to choose `get_n_day_weather_forecast` instead

integration-tests/models/test_tools_llama.py
drbh: oops 😬, removed in latest commit. thanks!

router/src/infer/tool_grammar.rs
```diff
@@ -20,19 +19,19 @@ impl ToolGrammar {
...
     pub fn apply(
         tools: Vec<Tool>,
-        tool_choice: ChatCompletionToolChoiceOption,
+        tool_choice: ToolChoice,
     ) -> Result<(Vec<Tool>, Option<JsonSchemaTool>), InferError> {
-        // if no tools are provided, we return None
+        // if no tools are provided, we return None and an empty vec
         if tools.is_empty() {
-            return Ok((tools, None));
+            return Ok((Vec::with_capacity(0), None));
```
Narsil, line 26: `vec![]` ?

router/src/lib.rs
```diff
@@ -1002,21 +996,18 @@ pub struct FunctionName {
...
-pub enum ChatCompletionToolChoiceOption {
+pub enum ToolChoice {
     /// Means the model can pick between generating a message or calling one or more tools.
-    #[schema(rename = "auto")]
+    #[default]
     Auto,
     /// Means the model will not call any tool and instead generates a message.
     #[schema(rename = "none")]
-    #[default]
     NoTool,
```
Narsil, line 1007: Isn't this `no_tool` with `snake_case`  ?This should mean a rename of this property or `None`,  no ?

I don't think `schema(rename)` imples a serde `rename`.

router/src/infer/tool_grammar.rs
drbh: updated in latest commit

router/src/lib.rs
```diff
@@ -850,13 +850,13 @@ pub(crate) struct ChatRequest {
     /// A specific tool to use. If not provided, the model will default to use any of the tools provided in the tools parameter.
     #[serde(default)]
     #[schema(nullable = true, default = "null", example = "null")]
-    pub tool_choice: Option<ChatCompletionToolChoiceOption>,
+    pub tool_choice: Option<ToolChoice>,
```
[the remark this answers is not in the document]
drbh: oh yes that worked, I was originally under the impression the Option was still needed to handle all http request but this is much cleaner. Thank you

router/src/lib.rs
```diff
@@ -850,13 +850,13 @@ pub(crate) struct ChatRequest {
...
 
     /// Response format constraints for the generation.
     ///
     /// NOTE: A request can use `response_format` OR `tools` but not both.
     #[serde(default)]
-    #[schema(nullable = true, default = "null", example = "null")]
+    #[schema(nullable = true, default = "auto", example = "auto")]
```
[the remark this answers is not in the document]
drbh: yes thank you, updated the correct argument in the latest commit

router/src/lib.rs
drbh: good catch, updated to `serde(rename)` and added a test to ensure that the `"none"` values is respected correctly

router/src/infer/tool_grammar.rs
Narsil: I don't see it.

drbh: `let tools_to_use = match tool_choice` was updated to return a `vec![]` rather than a `Vec::with_capacity(0)`, and the conditional was moved below `tools_to_use` to return none in both cases (no tools supplied, or if NoTools specified) https://github.com/huggingface/text-generation-inference/blob/16e4c38898ed36ee44237314ce823bf3d81ea90c/router/src/infer/tool_grammar.rs#L55

drbh:
merging optimistically as CI is green and issues have been resolved. Also mentioned in the internal #text-generation-inference Slack channel.

Will watch for any regressions and revert if needed

PR #2646, opened 2024-10-14
remove docker entrypoint

oOraph:
this entrypoint was meant to fix tgi on some cloud providers like gcp that do not ensure the ld cache is refreshed after adding some shared libraries in custom container hooks

this should not belong to tgi repo itself

cc @co42 @christophe-rannou @Narsil


christophe-rannou (approved):

oOraph:
side note: ci failures are due to missing creds

PR #2655, opened 2024-10-16
Support `e4m3fn` KV cache

danieldk:
# What does this PR do?

Add support for `e4m3fn` KV caches as well.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/server/text_generation_server/layers/attention/kv_cache.py
+++ b/server/text_generation_server/layers/attention/kv_cache.py
@@ -24,10 +24,8 @@ class KVCache:
     ):
         """Construct the key-value cache for a layer."""
 
-        if (
-            dtype.itemsize == 1
-            and dtype.is_floating_point
-            and (ATTENTION != "flashinfer" or SYSTEM != "cuda")
+        if dtype in {torch.float8_e5m2, torch.float8_e4m3fn} and (
+            ATTENTION != "flashinfer" or SYSTEM != "cuda"
         ):
             raise ValueError(
                 "FP8 KV cache is currently only supported for flashinfer on CUDA"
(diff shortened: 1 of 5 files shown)


server/text_generation_server/layers/attention/kv_cache.py
```diff
@@ -24,11 +24,13 @@ def __init__(
     ):
         """Construct the key-value cache for a layer."""
 
-        if dtype == torch.float8_e5m2 and (
-            ATTENTION != "flashinfer" or SYSTEM != "cuda"
+        if (
+            dtype.itemsize == 1
+            and dtype.is_floating_point
```
Narsil, line 29: Why not `in {torch.float8_e5m2, torch.float8_e4m3}` ?

danieldk: We might have other variants in the future as well (`e4m3fnuz`, `e5m2uz`).

Narsil: Fair enough. From my understanding they are actually the same dtypes only side effectifs on edge values (which we won't check against).

I don't even think torch have plans for explicit support https://pytorch.org/docs/stable/tensors.html#data-types

As you wish for merging, I just feel like having consistent checks everywhere will lead to fewer bugs.

danieldk: Yeah, that's a fair point. I think `in {torch.float8_e5m2, torch.float8_e4m3}` makes it easier to grep for all the places where one of these float types is used. So I'll update the PR.

PR #2661, opened 2024-10-17
fix: prefer inplace softmax to avoid copy

drbh:
This PR modifies `log_softmax` to operate in place, eliminating the need to copy large tensors. This optimization reduces memory consumption during warmup.

For instance, when using `meta-llama/Meta-Llama-3.1-8B-Instruct` on an L4, this change allows running the model with `--max-batch-prefill-tokens` increased from 7192 to 9874 without exceeding memory limits

### DIFF
--- a/server/text_generation_server/models/flash_causal_lm.py
+++ b/server/text_generation_server/models/flash_causal_lm.py
@@ -1922,8 +1922,9 @@ class FlashCausalLM(Model):
             batch.adapter_meta.adapter_indices = next_adapter_indices
 
         if prefill and prefill_logprobs:
-            # Get prefill logprobs
-            prefill_logprobs_tensor = torch.log_softmax(out, -1)
+            # Get prefill logprobs with inplace softmax (avoid copying the `out` tensor (max_batch_size * vocab_size))
+            torch.log_softmax(out, -1, out=out)
+            prefill_logprobs_tensor = out
             prefill_logprobs = torch.gather(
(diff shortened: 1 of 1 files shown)


Narsil:
> For instance, when using meta-llama/Meta-Llama-3.1-8B-Instruct on an L4, this change allows running the model with --max-batch-prefill-tokens increased from 7192 to 9874 without exceeding memory limits

With chunking in place, we don't want to increase max-batch-prefill-tokens anymore, but instead choose an optimal value :).

That being said, nice find, probably a lot of opportunities for such updates (let's focus on the large tensors though)

Narsil (approved):

server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -1922,8 +1922,9 @@ def generate_token(
             batch.adapter_meta.adapter_indices = next_adapter_indices
 
         if prefill and prefill_logprobs:
-            # Get prefill logprobs
-            prefill_logprobs_tensor = torch.log_softmax(out, -1)
+            # Get prefill logprobs with inplace softmax (avoid copying the `out` tensor (max_batch_size * vocab_size))
```
Narsil, line 1925:
```suggestion
            # Get prefill logprobs with inplace softmax (avoid copying the `out` tensor (max_batch_prefill_tokens * vocab_size))
```

There is not batch size anymore per-se.

PR #2665, opened 2024-10-18
CI job. Gpt awq 4

Narsil:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


server/text_generation_server/layers/awq/quantize/ipex.py
```diff
@@ -0,0 +1,48 @@
...
+            raise NotImplementedError("Only 4-bit are supported for now.")
+
+        self.in_features = qweight.shape[0]
+        self.out_features = qweight.shape[1] * 32 // w_bit
+
+        self.w_bit = w_bit
+        self.group_size = group_size if group_size != -1 else self.in_features
+        # quick sanity check (make sure aligment)
+        assert self.in_features % self.group_size == 0
+        assert self.out_features % (32 // self.w_bit) == 0
```
danieldk, line 23: Isn't this guaranteed by how `self.out_features` is defined above?

server/text_generation_server/layers/gptq/__init__.py
```diff
@@ -332,6 +336,7 @@ def get_weights_row(self, weights: Weights, prefix: str):
                     # Exllama implementation does not support row tensor parallelism with act-order, as
                     # it would require to reorder input activations that are split unto several GPUs
                     use_exllama = False
+                    desc_act = True
```
danieldk, line 339: I don't know why this ignores the value of `desc_act` given by the configuration, isn't that the source of truth? Do we expect some models to use activation sorting but lying about it in the config? Needs a comment.

server/text_generation_server/layers/gptq/__init__.py
```diff
@@ -321,7 +325,7 @@ def get_weights_row(self, weights: Weights, prefix: str):
             if g_idx is not None:
                 if (
                     not torch.equal(
-                        g_idx.cpu(),
+                        (g_idx - g_idx[0]).cpu(),
```
danieldk, line 328: This block needs a comment what is computed here. The old code was already difficult to read, but I the way I read it is that it is checking whether `g_idx` incrementing indices by group (so there is a `g_idx`, but no activation sorting).

But what is this subtraction doing?

danieldk: I think it's subtracting the first index, because say we have `g_idx` with `groupsize = 2`

```
[0 0 1 1 2 2 3 3]
```

If we have two shards, then it gets broken up into

```
[0 0 1 1] [2 2 3 3]
```

And then the check fails for the second shard. Subtracting `g_idx[0]` then changes this into

```
[0 0 1 1] [0 0 1 1]
```

Making the check valid. But there should be a comment to make it easier to understand without knowing all the context.

server/text_generation_server/layers/gptq/ipex.py
```diff
@@ -0,0 +1,126 @@
...
+                bias=self.bias,
+                group_size=self.groupsize,
+                g_idx=g_idx,
+                quant_method=ipex.llm.quantization.QuantMethod.GPTQ_GEMM,
+                dtype=ipex.llm.quantization.QuantDtype.INT4,
+            )
+        )
+
+    @classmethod
+    def new(cls, bits, groupsize, infeatures, outfeatures, bias):
```
danieldk, line 44: Doesn't seem used?

server/text_generation_server/layers/gptq/ipex.py
```diff
@@ -0,0 +1,126 @@
...
+        g_idx = torch.tensor(
+            [i // groupsize for i in range(infeatures)], dtype=torch.int32
+        )
+        if bias:
+            bias = torch.zeros((outfeatures), dtype=torch.float16)
+        else:
+            bias = None
+        return cls(qweight, qzeros, scales, g_idx, bias, bits, groupsize)
+
+    def pack(self, linear, scales, zeros, g_idx=None):
```
danieldk, line 65: Not used.

server/text_generation_server/layers/gptq/__init__.py
Narsil: YEs this is the core idea of the fix. The logic was only working on TP=1, it's now working on TP>1 (to detect that g_idx is redundant and can be safely ignored)

server/text_generation_server/layers/gptq/ipex.py
Narsil: It's all copied from the cuda code.

server/text_generation_server/layers/gptq/ipex.py
Narsil: It's all copied from the cuda code.

server/text_generation_server/layers/awq/quantize/ipex.py
Narsil: This is all copied from the original GTPQ code.

danieldk: 👍

server/text_generation_server/layers/gptq/__init__.py
danieldk: Yeah makes sense, would be a good to have a short comment here though so that future-us don't have to re-parse what is going on here.

server/text_generation_server/layers/gptq/__init__.py
danieldk: ack

server/text_generation_server/layers/gptq/__init__.py
Narsil, line 328:
```suggestion
                        # Remove g_idx[0] to adapt the check with TP>1.
                        (g_idx - g_idx[0]).cpu(),
```

PR #2686, opened 2024-10-24
fix: improve find_segments via numpy diff

drbh:
This PR updates `find_segments` to use a numpy array rather than a list. This allow us to a more efficiently find the segments via masking. 

when testing the method in isolation this results in a ~3x speedup on 4096 indices

```
Benchmarking results:
--------------------------------------------------------------------------------
        Size   Original (ms)  Optimized (ms)    Speedup
--------------------------------------------------------------------------------
        4096            0.32            0.08       3.94
        8192            0.59            0.14       4.13
       16384            1.24            0.20       6.06
       32768            2.31            0.32       7.12
       65536            4.57            0.58       7.93
```

### DIFF
--- a/server/text_generation_server/utils/segments.py
+++ b/server/text_generation_server/utils/segments.py
@@ -5,30 +5,25 @@
 from typing import List, Tuple, Union
 
 import torch
+import numpy as np
 
 
-# FIXME: this should be optimized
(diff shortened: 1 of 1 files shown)


OlivierDehaene (approved):
Nice

PR #2688, opened 2024-10-24
Switch from fbgemm-gpu w8a8 scaled matmul to vLLM/marlin-kernels

danieldk:
# What does this PR do?

Performance and accuracy of these kernels are on par (tested with Llama 70B and 405B). Removes a dependency and resolves some stability issues we have been seeing.

Two test snapshots had to be changed. It seems like the perturbations come from switching from `bfloat16` (which was the only dtype supported by fbgemm FP8 quantization) to `float16` which is our default dtype for Llama (even though recent Llama checkpoints are `bfloat16`).

I added an environment variable `USE_CUTLASS_W8A` to force use of the cutlass kernels on compute capability 8. The question why we use FP8-Marlin on compute capability 8.9 comes up occasionally (because it's faster and probably more accurate because it's in FP16), so this is a backdoor to use W8A8 on 8.9 in case somebody really wants it.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


drbh (approved):
LGTM

PR #2689, opened 2024-10-24
Support qwen2 vl

drbh:
This is a work in progress PR to support qwen2-vl. Currently these changes include loading the model weights and a functioning vision model. Remaining work is to adjust the existing qwen2 model to handle multimodal requests/positional embeddings.

status:
- [X] load weights
- [X] support prefill warmup run
- [X] accept chat image request and process image
- [X] align vision model output with reference impl
- [X] correctly merge the processed image and text model
- [X] avoid any reshapes and allocations during runtime possible

remaining:
- [x] resolve remaining bug with position ids
- [x] align test output with reference
- [x] cleanup remaining todos/refactors/improvements

further
- make improvements
- improve test coverage

### DIFF
--- a/server/text_generation_server/layers/tensor_parallel.py
+++ b/server/text_generation_server/layers/tensor_parallel.py
@@ -144,7 +144,7 @@ class TensorParallelColumnLinear(SuperLayer):
             num_key_value_heads=num_key_value_heads,
         )
         if bias:
-            raise NotImplementedError("packed_qkv only implemented for baichuan")
+            bias = weights.get_tensor(f"{prefix}.bias")
         else:
             bias = None
         linear = get_linear(weight, bias)
--- a/server/text_generation_server/layers/tensor_parallel.py
+++ b/server/text_generation_server/layers/tensor_parallel.py
@@ -144,7 +144,7 @@ class TensorParallelColumnLinear(SuperLayer):
             num_key_value_heads=num_key_value_heads,
         )
         if bias:
-            bias = weights.get_tensor(f"{prefix}.bias")
+            raise NotImplementedError("packed_qkv only implemented for baichuan")
         else:
             bias = None
         linear = get_linear(weight, bias)
--- a/server/text_generation_server/models/custom_modeling/flash_qwen2_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/flash_qwen2_modeling.py
@@ -130,28 +130,23 @@ class Qwen2Attention(torch.nn.Module):
         query = query.view(-1, self.num_heads, self.head_size)
         kv = kv.view(-1, 2, self.num_key_value_heads, self.head_size)
 
-        # TODO: correctly handle the multimodal case
-        if False:
-            self.rotary_emb(query, torch.select(kv, dim=1, index=0), cos, sin)
-        else:
-            # multimodal rotary
-            unsqueeze_dim = 1
-            mrope_section = self.mrope_section * 2
-            cos = torch.cat(
-                [m[i % 3] for i, m in enumerate(cos.split(mrope_section, dim=-1))],
-                dim=-1,
-            ).unsqueeze(unsqueeze_dim)
-            sin = torch.cat(
-                [m[i % 3] for i, m in enumerate(sin.split(mrope_section, dim=-1))],
-                dim=-1,
-            ).unsqueeze(unsqueeze_dim)
+        _query = query.clone()
+        _cos = cos.clone()
+        _sin = sin.clone()
 
-            _query = query.transpose(0, 1).unsqueeze(0)
-            _key = torch.select(kv, dim=1, index=0).transpose(0, 1).unsqueeze(0)
-            q_embed = (_query * cos) + (rotate_half(_query) * sin)
-            k_embed = (_key * cos) + (rotate_half(_key) * sin)
-            query = q_embed.squeeze(0).transpose(0, 1)
-            kv[:, 0] = k_embed.squeeze(0).transpose(0, 1)
+        self.rotary_emb(_query, torch.select(kv, dim=1, index=0), cos, sin)
+
+        _cos = torch.cat((_cos, _cos), dim=-1)
+        _sin = torch.cat((_sin, _sin), dim=-1)
+        q_emb = (_query * _cos).reshape(2, 1, -1) + (
+            rotate_half(_query) * _sin
+        ).reshape(2, 1, -1)
+        k_emb = (torch.select(kv, dim=1, index=0) * _cos).reshape(2, 1, -1) + (
+            rotate_half(torch.select(kv, dim=1, index=0)) * _sin
+        ).reshape(2, 1, -1)
+
+        query = q_emb.reshape(-1, self.num_heads, self.head_size)
+        kv[:, 0] = k_emb.reshape(-1, self.num_key_value_heads, self.head_size)
 
         if prefill_cache_indices is not None:
             kv_to_cache = kv[prefill_cache_indices]
@@ -299,9 +294,6 @@ class Qwen2Model(torch.nn.Module):
         process_group = weights.process_group
         self.tp_rank = process_group.rank()
         self.tp_world_size = process_group.size()
-        self.embed_tokens = TensorParallelEmbedding(
-            prefix=f"{prefix}.embed_tokens", weights=weights
-        )
         self.layers = nn.ModuleList(
             [
                 Qwen2Layer(
@@ -325,7 +317,7 @@ class Qwen2Model(torch.nn.Module):
 
     def forward(
         self,
-        input_ids: torch.Tensor,
+        inputs_embeds: torch.Tensor,
         position_ids: torch.Tensor,
         cu_seqlen_prefill: Optional[torch.Tensor],
         kv_cache: List[Tuple[torch.Tensor, torch.Tensor]],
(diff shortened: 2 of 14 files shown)


server/text_generation_server/layers/tensor_parallel.py
```diff
@@ -144,7 +144,7 @@ def load_qkv(
             num_key_value_heads=num_key_value_heads,
         )
         if bias:
-            raise NotImplementedError("packed_qkv only implemented for baichuan")
+            bias = weights.get_tensor(f"{prefix}.bias")
```
Narsil, line 147: This is wrong, no ?

We get the whole bias on Row Parallel, for column, you need to take the actual slice, which for qkv you need to follow the same layout as the weights I think (except on dim=0 instead of dim=1)

server/text_generation_server/models/custom_modeling/flash_qwen2_modeling.py
```diff
@@ -122,7 +130,28 @@ def forward(
         query = query.view(-1, self.num_heads, self.head_size)
         kv = kv.view(-1, 2, self.num_key_value_heads, self.head_size)
 
-        self.rotary_emb(query, torch.select(kv, dim=1, index=0), cos, sin)
+        # TODO: correctly handle the multimodal case
+        if False:
+            self.rotary_emb(query, torch.select(kv, dim=1, index=0), cos, sin)
```
Narsil, line 135: This should be correct. Ignore tiny differences there, this code is exactly what you have underneath (and much more efficient).

I read transformers comment on this, it seems from what I'm reading that they are just applying part of the tensors there, so a regular slicing should do the work.

The problem with the other part, is that our cos, sin are layed out differently than theirs, so you're gonna have issues keeping transformers code and merge it with our own.

server/text_generation_server/models/custom_modeling/flash_qwen2_modeling.py
```diff
@@ -306,12 +335,24 @@ def forward(
         max_s: int,
         true_max_s: int,
         prefill_cache_indices: Optional[torch.Tensor],
+        inputs_embeds: Optional[torch.Tensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
     ) -> torch.Tensor:
-        hidden_states = self.embed_tokens(input_ids)
+
+        # if inputs_embeds are supplied from an external model (vision model) then avoid embedding input_ids
+        if inputs_embeds is not None:
```
Narsil, line 343: No, remove this.

The way it's done here, is that we life input_ids to the parent class, and this always takes `input_embeds`. Makes signatures much cleaner. (It's already done for llama this way if you want to check)

server/text_generation_server/models/custom_modeling/flash_qwen2_modeling.py
```diff
@@ -306,12 +335,24 @@ def forward(
         max_s: int,
         true_max_s: int,
         prefill_cache_indices: Optional[torch.Tensor],
+        inputs_embeds: Optional[torch.Tensor] = None,
+        attention_mask: Optional[torch.Tensor] = None,
```
Narsil, line 339: No attention_mask

server/text_generation_server/models/custom_modeling/qwen2_vl.py
```diff
@@ -0,0 +1,550 @@
...
+                    (input_ids == self.image_token_id)
+                    .unsqueeze(-1)
+                    .expand_as(inputs_embeds)
+                    .to(inputs_embeds.device)
+                )
+                image_embeds = image_embeds.to(
+                    inputs_embeds.device, inputs_embeds.dtype
+                )
+                # input embeddings are masked with image embeddings
+                inputs_embeds = inputs_embeds.masked_scatter(image_mask, image_embeds)
```
Narsil, line 415: Why does

```python
inputs_embeds[input_ids == self.image_token_id] = image_embeds
```

Doesn't work ?

server/text_generation_server/models/custom_modeling/qwen2_vl.py
```diff
@@ -0,0 +1,550 @@
...
+        image_sizes: Optional[torch.LongTensor] = None,
+        adapter_data: Optional[torch.Tensor] = None,
+        cross_attention_states: Optional[torch.Tensor] = None,
+        image_indices=None,
+    ):
+        # make an attention_mask that is (batch_size, sequence_length)
+        attention_mask = torch.ones_like(
+            input_ids, dtype=torch.bool, device=input_ids.device
+        )
+        inputs_embeds = self.text_model.embed_tokens(input_ids)
```
Narsil, line 399: When you lift this, this will be gone from the `text_model` and be here directly.

server/text_generation_server/models/custom_modeling/qwen2_vl.py
```diff
@@ -0,0 +1,550 @@
...
+                3,
+                input_ids.shape[0],
+                input_ids.shape[1],
+                dtype=input_ids.dtype,
+                device=input_ids.device,
+            )
+            image_index, video_index = 0, 0
+
+            for i, input_ids in enumerate(total_input_ids):
+                if attention_mask is not None:
```
Narsil, line 431: This whole thing is extremely poor code (lots of loops, lots of CPU/GPU back&forth).

I think ditching it altogether will be easier than trying to adapt it.

server/text_generation_server/layers/tensor_parallel.py
drbh: after comparing with transformers it seems like `weights.get_tensor(f"{prefix}.bias")` and `weights.get_sharded(f"{prefix}.qkv.bias", dim=0)` return the exact bias as the one in the reference.

I've reverted the change within `tensor_parallel.py::load_qkv` in favor of setting the bias after creating the linear in `qwen_vl.py` via `weights.get_sharded(f"{prefix}.qkv.bias", dim=0)`

for reference:
```python
self.qkv = TensorParallelColumnLinear.load_qkv(
    config,
    prefix=f"{prefix}.qkv",
    weights=weights,
    bias=False,
    num_heads=self.num_heads,
    num_key_value_heads=self.num_heads,
)
self.qkv.linear.bias = weights.get_sharded(f"{prefix}.qkv.bias", dim=0)
```

this hopefully makes the qkv loading a bit more clear

server/text_generation_server/models/custom_modeling/flash_qwen2_modeling.py
drbh: yes agreed this is much cleaner. I've updated the classes in the latest commit to follow the same pattern as llama

server/text_generation_server/models/custom_modeling/flash_qwen2_modeling.py
drbh: removed in latest commit

server/text_generation_server/models/custom_modeling/qwen2_vl.py
drbh: ha yea thats a much better way to write this! I've included this change along with a overall better rewrite of this logic in the latest commit.

server/text_generation_server/models/custom_modeling/qwen2_vl.py
drbh: updated in latest commit

server/text_generation_server/models/custom_modeling/qwen2_vl.py
drbh: yea agreed, in the latest commit i've moved this logic into a function `get_position_ids` and rewritten a more simple version that avoids most of the gpu/cpu copies. I'll revisit later to see if I can simplify further (avoid the loop) but the changes may provide a bit of a performance and readability improvement.

server/text_generation_server/models/vlm_causal_lm.py
```diff
@@ -343,6 +363,16 @@ def forward(
...
 
+        if hasattr(self.model, "get_position_ids"):
+            if position_ids.shape[0] != 1:
+                position_ids = self.model.get_position_ids(
+                    input_ids.unsqueeze(0), batch.image_grid_thw
+                )
+                batch.position_ids = position_ids[0, 0, :]
+            else:
+                position_ids = position_ids.repeat(3, 1, 1).clone()
+                batch.position_ids = position_ids[0, 0, :]
```
Narsil, line 374: This seems very wrong, no ?

server/text_generation_server/models/vlm_causal_lm.py
```diff
@@ -343,6 +363,16 @@ def forward(
             max_s = batch.max_current_length
             lm_head_indices = batch.prefill_head_indices
 
+        if hasattr(self.model, "get_position_ids"):
+            if position_ids.shape[0] != 1:
+                position_ids = self.model.get_position_ids(
+                    input_ids.unsqueeze(0), batch.image_grid_thw
+                )
+                batch.position_ids = position_ids[0, 0, :]
```
Narsil, line 371: Why create so many position ids, just to discard them ?

server/text_generation_server/models/custom_modeling/qwen2_vl.py
```diff
@@ -0,0 +1,509 @@
...
+            prefix="lm_head", weights=weights, config=config, bias=False
+        )
+        self.norm = FastRMSNorm.load(
+            prefix="model.norm",
+            weights=weights,
+            eps=config.rms_norm_eps,
+        )
+        self.device = weights.device
+
+    def get_position_ids(
```
Narsil, line 377: Seems overly complex and bloated.

Let's keep this if it's working, but it's definitely fixable I think

server/text_generation_server/models/vlm_causal_lm.py
```diff
@@ -343,6 +363,16 @@ def forward(
             max_s = batch.max_current_length
             lm_head_indices = batch.prefill_head_indices
 
+        if hasattr(self.model, "get_position_ids"):
+            if position_ids.shape[0] != 1:
+                position_ids = self.model.get_position_ids(
+                    input_ids.unsqueeze(0), batch.image_grid_thw
```
Narsil, line 369: No unsqueeze, fix the function.

Narsil (approved):

drbh:
merging to enable qwen2-vl

Will follow up with an improvement PR soon. specifically:
- [ ] improve decode
- [ ] improve vision head
- [ ] improve batch to handle multi dimensional position ids
- [ ] remove complex position logic if possible

PR #2694, opened 2024-10-25
Check if allowed tokens is None

aW3st:
Bug fix for the rebased PR


server/text_generation_server/utils/logits_process.py
```diff
@@ -498,9 +498,10 @@ def __call__(
     ):
         if fsm_grammar_state == -1 or self.fsm is None:
             return logits
-        allowed_tokens = self.fsm.allowed_token_ids(fsm_grammar_state)
+        allowed_tokens = self.fsm.get_next_instruction(fsm_grammar_state).tokens
         mask = torch.full_like(logits, -math.inf)
-        mask[:, allowed_tokens] = 0
+        if allowed_tokens is not None:
+	    mask[:, allowed_tokens] = 0
```
aW3st, line 504: @Narsil `Generate.tokens` has type `Optional[List[int]]`, so I think we need to add this check.

PR #2711, opened 2024-10-31
Fix prefix caching + speculative decoding

tgaddair:
# What does this PR do?

Following #2402, it was [remarked](https://github.com/huggingface/text-generation-inference/pull/2402#issuecomment-2419533653) by @ZTianle that enabling speculative decoding when using prefix caching was resulting in incorrect results. We also internally noticed that it sometimes caused CUDA device-side assertion failures. 

The root cause was determined to be incorrect slot assignments that were occurring when calling `slots.expand` and incrementing the slots by `arange_int`. This assumed that the next slots in the allocation were the next ordinal slot IDs, but when using prefix caching, this often is not the case. Due to the radix tree format, slot IDs are frequently discontiguous, and so we need to expand the `slot_indices` and use those to index into the allocated slots rather than expanding the slots directly.

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

cc @OlivierDehaene @Narsil @danieldk

### DIFF
--- a/server/text_generation_server/models/flash_causal_lm.py
+++ b/server/text_generation_server/models/flash_causal_lm.py
@@ -887,11 +887,12 @@ class FlashCausalLMBatch(Batch):
             fsm_grammar_states=fsm_grammar_states,
         )
 
-        speculative_ids = (
-            torch.cat([b.speculative_ids for b in batches], dim=0)
-            if batches[0].speculative_ids is not None
-            else None
-        )
+        # We skip computing the speculative_ids when the batch size is too large, so
+        # we must check that all batches have them, otherwise they must be discarded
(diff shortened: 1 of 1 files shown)


Narsil:
Thanks for the PR, accepting this one since the fixed seems to work (will patch main directly after)

ZTianle:
Thank you @tgaddair for fixing the speculative decoding issue with prefix caching. I've noticed that while this fix works well with flash decoding+prefix caching, there still appear to be compatibility issues when using speculative decoding together with flashinfer  - specifically, it continues to generate incorrect outputs.
Has there been any investigation into making speculative decoding work properly with the flashinfer and flashinfer + prefix caching combination? Would be helpful to understand if this is on the roadmap or if there are known technical challenges preventing this integration. @Narsil @danieldk

PR #2721, opened 2024-11-04
feat: support flash attention 2 in qwen2 vl vision blocks

drbh:
This PR adds support for flash attention within the vision blocks of qwen2vl. This improves latency and reduces spikes in memory usage from sdpa attention (helps avoid oom'ing on large images).

running
```
text-generation-launcher --model-id Qwen/Qwen2-VL-7B-Instruct
``` 
on a single `L4`'s with a small model this has a ~5% perf improvement: prev: `~67ms` per token current: `~63ms` per token.

ps** it's likely the performance improvement scales with the size/number of layers however I have not benchmarked any larger models to measure the difference.

### DIFF
--- a/server/text_generation_server/models/custom_modeling/qwen2_vl.py
+++ b/server/text_generation_server/models/custom_modeling/qwen2_vl.py
@@ -22,9 +22,11 @@ from torch import nn
 from text_generation_server.utils.import_utils import SYSTEM
 
 if SYSTEM == "ipex":
-    pass
+    import intel_extension_for_pytorch as ipex
 else:
-    pass
+    import flash_attn_2_cuda
+
+import numpy as np
 
 from transformers.activations import ACT2FN
 import torch.nn.functional as F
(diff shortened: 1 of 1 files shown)


server/text_generation_server/models/custom_modeling/qwen2_vl.py
```diff
@@ -117,37 +120,60 @@ def forward(
...
-        key = key.transpose(0, 1)
-        value = value.transpose(0, 1)
-
-        # apply attention
-        attn_output = F.scaled_dot_product_attention(
-            query, key, value, attention_mask, dropout_p=0.0
-        )
-        attn_output = attn_output.transpose(0, 1)
+        # calc maximum sequence length for any batch
+        max_seqlen = torch.max(cu_seqlens[1:] - cu_seqlens[:-1])
```
OlivierDehaene, line 125: It may be interesting to compute this value once in the `Qwen2VisionModel.forward` instead.

drbh: great point, updated in the latest commit to only calculate the `max_seqlen` once!

drbh:
optimistically merging as the `max_seqlen` comment was addressed, these changes are needed in a in progress PR, and the only failing tests are unrelated to this PR (failing elsewhere and unrelated code). 

Watching for regressions and will revet if required!

PR #2723, opened 2024-11-04
Add llama.cpp backend 

mfuntowicz:


Dockerfile.llamacpp
```diff
@@ -0,0 +1,74 @@
...
+ENV DEBIAN_FRONTEND=noninteractive
+
+RUN --mount=type=cache,target=/var/cache/apt,sharing=locked \
+    --mount=type=cache,target=/var/lib/apt,sharing=locked \
+    apt update && \
+    apt upgrade -y && \
+    apt install -y \
+    numactl \
+    openssl \
+    python3.11-dev
```
Hugoch, line 67: I think we are missing `transformers` here for tokenization, so better add it in a venv.
```suggestion
    python3.11-dev \
    python3.11-venv \
    ibgomp1
```

Dockerfile.llamacpp
```diff
@@ -0,0 +1,74 @@
...
+
+COPY --from=builder /usr/src/text-generation-inference/target/release-opt/text-generation-backend-llamacpp /usr/src/text-generation-inference/text-generation-launcher
+COPY --from=builder /usr/src/text-generation-inference/dist /usr/
+
```
Hugoch, line 71:
```suggestion
COPY --from=builder /usr/src/text-generation-inference/backends/llamacpp/requirements.txt requirements.txt
RUN /usr/bin/python3.11 -m venv /usr/src/text-generation-inference/venv
ENV PATH="/usr/src/text-generation-inference/venv/bin:$PATH"
RUN pip3 install --no-cache-dir -r requirements.txt
```

Hugoch: It needs requirements.txt with transformers to be created

backends/llamacpp/src/backend.rs
```diff
@@ -0,0 +1,423 @@
...
+            ));
+        }
+
+        // Allocate the multi-consumer queue to orchestrate all the workers
+        let (backlog_submitter, backlog_receiver) = mpmc_unbounded();
+
+        // Allocate all the workers
+        let cores_allocation = get_cores_allocation(num_cores_per_instance as usize);
+        cores_allocation.iter().for_each(|affinity| {
+            match Self::allocate_worker(path, num_cores_per_instance as u32) {
```
Hugoch, line 183: You do not handle the case where `num_cores_per_instance` is not set as options -> it is then 0 and makes llama.cpp crash

Hugoch (commented):
Thanks @mfuntowicz ! I can't get a correct generation using the Docker so I think something is wrong with tokenizer. I'll dig deeper later today!

backends/llamacpp/src/backend.rs
mfuntowicz: Fixed  in 298367cdfd2f33cc9699b2748f1d8edfbf3c66e1

backends/llamacpp/src/backend.rs
```diff
@@ -0,0 +1,423 @@
...
+
+    // Make sure each instance has some cores available
+    let mut effective_num_cores_per_instance = match num_cores_per_instance {
+        0 => cores_count,
+        _ => num_cores_per_instance,
+    };
+
+    // If we have spare cores, let's see if we can give everyone one more core
+    let num_instances = cores_count / effective_num_cores_per_instance;
+    if cores_count - (num_instances * effective_num_cores_per_instance) >= num_instances {
```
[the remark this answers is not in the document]
mfuntowicz: Fixed 274cfce

backends/llamacpp/src/main.rs
```diff
@@ -0,0 +1,211 @@
...
+        disable_grammar_support,
+        max_client_batch_size,
+        usage_stats,
+    } = args;
+
+    if let Some(Commands::PrintSchema) = command {
+        use utoipa::OpenApi;
+        let api_doc = ApiDoc::openapi().to_pretty_json().unwrap();
+        println!("{}", api_doc);
+        std::process::exit(0);
```
co42, line 121:
```suggestion
        return Ok(());
```

backends/llamacpp/src/backend.rs
```diff
@@ -0,0 +1,496 @@
...
+    mut queue: UnboundedReceiver<(GenerationContext, UnboundedSender<InferResult>)>,
+    backlog: MpmcSender<(GenerationContext, UnboundedSender<InferResult>)>,
+) {
+    // Let's receive incoming requests
+    loop {
+        match queue.recv().await {
+            None => break,
+            Some((ctx, sender)) => {
+                if let Err(e) = backlog.send((ctx, sender)).await {
+                    todo!("What do we do")
```
co42, line 323: Looks like it can only happen id the channel is closed (all receivers are drop), so I guess you can break the loop and just return / exit

co42 (commented):
Nice !

Issue #1802, opened 2024-04-24
Error "Failed to buffer the request body: length limit exceeded" when supplying base64 encoded images greater than 1MB in prompt


akowalsk:
### System Info

text-generation-launcher --env
```
Runtime environment:
Target: x86_64-unknown-linux-gnu
Cargo version: 1.75.0
Commit sha: 2d0a7173d4891e7cd5f9b77f8e0987b82a339e51
Docker label: sha-2d0a717
nvidia-smi:
Wed Apr 24 19:58:49 2024
   +-----------------------------------------------------------------------------------------+
   | NVIDIA-SMI 550.54.15              Driver Version: 550.54.15      CUDA Version: 12.4     |
   |-----------------------------------------+------------------------+----------------------+
   | GPU  Name                 Persistence-M | Bus-Id          Disp.A | Volatile Uncorr. ECC |
   | Fan  Temp   Perf          Pwr:Usage/Cap |           Memory-Usage | GPU-Util  Compute M. |
   |                                         |                        |               MIG M. |
   |=========================================+========================+======================|
   |   0  NVIDIA GeForce RTX 3090        On  |   00000000:01:00.0 Off |                  N/A |
   |  0%   23C    P8             16W /  350W |    8450MiB /  24576MiB |      0%      Default |
   |                                         |                        |                  N/A |
   +-----------------------------------------+------------------------+----------------------+
   |   1  NVIDIA GeForce RTX 3090        On  |   00000000:21:00.0 Off |                  N/A |
   |  0%   25C    P8             20W /  350W |    8418MiB /  24576MiB |      0%      Default |
   |                                         |                        |                  N/A |
   +-----------------------------------------+------------------------+----------------------+
```

model info
```
{
  "model_id": "/opt/ml/checkpoint/llava-v1.6-mistral-7b-hf",
  "model_sha": null,
  "model_dtype": "torch.float16",
  "model_device_type": "cuda",
  "model_pipeline_tag": null,
  "max_concurrent_requests": 128,
  "max_best_of": 2,
  "max_stop_sequences": 4,
  "max_input_length": 24576,
  "max_total_tokens": 32768,
  "waiting_served_ratio": 1.2,
  "max_batch_total_tokens": 65536,
  "max_waiting_tokens": 20,
  "max_batch_size": null,
  "validation_workers": 2,
  "max_client_batch_size": 4,
  "version": "2.0.1",
[...]

PR #2726, opened 2024-11-05
feat: add payload limit

OlivierDehaene:
Related to #1802

### DIFF
--- a/backends/trtllm/src/main.rs
+++ b/backends/trtllm/src/main.rs
@@ -62,6 +62,8 @@ struct Args {
     executor_worker: PathBuf,
     #[clap(default_value = "on", long, env)]
     usage_stats: usage_stats::UsageStatsLevel,
+    #[clap(default_value = "2000000", long, env)]
+    payload_limit: usize,
 }
 
 async fn get_tokenizer(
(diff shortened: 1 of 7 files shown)


server/text_generation_server/models/flash_causal_lm.py
```diff
@@ -962,9 +962,9 @@ def prepare_for_prefill(self):
         self.input_lengths_tensor = torch.tensor(
             self.input_lengths, dtype=torch.int32, device=device
         )
-        self.cu_seqlen_prefill = torch.nn.functional.pad(
-            torch.cumsum(self.input_lengths_tensor, dim=0), (1, 0)
-        ).to(torch.int32)
+        cu_seqlen_prefill = self.input_lengths_tensor.new_zeros(len(self) + 1)
```
danieldk, line 965: Are these changes in the same PR intentional?

backends/trtllm/src/main.rs
```diff
@@ -287,6 +290,7 @@ async fn main() -> Result<(), TensorRtLlmBackendError> {
         tokenizer_name,
         tokenizer_config_path,
         revision,
+        false,
```
Narsil, line 293: Is this intentional ?

PR #2733, opened 2024-11-08
Support continue final message

drbh:
This PR adds the `continue_final_message` param to chat requests to allow users to continue the last message. This is useful when prefilling a response from the model. Please see https://huggingface.co/docs/transformers/main/en/chat_templating#what-does-continuefinalmessage-do for details

### DIFF
--- a/integration-tests/models/test_continue_final_message.py
+++ b/integration-tests/models/test_continue_final_message.py
@@ -1,0 +1,81 @@
+import pytest
+import requests
+
+
+@pytest.fixture(scope="module")
+def llama_continue_final_message_handle(launcher):
+    with launcher(
+        "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+        num_shard=1,
+        disable_grammar_support=False,
+        use_flash_attention=False,
+    ) as handle:
+        yield handle
+
+
+@pytest.fixture(scope="module")
+async def llama_continue_final_message(llama_continue_final_message_handle):
+    await llama_continue_final_message_handle.health(300)
+    return llama_continue_final_message_handle.client
+
+
+def test_llama_completion_single_prompt(
+    llama_continue_final_message, response_snapshot
+):
+    response = requests.post(
+        f"{llama_continue_final_message.base_url}/v1/chat/completions",
+        json={
+            "model": "tgi",
+            "messages": [
+                {"role": "system", "content": "system message"},
+                {"role": "user", "content": "user message"},
+                {"role": "assistant", "content": "assistant message"},
+            ],
+            "max_tokens": 30,
+            "stream": False,
+            "seed": 1337,
+            "continue_final_message": False,
+        },
+        headers=llama_continue_final_message.headers,
+        stream=False,
+    )
+    response = response.json()
+    print(response)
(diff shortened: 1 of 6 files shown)


integration-tests/models/test_continue_final_message.py
```diff
@@ -0,0 +1,81 @@
...
+import requests
+
+
+@pytest.fixture(scope="module")
+def llama_continue_final_message_handle(launcher):
+    with launcher(
+        "TinyLlama/TinyLlama-1.1B-Chat-v1.0",
+        num_shard=1,
+        disable_grammar_support=False,
+        use_flash_attention=False,
```
Narsil, line 11: Remove those.

integration-tests/models/test_continue_final_message.py
```diff
@@ -0,0 +1,81 @@
...
+            "model": "tgi",
+            "messages": [
+                {"role": "system", "content": "system message"},
+                {"role": "user", "content": "user message"},
+                {"role": "assistant", "content": "assistant message"},
+            ],
+            "max_tokens": 30,
+            "stream": False,
+            "seed": 1337,
+            "continue_final_message": False,
```
Narsil, line 37: Why can't we assume that having a final assistant message means that users what continuation ?

There's no use case where the LLM produces a user output right ?

drbh: thats an interesting idea, currently a conversation could hypothetically contain back to back assistant messages (assuming the template allows it)

ie.
```
<system>
<user>
<assistant>
<assistant>
```

I could see this pattern being used when injecting some information for the following assistant response. 

```python
"messages": [
    {"role": "system", "content": "system message"},
    {"role": "user", "content": "whats ticket 10 about?"},
    {"role": "assistant", "content": "Ticket Information:\nTicket ID: 10\nTicket Title: Ticket 10\nTicket Description: This is a test ticket\nTicket Status: Open\nTicket Priority: High\nTicket Type: Bug\nTicket Assignee: John Doe\nTicket Reporter: Jane Doe\nTicket Created: 2022-01-01 00:00:00\nTicket Updated: 2022-01-01 00:00:00\n"},
],
```
and the user may expect a new message, rather than a continuation of the last one.

however, it would be nice to disallow this pattern to avoid the extra argument, and always just continue the last message if its an assistant (or other heuristic)

Narsil: Does the pattern exist in the wild or not ? If there's no evidence, let's not guess about what might become, and focus instead on what is.

We can add a flag when the current behavior proposal needs to be added (for instance when there's a new use case).

integration-tests/models/test_continue_final_message.py
drbh: removed in latest commit

integration-tests/models/test_continue_final_message.py
drbh: great point! I've removed the `continue_final_message` and prefer assuming if the final message is an assistant we will continue it

danieldk (approved):
Looks good to me, but I am not familiar with this part of the code.

PR #2756, opened 2024-11-18
Enable qwen2vl video

drbh:
This PR is a work in progress that explores adding support for video inputs with Qwen2-VL. Thank you @mfarre for getting this effort started.

TODOS
- [X] suport `video_url`s
- [X] fetch video contents in router
- [X] update protobufs to support video chunks
- [ ] handle padding video token inputs
- [ ] tokenize video bytes
- [ ] integrate video logic with vision model (update position ids)


router/src/validation.rs
mfarre, line 1: I think it would help if at this point we already sample the video at 1fps and resize the frames to 360x420 if those are bigger.
We only convert to tokens the frames that we end up consuming for inference. qwen has this specific 'smart' logic and I think other models will have other logic. Where do you think is best to place the smart logic selection of qwen? I don't find it bad to be already in validation if afterwards we can launch other video models with the qwen frame selection logic, it can actually be interesting.

Narsil:
It still doesn't work @drbh

server/text_generation_server/models/vlm_causal_lm.py
```diff
@@ -190,6 +232,30 @@ def batch_tokenized_inputs(
...
+                                i * bytes_per_frame : (i + 1) * bytes_per_frame
+                            ]
+                            frame = frame.reshape(
+                                chunk.video.height, chunk.video.width, 3
+                            )
+                            frames.append(frame)
+
+                        video_frame_buf = np.stack(frames)
+                        frame_nchw_tensor = torch.from_numpy(video_frame_buf).permute(
+                            0, 3, 1, 2
```
Narsil, line 256: This can and should be done without reallocating once.

server/text_generation_server/models/vlm_causal_lm.py
```diff
@@ -212,9 +291,10 @@ def batch_tokenized_inputs(
                         processor, image_inputs, config, image_id
                     )
                     image_id += 1
+                elif chunk_type == "video":
+                    full_text += video_text_replacement(processor, video_inputs, config)
```
Narsil, line 295: else: raise Error ( we're starting to have many types of chunks)

.github/workflows/autodocs.yaml
```diff
@@ -20,7 +20,7 @@ jobs:
     - name: Install Protocol Buffers compiler
       run: |
         sudo apt-get update
-        sudo apt-get install -y protobuf-compiler libprotobuf-dev
+        sudo apt-get install -y protobuf-compiler libprotobuf-dev clang libavcodec-dev libavfilter-dev libavdevice-dev libavformat-dev libavutil-dev pkg-config
```
Narsil, line 23: No dev packages please (we shouldn't need them most likely).

Why is clang in there?

.github/workflows/tests.yaml
```diff
@@ -43,7 +43,9 @@ jobs:
       - name: Install
         run: |
           sudo apt update
-          sudo apt install python3.11-dev -y
+          sudo apt install python3.11-dev python3.11-venv python3-pip clang libavcodec-dev libavfilter-dev libavdevice-dev libavformat-dev libavutil-dev pkg-config -y
+          export PKG_CONFIG_PATH=$PKG_CONFIG_PATH:/usr/lib/x86_64-linux-gnu/pkgconfig
```
Narsil, line 47: Not env shenanigans

flake.nix
```diff
@@ -115,15 +115,17 @@
...
+                cargo
                 client
+                clippy
+                ffmpeg
+                launcher
                 openssl.dev
                 pkg-config
-                cargo
+                router
+                rustPlatform.bindgenHook
```
Narsil, line 126: No bindgenhook.

ffmpeg is ok, but please let's keep the diff readable (no sorting).

integration-tests/models/test_flash_qwen2_vl_video.py
```diff
@@ -0,0 +1,75 @@
...
+
+
+@pytest.fixture(scope="module")
+def qwen2_vl_handle(launcher):
+    with launcher(
+        "Qwen/Qwen2-VL-7B-Instruct",
+        max_input_length=10_000,
+        max_batch_prefill_tokens=10_000,
+        max_total_tokens=10_001,
+        cuda_graphs=[0],
```
Narsil, line 13: Remove all of these. We shouldn't need them.
We NEED to make sure users don't have to set anything.

integration-tests/models/test_flash_qwen2_vl_video.py
```diff
@@ -0,0 +1,75 @@
...
+        json={
+            "model": "tgi",
+            "messages": [
+                {
+                    "role": "user",
+                    "content": [
+                        {
+                            "type": "video_url",
+                            "video_url": {
+                                "url": "https://test-videos.co.uk/vids/bigbuckbunny/mp4/h264/360/Big_Buck_Bunny_360_10s_1MB.mp4"
```
Narsil, line 38: Let's put a URL under our control please.
Cehck what we have for images.

integration-tests/models/test_flash_qwen2_vl_video.py
```diff
@@ -0,0 +1,75 @@
...
+                    break
+                print("=", line)
+                try:
+                    response = json.loads(line)
+                    # print(response)
+                    last_response = response
+                    full_text += response["choices"][0]["delta"]["content"]
+                except json.JSONDecodeError:
+                    pass
+
```
Narsil, line 74: Where's the actual assert?

router/src/lib.rs
```diff
@@ -1229,6 +1230,9 @@ impl From<Message> for TextMessage {
                     .map(|chunk| match chunk {
                         MessageChunk::Text { text } => text,
                         MessageChunk::ImageUrl { image_url } => format!("![]({})", image_url.url),
+                        MessageChunk::VideoUrl { video_url } => {
+                            format!("<video>({})", video_url.url)
```
Narsil, line 1234: Let's remove this altogther. We don't want the serialized shenanigan anymore.

router/Cargo.toml
```diff
@@ -74,3 +77,4 @@ default = ["ngrok"]
 ngrok = ["dep:ngrok"]
 google = []
 kserve = []
+video = ["ffmpeg-next", "mp4parse", "tempfile"]
```
Narsil, line 80: Why so many deps?

router/src/validation.rs
```diff
@@ -645,13 +808,64 @@ fn prepare_input<T: TokenizerTrait>(
...
+            // handle video content first
+            for chunk in VIDEO_RE.find_iter(&inputs) {
+                let chunk_start = chunk.start();
+                let chunk_end = chunk.end();
+                if chunk_start != start {
+                    input_chunks.push(Chunk::Text(inputs[start..chunk_start].to_string()));
+                    tokenizer_query.push_str(&inputs[start..chunk_start]);
+                }
+                let processed_video = match config {
+                    Idefics | Mllama | Idefics2(_) | Paligemma(_) | LlavaNext(_) => {
```
Narsil, line 832: Why ? They do not support video for now. So let's not do anything.

router/src/validation.rs
```diff
@@ -645,13 +808,64 @@ fn prepare_input<T: TokenizerTrait>(
 ) -> Result<(tokenizers::Encoding, Vec<Chunk>), ValidationError> {
     use Config::*;
     static RE: Lazy<Regex> = Lazy::new(|| Regex::new(r"!\[\]\([^\)]*\)").unwrap());
+    // Add video regex
+    static VIDEO_RE: Lazy<Regex> =
```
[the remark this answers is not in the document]
Narsil: No regex + No string shenanigans.

router/src/validation.rs
```diff
@@ -536,6 +545,140 @@ fn format_to_mimetype(format: ImageFormat) -> String {
...
+pub fn fetch_video(
+    _input: &str,
+    _target_width: u32,
+    _target_height: u32,
+) -> Result<ProcessedVideo, ValidationError> {
+    Err(ValidationError::VideoNotSupported)
+}
+
+#[cfg(feature = "video")]
+pub fn fetch_video(
```
Narsil, line 558: This overall seems super convoluted.

- Why do we need a file for something that should stay in RAM ?
- We are re-encoding the video here, this is fine if we are agressively trimming the output video (like removing extra frames). We have to assume ppl are going to send 4k video of 10hour long content.
- We are binding to `ffmpeg` which seems to add a lot of dependency complexity. Why not just depend on `ffmpeg` binary, and be done with it ? Seems much simpler, and we're not looking at all at the contents it seems.

For the tempfile, it *might* be the simplest. But I feel we need to discuss options before doing it that way.

router/src/validation.rs
```diff
@@ -624,6 +767,26 @@ fn image_tokens(
...
+    use Config::*;
+
+    match config {
+        Qwen2Vl(_config) => {
+            let min_frames = 2_f32;
+            let max_frames = 256_f32;
+            // make sure the frames are within the range and are even
+            let nframes = (sampled_frames).max(min_frames).min(max_frames);
+            let nframes = (nframes / 2.0).round() as usize * 2;
+            let num_tokens = nframes * height as usize * width as usize / 1541;
```
Narsil, line 780: Keep everything in usize and use regular division or  `div_ceil`. Should be much simpler.

Where is that `1514` coming from ?

router/src/validation.rs
```diff
@@ -718,6 +952,20 @@ impl ChunksToString for Vec<Chunk> {
...
             }
+            Chunk::Video(Video {
+                data,
+                mimetype,
+                width,
+                height: _,
+                num_frames: _,
+            }) => {
+                // TODO: revisit if we should limit video support to v3 - to avoid sending very large base64 strings
+                let encoded = STANDARD.encode(data);
```
Narsil, line 963: Let's revisit now.

server/text_generation_server/models/custom_modeling/qwen2_vl.py
```diff
@@ -424,33 +425,50 @@ def get_position_ids(
...
+                # only copy the sum of the image and video tokens GPU<->CPU
                 image_count = (vision_tokens == self.image_token_id).sum().item()
+                video_count = (vision_tokens == self.video_token_id).sum().item()
 
                 current_pos = 0
-                for _ in range(image_count):
-                    # copy the value position of the next image token from GPU<->CPU
-                    next_image_pos = (
-                        (input_ids[current_pos:] == self.image_token_id)
+                for _ in range(image_count + video_count):
```
Narsil, line 445: This looks pretty bloated.

It was like that before but it's still seems quite bloated.
There should be 1 replace for all images, 1 replace for all videos.

Not super urgent if this works (as it's limited to qwen).

Also seems everything named image was renamed to video.. Sounds like better names can be found.

server/text_generation_server/models/vlm_causal_lm.py
drbh: good point, the latest commit updates this to avoid the loop and prefer a vectorized method with numpy and torch.

server/text_generation_server/models/vlm_causal_lm.py
drbh: good catch, updated

integration-tests/models/test_flash_qwen2_vl_video.py
drbh: agreed removed in latest commit

PR #2766, opened 2024-11-20
fix: set outlines version to 0.1.3 to avoid caching serialization issue

drbh:
This PR fixes the outline version to `0.1.3` to avoid a bug introduced in `0.1.4` when attempting to cache the tokenizer

related: https://github.com/dottxt-ai/outlines/issues/1274 and https://github.com/dottxt-ai/outlines-core/issues/95


drbh:
merging this hot fix as it resolves the failing tests on main and avoid a runtime crash

PR #2754, opened 2024-11-18
Incomplete generation stream fix

sywangyi:
server launch baichuan-inc/Baichuan2-13B-Chat 

using vllm to test tgi backend, command like
python benchmark_serving.py --backend tgi --model baichuan-inc/Baichuan2-13B-Chat --trust-remote-code --dataset-name random --random-input-len 1024 --random-output-len 128 --request-rate inf --num-prompts 100 --endpoint /generate_stream --host 0.0.0.0 --port 80 --seed 10

server hangs, log like
2024-11-18T08:46:35.673546Z ERROR text_generation_router::server: router/src/server.rs:656: Incomplete generation stream
2024-11-18T08:46:35.673576Z ERROR text_generation_router::server: router/src/server.rs:656: Incomplete generation stream

PR #2770, opened 2024-11-21
fix: incomplete generations w/ single tokens generations and models that did not support chunking

OlivierDehaene:
Related to #2754


danieldk (approved):
Not very familiar with this, but looks good.

Issue #2440, opened 2024-08-21
[Volta] [No flash attention] Llama 3.1 8B Instruct failed to start - "< not supported between instances of 'NoneType' and 'int'"


ladi-pomsar:
### System Info

Hi everyone, when trying to update from Llama 3 8B Instruct to Llama 3.1 8B Instruct, I noticed a crash:

```bash
Args {
    model_id: "meta-llama/Meta-Llama-3.1-8B-Instruct",
    revision: None,
    validation_workers: 2,
    sharded: Some(
        false,
    ),
    num_shard: None,
    quantize: None,
    speculate: None,
    dtype: None,
    trust_remote_code: false,
    max_concurrent_requests: 128,
    max_best_of: 2,
    max_stop_sequences: 4,
    max_top_n_tokens: 5,
    max_input_tokens: Some(
        1500,
    ),
    max_input_length: None,
    max_total_tokens: None,
    waiting_served_ratio: 0.3,
    max_batch_prefill_tokens: None,
    max_batch_total_tokens: None,
    max_waiting_tokens: 20,
    max_batch_size: None,
    cuda_graphs: None,
    hostname: "llm2.internal",
    port: 80,
    shard_uds_path: "/tmp/text-generation-server",
    master_addr: "localhost",
    master_port: 29500,
    huggingface_hub_cache: None,
    weights_cache_override: None,
    disable_custom_kernels: false,
    cuda_memory_fraction: 1.0,
    rope_scaling: None,
    rope_factor: None,
    json_output: false,
    otlp_endpoint: None,
    otlp_service_name: "text-generation-inference.router",
    cors_allow_origin: [],
    api_key: None,
    watermark_gamma: None,
    watermark_delta: None,
    ngrok: false,
    ngrok_authtoken: None,
    ngrok_edge: None,
    tokenizer_config_path: None,
    disable_grammar_support: false,
    env: false,
    max_client_batch_size: 4,
    lora_adapters: None,
    usage_stats: On,
}
2024-08-21T12:39:33.092070Z  INFO hf_hub: Token file not found "/data/token"    
2024-08-21T12:39:33.092172Z  INFO text_generation_launcher: Default `max_total_tokens` to 4096
2024-08-21T12:39:33.092177Z  INFO text_generation_launcher: Default `max_batch_prefill_tokens` to 1550
2024-08-21T12:39:33.092179Z  INFO text_generation_launcher: Using default cuda graphs [1, 2, 4, 8, 16, 32]
[...]

PR #2774, opened 2024-11-22
fix: only use eos_token_id as pad_token_id if int

dvrogozh:
LLama 3 has a list of values as `eos_token_id`:
```
  "['<|end_of_text|>', '<|eom_id|>', '<|eot_id|>']"
```
This breaks tokenizer since it expects single value. This commit uses `tokenizer.eos_token_id` instead in such a case.

Fixes: #2440

CC: @Narsil @zucchini-nlp @ladi-pomsar


zucchini-nlp:
@dvrogozh Cool, looks good to me. We also need a review from TGI team

dvrogozh:
@Narsil, @OlivierDehaene : please, help to review.

Narsil (approved):
LGTM

PR #2786, opened 2024-11-29
Install `text-generation-server` from `poetry.lock` export

alvarobartt:
# What does this PR do?

This PR installs the `text-generation-server` Python requirements from an exported `requirements.txt`-like file generated out of the `poetry.lock`, to be able to reuse the generated lock with the pinned dependencies and avoid dependency issues or conflicts when building outdated Dockerfiles in case any dependency is loose, as any potential conflict is always solved before the release and the lock is always working.

## Before submitting

- [x] Did you read the [contributor guidelines](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests), Pull Request section?
- [ ] Was this discussed/approved via a Github issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [ ] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

## Who can review?

cc @OlivierDehaene

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -199,7 +199,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
 COPY --from=pytorch-install /opt/conda /opt/conda
 
 # Export text-generation-server Python requirements from poetry lock file
-FROM poetry-install AS poetry-requirements
+FROM conda-install AS poetry-requirements
 
 COPY server/poetry.lock poetry.lock
 COPY server/pyproject.toml pyproject.toml
--- a/Dockerfile
+++ b/Dockerfile
@@ -211,7 +211,7 @@ RUN pip install poetry && poetry export -f requirements.txt  \
 FROM conda-install AS base
 
 # Copy the requirements file generated from the poetry lock
-COPY --from=poetry-requirements /usr/src/requirements_poetry.txt requirements_poetry.txt
+COPY --from=poetry-requirements /usr/src/requirements_poetry.txt server/requirements_poetry.txt
 
 # Copy build artifacts from flash attention builder
 COPY --from=flash-att-builder /usr/src/flash-attention/build/lib.linux-x86_64-cpython-311 /opt/conda/lib/python3.11/site-packages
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -198,6 +198,21 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
 # Copy conda with PyTorch installed
 COPY --from=pytorch-install /opt/conda /opt/conda
 
+# Export text-generation-server Python requirements from poetry lock file
+FROM poetry-install AS poetry-requirements
```
alvarobartt, line 202:
```suggestion
FROM conda-install AS poetry-requirements
```

Dockerfile
```diff
@@ -233,7 +250,8 @@ COPY server/Makefile server/Makefile
 RUN cd server && \
     make gen-server && \
     pip install -r requirements_cuda.txt && \
-    pip install ".[attention, bnb, accelerate, compressed-tensors, marlin, moe, quantize, peft, outlines]" --no-cache-dir && \
+    pip install -r requirements_poetry.txt --no-cache-dir && \
```
Narsil, line 253: What's the issue again ? Why are we not using poetry directly ?

Also maybe we should take the opportunity to switch to `uv` ? https://docs.astral.sh/uv/pip/compile/#locking-requirements I have no idea, but it seems like the poetry lock isn't locking tightly enough, maybe ?

alvarobartt: Sorry for the delay on this PR, and happy new year BTW! So the issue is that since we're not using `poetry` as the pip installer the locked dependencies are not used leading to potential issues when partners as e.g. Google build the containers ahead in time, as those dependencies are installed from the `pyproject.toml` rather than the lock.

I know this is not an issue impacting directly TGI per se, but for robustness and reproducibility we should try to lock those and install those from a lock. The installation from the lock can either be via `poetry` (the current approach), or switching to `uv` instead and installing from the `uv.lock`.

Maybe we should start an internal conversation and decide what's best, happy to help whatever the solution / proposal is!

alvarobartt: Thanks, closed in favour of https://github.com/huggingface/text-generation-inference/pull/2919

Issue #2780, opened 2024-11-26
I encountered the same issue while using `baichuan2-13B-chat`..


Lacacy:
I encountered the same issue while using `baichuan2-13B-chat`..
I extracted the chat parameters from baichuan2's [generation_config.json](https://huggingface.co/baichuan-inc/Baichuan2-13B-Chat/blob/main/generation_config.json), and when I call the tgi interface, the result is as follows. 
![image](https://github.com/huggingface/text-generation-inference/assets/46644537/7e7f561c-f31e-43a0-8696-7ecd65fba9c5)

When I invoke the `chat` method, the result is as follows. 
![image](https://github.com/huggingface/text-generation-inference/assets/46644537/fb724024-b02e-443f-80e7-221d9e7a487b)

Here are the deployment parameters for the tgi.

``` bash
--max-batch-prefill-tokens 4096 --max-input-length 4096 --max-total-tokens 4608
```

_Originally posted by @zTaoplus in https://github.com/huggingface/text-generation-inference/issues/981#issuecomment-1734882450_

PR #2798, opened 2024-12-03
Update tensor_parallel.py

Lacacy:
Resolve the issue of abnormal conversation performance in the Baichuan large model.

# Fix the bug in the norm_head adaptation for Baichuan. 

Fixes https://github.com/huggingface/text-generation-inference/issues/2780

https://huggingface.co/baichuan-inc/Baichuan2-13B-Chat/blob/main/modeling_baichuan.py#:~:text=self.weight.data%20%3D%20nn.functional.normalize(self.weight)

image

@OlivierDehaene OR @Narsil

### DIFF
--- a/server/text_generation_server/layers/tensor_parallel.py
+++ b/server/text_generation_server/layers/tensor_parallel.py
@@ -66,6 +66,11 @@ class TensorParallelHead(SuperLayer):
             weight = weights.get_tensor(f"{prefix}.weight")
             should_gather = False
 
+        if config.model_type == "baichuan":
+            # Resolve the issue of abnormal conversation performance in the Baichuan large model.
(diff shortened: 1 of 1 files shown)


Narsil:
We cannot really accept this. 
This is a bug in Baichuan weights, not in our code.

The issue with your proposed fix is that we support tensor parallelism (TP), which means `weight` values will depend on what TP value you're using, leading to potentially even more massive discrepancies.
The "true" fix in that sense would be to load the entire weight, normalize it, and then split it across GPU, but it will lead to other issues, the first of which will be excess of VRAM usage, which can cause unwanted OOMs.
Baichuan should fix their weights (unless there's a valid reason to keep the unnormalized weights, but I don't think there's one).

PR #2815, opened 2024-12-10
Flash decoding kernel adding and prefill-chunking and prefix caching enabling in intel cpu/xpu

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


sywangyi:
@Narsil @OlivierDehaene please help review it.

Dockerfile_intel
```diff
@@ -222,9 +224,9 @@ COPY --from=builder /usr/src/target/release-opt/text-generation-router /usr/loca
 COPY --from=builder /usr/src/target/release-opt/text-generation-launcher /usr/local/bin/text-generation-launcher
 
 FROM ${PLATFORM} AS final
-ENV ATTENTION=paged
-ENV PREFIX_CACHING=0
-ENV PREFILL_CHUNKING=0
+ENV ATTENTION=flashdecoding
```
Narsil, line 227: Can we use a different name. It doesn't seem to be the same kernel at all.
Different block size, different cache layout.

Other parts will fail with this, and it makes the reasoning too complex.

`flashdecoding-ipex` maybe ?

yao-matrix: @Narsil , i have different thinking on this. From me, I will treat them as same algorithm, they both implemented flash attention with paged kv-cache. Algorithm-wise they are same, but since different architecture and hardware configuration like tensor engine number, tensor engine mat-mul size(MxKxN), so different hardwares may have different optimal layout or BLOCK_SIZE.  So, from my understanding, flashdecoding is about its algorithm and some state(e.g. kv cache) may each hardware has its own optimal. WDYT

yao-matrix: "Other parts will fail with this, and it makes the reasoning too complex."

for above, could you help elaborate more on this concern? we can think how to minimize the complex

Narsil: Yes we need to add the entry on the rust side with the proper block size and all.

I think we can avoid the detection part because are burning the ATTENTION value in the dockerfile.

Narsil (approved):
LGTM

PR #2824, opened 2024-12-11
fix facebook/opt-125m not working issue

sywangyi:
crash log like

RuntimeError: weight decoder.embed_tokens.weight does not exist rank=0

opt-125m has prefix "model" while opt-6.7b does not have prefix

### DIFF
--- a/server/text_generation_server/models/custom_modeling/opt_modeling.py
+++ b/server/text_generation_server/models/custom_modeling/opt_modeling.py
@@ -99,7 +99,7 @@ class OPTLearnedPositionalEmbedding(nn.Module):
         self.offset = 2
         self.weight = nn.Parameter(
             weights.get_tensor(
-                f"{prefix + '.' if prefix else ''}decoder.embed_positions.weight"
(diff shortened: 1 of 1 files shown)


sywangyi:
@Narsil please help review

sywangyi:
@drbh

Narsil (commented):
Any particular reason this model is necessary ?

It's a dummy model released 2 years ago.. I don't think making any modification here is worth it. It's not flash, not anything. There are a lot of such impossible to track weird naming issues on legacy models on the hub.

sywangyi:
I try this model because I use https://github.com/EleutherAI/lm-evaluation-harness to check tgi accuracy. and in readme.
like 

We also support using your own local inference server with servers that mirror the OpenAI Completions and ChatCompletions APIs.
lm_eval --model local-completions --tasks gsm8k --model_args model=facebook/opt-125m,base_url=http://{yourip}:8000/v1/completions,num_concurrent=1,max_retries=3,tokenized_requests=False,batch_size=16

PR #2825, opened 2024-12-11
Enable FP8 Per-Tensor Scales and Integrate Marlin/MoE Kernels Repo for ROCm

mht-sharma:
# What does this PR do?

Fixes following issues:
1. Use FP8 per tensor scales for ROCM (wherever possible) to leverage `scaled_mm` efficiently.
2. Fixes random NAN issues with custom PA in rocm
3. Uses `marlin_kernels` and `moe_kernels` repositories for ROCM

## Before submitting
      Pull Request section?
      to it if that's the case.


Dockerfile_amd
```diff
@@ -267,6 +267,24 @@ COPY server/exllamav2_kernels/ .
...
 RUN python setup.py build
 
+FROM kernel-builder AS marlin-kernels
+WORKDIR /usr/src
+ENV MARLIN_KERNELS_BRANCH=v0.3.6
+ENV VLLM_TARGET_DEVICE=rocm
+RUN git clone https://github.com/danieldk/marlin-kernels.git && \
+    cd marlin-kernels && \
+    git checkout ${MARLIN_KERNELS_BRANCH} && \
+    python setup.py install
```
danieldk, line 277: Not sure how hard it'll be, but it would be nice to have precompiled wheels in the future to cut down the build times.

server/text_generation_server/layers/fp8.py
```diff
@@ -79,11 +82,37 @@ def normalize_e4m3fn_to_e4m3fnuz(
     return weight, weight_scale, input_scale
 
 
+def per_tensor_dequantize(
+    tensor: torch.Tensor, inv_scale: Union[float, torch.Tensor]
+) -> torch.Tensor:
+    fake_qweight = tensor.to(torch.float16)
```
danieldk, line 88: I think this should be the model dtype? (The quantized numbers could represent values that are only in `bfloat16`'s range.)

server/text_generation_server/layers/fp8.py
```diff
@@ -259,16 +297,18 @@ def get_weights_row(self, weights: "Weights", prefix: str):
...
+
             input_scale = None
             if weights.has_tensor(f"{prefix}.input_scale"):
-                input_scale = weights.get_tensor(
-                    f"{prefix}.input_scale", to_dtype=False
-                ).reshape(-1)
+                input_scale = (
+                    weights.get_tensor(f"{prefix}.input_scale", to_dtype=False)
+                    .reshape(-1)
+                    .max()
```
danieldk, line 310: Is this safe? If `input_scale` is a vector, doesn't it need to be dequantized with the original scales first and then requantized with the new max?

server/text_generation_server/layers/fp8.py
mht-sharma: Yes made the changes

server/text_generation_server/layers/fp8.py
mht-sharma: I think this should be fine since it is the input_scale. Dequantization doesn’t make sense in this context because the input will be unquantized and will simply be quantized using this representative scale.

danieldk (approved):
LGTM, would be good to get a successful CI run before merging.

Issue #2838, opened 2024-12-13
Model warmup fails after adding Triton indexing kernels


YaserJaradeh:
### System Info

I was using v2.3.1 via docker and everything was working. When I updated to later versions including the latest my TGI doesn't start due to an error:

```
2024-12-12T14:26:52.973549Z  INFO hf_hub: Token file not found "/data/token"    
2024-12-12T14:26:54.846408Z  INFO text_generation_launcher: Forcing attention to 'flashdecoding' because head dim is not supported by flashinfer, also disabling prefix caching
2024-12-12T14:26:54.846426Z  INFO text_generation_launcher: Using attention flashdecoding - Prefix caching 0
2024-12-12T14:26:54.846433Z  INFO text_generation_launcher: Sharding model on 2 processes
2024-12-12T14:26:54.931439Z  INFO text_generation_launcher: Default `max_batch_prefill_tokens` to 30821
2024-12-12T14:26:54.931470Z  INFO text_generation_launcher: Using default cuda graphs [1, 2, 4, 8, 16, 32]
2024-12-12T14:26:54.931727Z  INFO download: text_generation_launcher: Starting check and download process for microsoft/Phi-3.5-mini-instruct
2024-12-12T14:26:57.914690Z  INFO text_generation_launcher: Files are already present on the host. Skipping download.
2024-12-12T14:26:58.250499Z  INFO download: text_generation_launcher: Successfully downloaded weights for microsoft/Phi-3.5-mini-instruct
2024-12-12T14:26:58.251011Z  INFO shard-manager: text_generation_launcher: Starting shard rank=0
2024-12-12T14:26:58.251055Z  INFO shard-manager: text_generation_launcher: Starting shard rank=1
2024-12-12T14:27:00.870304Z  INFO text_generation_launcher: Using prefix caching = False
2024-12-12T14:27:00.870362Z  INFO text_generation_launcher: Using Attention = flashdecoding
2024-12-12T14:27:06.425419Z  INFO text_generation_launcher: Using prefill chunking = True
2024-12-12T14:27:06.535239Z  INFO text_generation_launcher: Server started at unix:///tmp/text-generation-server-0
2024-12-12T14:27:06.536669Z  INFO text_generation_launcher: Server started at unix:///tmp/text-generation-server-1
2024-12-12T14:27:06.572585Z  INFO shard-manager: text_generation_launcher: Shard ready in 8.307980962s rank=0
[...]

PR #2848, opened 2024-12-16
Update Dockerfile to use devel image for compatibility

YaserJaradeh:
# What does this PR do?

The TGI server fails to start due to missing Python headers during the compilation of Triton indexing kernels. The solution is to change the base image to nvidia/cuda:12.4.1-devel-ubuntu22.04 to match the builder image, ensuring the necessary headers are included.
This change increases the image size but resolves the startup issue.

Fixes # (issue)
This pull request addresses the issue [#2838](https://github.com/huggingface/text-generation-inference/issues/2838) 

## Before submitting
      Pull Request section?
      to it if that's the case.
- [x] Did you make sure to update the documentation with your changes? Here are the

## Who can review?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -174,7 +174,7 @@ COPY server/Makefile-flashinfer Makefile
 RUN make install-flashinfer
 
 # Text Generation Inference base image
-FROM nvidia/cuda:12.1.0-base-ubuntu22.04 AS base
+FROM nvidia/cuda:12.4.1-devel-ubuntu22.04 AS base
 
 # Conda env
 ENV PATH=/opt/conda/bin:$PATH \


YaserJaradeh:
@Narsil

scriptator:
This PR solves my issue descibed [here](https://github.com/huggingface/text-generation-inference/issues/2838#issuecomment-2586691348), thank you!

scriptator:
I just noticed that you changed the implementation between me building the image and my previous message. Should I test again?

YaserJaradeh:
> I just noticed that you changed the implementation between me building the image and my previous message. Should I test again?

Not yet! still broken I just pushed to try to build it on my server but so far it is not working! will ping you when it is 💯

YaserJaradeh:
@KreshLaDoge I tried multiple variants of using only `python3.11-dev` but that didn't work! I also tried copying the headers and libraries from the pytorch building stage into the final image and that didn't work either. Furthermore, I tried with a combination of `python3.11-dev, cuda CLI, and build-essential` and also I wasn't able to get working!

Any ideas about how to get it to work, or reduce the size of the final image?

scriptator:
Do you have a way of reproducing the error more quickly than building the entire image from scratch every time? And do you manage to get to the exact compiler error? If you could share that, I might take a look.

Narsil:
As previously suggested, the fix cannot be accepted as-is. It bloats the image way too much (20GB vs 12GB).

First we need to reproduce locally, then figure out why the hell triton wants to recompile something (not a kernel obviously, since it tries `gcc`)

danieldk:
Besides the above it would also be useful to post the contents of the following files on the host system:

`/var/run/cdi/nvidia-container-toolkit.json`

`/etc/cdi/nvidia-container-toolkit.json`

(Or if they don't exist, check in these directories for a JSON file for nvidia.)

PR #2858, opened 2024-12-19
Efficient Transformers backend support

Cyrilvallez:
# What does this PR do?

Initial draft to support `transformers` as a (more efficient) backend in TGI. https://github.com/huggingface/transformers/pull/35235 added support for a bunch of models already, and more will come progressively.

However, I do need some guidance on how to best support multi-gpu setups 🤗

cc @OlivierDehaene @Narsil

### DIFF
--- a/server/text_generation_server/models/__init__.py
+++ b/server/text_generation_server/models/__init__.py
@@ -12,7 +12,7 @@ import os
 
 from loguru import logger
 from transformers.configuration_utils import PretrainedConfig
-from transformers.models.auto import modeling_auto, modeling_task
+from transformers.models.auto import modeling_auto
 from huggingface_hub import hf_hub_download, HfApi
 from typing import Optional, List, Dict
 from pathlib import Path
@@ -380,12 +380,14 @@ def get_model(
         logger.info(
             "TGI's flash enabled models could either not be loaded or are disabled, using Transformers fallback."
         )
-        try:
-            transformers_model_class = getattr(transformers, modeling_auto.MODEL_FOR_CAUSAL_LM_MAPPING_NAMES[model_type])
-        except KeyError:
-            transformers_model_class = modeling_task.AutoForCausalLM
 
-        if transformers_model_class._supports_flash_attn_2:
+        transformers_model_class = getattr(transformers, modeling_auto.MODEL_FOR_CAUSAL_LM_MAPPING_NAMES[model_type])
+        # Ugly check but works in the meantime
+        model_path = os.path.join(os.path.dirname(transformers.__file__), "models", model_type, f"modeling_{model_type}.py")
+        with open(model_path) as file:
+            has_fa2_class = f"FlashAttention2(" in file.read()
+
+        if transformers_model_class._supports_flash_attn_2 and not has_fa2_class:
             logger.info(
                 f"Transformers' {model_type} implementation supports ragged tensors format (single dimension for "
                 "batch and sequence length). All TGI's batching/caching optimizations are enabled."
--- a/server/text_generation_server/models/__init__.py
+++ b/server/text_generation_server/models/__init__.py
@@ -20,6 +20,7 @@ from pathlib import Path
 from text_generation_server.utils.speculate import get_speculate, set_speculate
 from text_generation_server.models.model import Model
 from text_generation_server.models.causal_lm import CausalLM, CausalLMBatchKeysLast
+from text_generation_server.models.transformers_flash_causal_lm import TransformersFlashCausalLM
 from text_generation_server.models.custom_modeling.opt_modeling import OPTForCausalLM
 from text_generation_server.models.custom_modeling.mpt_modeling import (
     MPTForCausalLM,
@@ -28,7 +29,7 @@ from text_generation_server.models.bloom import BloomCausalLMBatch
 from text_generation_server.models.custom_modeling.bloom_modeling import (
     BloomForCausalLM,
 )
-from text_generation_server.models.globals import ATTENTION
+from text_generation_server.models.globals import ATTENTION, USE_CUSTOM_MODELING
 from text_generation_server.models.seq2seq_lm import Seq2SeqLM
 from text_generation_server.models.galactica import GalacticaCausalLMBatch
 from text_generation_server.models.custom_modeling.neox_modeling import (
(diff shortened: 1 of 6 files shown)


server/text_generation_server/models/__init__.py
```diff
@@ -28,7 +30,7 @@
 from text_generation_server.models.custom_modeling.bloom_modeling import (
     BloomForCausalLM,
 )
-from text_generation_server.models.globals import ATTENTION
+from text_generation_server.models.globals import ATTENTION, USE_CUSTOM_MODELING
```
Narsil, line 33: Remove this global flag. We already have `FLASH_ATTENTION` which serves the same purpose.

We do not want to add flags as much a possible.

server/text_generation_server/models/__init__.py
```diff
@@ -366,12 +368,33 @@ def get_model(
...
 
+    transformers_causal_lm_class = CausalLM
+    if not USE_CUSTOM_MODELING:
+        logger.info(
+            "TGI's flash enabled models could either not be loaded or are disabled, using Transformers fallback."
+        )
+
+        transformers_model_class = getattr(transformers, modeling_auto.MODEL_FOR_CAUSAL_LM_MAPPING_NAMES[model_type])
+
+        if transformers_model_class._supports_flex_attn:
```
Narsil, line 386: This is depending on private behavior. I know we handle `transformers` but as a good citizen ideally we'd only depend on actual public behavior.

server/text_generation_server/models/__init__.py
```diff
@@ -366,12 +368,33 @@ def get_model(
...
+    global USE_CUSTOM_MODELING
 
     config_dict, _ = PretrainedConfig.get_config_dict(
         model_id, revision=revision, trust_remote_code=trust_remote_code
     )
     model_type = config_dict.get("model_type", None)
 
+    transformers_causal_lm_class = CausalLM
+    if not USE_CUSTOM_MODELING:
+        logger.info(
```
Narsil, line 380: No need to have 3 messages, that's spammy.

0 is quite fine by me (other log messages will rise in various implementations making it clear what's happening.)

server/text_generation_server/models/__init__.py
```diff
@@ -674,7 +697,7 @@ def get_model(
                 FLASH_ATT_ERROR_MESSAGE.format("Sharded Santacoder")
             )
         else:
-            return CausalLM.fallback(
+            return transformers_causal_lm_class.fallback(
```
Narsil, line 700: In general, I like to remove indirections.

Here, `transformers_causal_lm_class` is not known by the reader, he requires looking up where that's define which means following the flow of code is hard.

We *know* if models support flex attention or not. We can hardcode them `CausalLM` -> `TransformersFlashCausalLM`.

That removes the need to "guess" and the dependency on the private bit.

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,565 @@
...
+def tgi_flash_attention_forward(
+    module,
+    query_states: torch.Tensor,
+    key_states: torch.Tensor,
+    value_states: torch.Tensor,
+    attention_mask: torch.Tensor,
+    softmax_scale: Optional[float] = None,
+    sliding_window: Optional[int] = None,
+    softcap: Optional[float] = None,
+    **kwargs,
```
Narsil, line 44: No kwargs.

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,565 @@
...
+            torch_dtype=dtype,
+            device_map=("auto" if device_count > 1 else None),
+            load_in_8bit=quantize == "bitsandbytes",
+            trust_remote_code=trust_remote_code,
+            attn_implementation="tgi",
+            tp_plan="auto" if world_size > 1 else None,
+        )
+
+        if device_count == 1 and quantize != "bitsandbytes":
+            model = model.to(device)
```
Narsil, line 156: Can't we have `device_map="auto"` all the time ? Would be simpler, no ?

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,565 @@
...
+        kv_cache,
+        block_tables,
+        slots,
+        seqlen,
+        max_s,
+        prefill_cache_indices,
+        lm_head_indices,
+    ):
+        hidden_states = self.model.model.forward(
+            input_ids=input_ids[None, ...],  # expand dim to easily fit transformers
```
Narsil, line 255: `unsqueeze(0)` or real reshape.

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,565 @@
...
+        cu_seqlen_prefill,
+        kv_cache,
+        block_tables,
+        slots,
+        seqlen,
+        max_s,
+        prefill_cache_indices,
+        lm_head_indices,
+    ):
+        hidden_states = self.model.model.forward(
```
Narsil, line 254: IS that consistent enough ?? I thought some models defined `self.transformer` instead of `self.model`.

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,565 @@
...
+        if lm_head_indices is not None:
+            hidden_states = hidden_states[lm_head_indices]
+        logits = self.model.lm_head.forward(hidden_states)
+        return logits
+
+
+    def forward(
+        self, batch: FlashCausalLMBatch, adapter_data: AdapterBatchData
+    ) -> Tuple[torch.Tensor, Optional[torch.Tensor]]:
+        # NOTE: adapter_data: not supported
```
Narsil, line 279: We need a hard fail at least for every config that would otherwise be silently ignored.

speculation, and adapter data (the checks might already exist)

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,565 @@
...
+            kv_head_mapping=self.kv_head_mapping,
+        )[0].squeeze(dim=0)
+        # And compute logits from the lm_head, slicing correctly the indices
+        if lm_head_indices is not None:
+            hidden_states = hidden_states[lm_head_indices]
+        logits = self.model.lm_head.forward(hidden_states)
+        return logits
+
+
+    def forward(
```
Narsil, line 276: It seems everything downward is a copy from what was before.

What are the key differences if any ?

server/text_generation_server/models/transformers_flash_causal_lm.py
Cyrilvallez: They are needed here to easily "absorb" whatever is passed internally in Transformers and not used in tgi's attention. Made it more explicit for the arguments we do use though

server/text_generation_server/models/transformers_flash_causal_lm.py
Cyrilvallez: As of now yes - all models that are supported by our refactors are consistent with that naming. However I agree that it is quite an ugly workaround, and I'll open a PR asap in Transformers to allow logit slicing with Tensor (for now we only support `int` slicing with `num_logits_to_keep`)

server/text_generation_server/models/transformers_flash_causal_lm.py
Cyrilvallez: The calls to `self.model.forward()` are replaced by `self._model_forward()`

server/text_generation_server/models/__init__.py
Cyrilvallez: Solved by https://github.com/huggingface/transformers/pull/35757 in transformers

server/text_generation_server/models/transformers_flash_causal_lm.py
Cyrilvallez: Solved it with https://github.com/huggingface/transformers/pull/35757 in Transformers for cleaner and more robust interface

server/text_generation_server/models/__init__.py
Cyrilvallez: IMO the dynamic behavior is simpler as we will roll support for more and more models in transformers

Cyrilvallez: But can obviously be changed if this is a blocker on your side 😁

server/text_generation_server/models/transformers_flash_causal_lm.py
Narsil: If it's consistent now, then by all means let's use that. We don't care about legacy transformers versions :)

server/text_generation_server/models/transformers_flash_causal_lm.py
Narsil: Yes, then we can mark `kwargs` as `_kwargs` (just trying to explicitly say there might be arguments we do not use)

PR #2870, opened 2024-12-29
chore: Update jsonschema to 0.28.0

Stranger6667:
This PR updates `jsonschema` to `0.28.0` and makes use of its newer APIs.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -456,18 +456,18 @@ dependencies = [
 
 [[package]]
 name = "bit-set"
-version = "0.5.3"
+version = "0.8.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "0700ddab506f33b20a03b13996eccd309a48e5ff77d0d95926aa0210fb4e95f1"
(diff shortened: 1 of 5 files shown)


danieldk:
Thanks! The version bumps seem to cause issues elsewhere: https://github.com/huggingface/text-generation-inference/actions/runs/12686931224/job/35360265368?pr=2893#step:7:1626

Stranger6667:
Hm, `reqwest` does not have the `blocking` feature enabled, even though this submodule is explicitly used in `validation.rs:539`

Stranger6667:
I pushed a commit that adds `blocking` to `reqwest`, this should pass now. Sorry for not checking it earlier!

danieldk:
> Hm, `reqwest` does not have the `blocking` feature enabled, even though the `blocking` submodule enabled by this feature is explicitly used in `validation.rs:539`

Ah yes, it was implicitly enabled by `jsonschema`, but since `jsonschema` uses 0.12.x now it didn't happen anymore. Thanks for the fix!

danieldk (approved):

PR #2886, opened 2025-01-07
Give TensorRT-LLMa proper CI/CD 😍

mfuntowicz:


paulinebm:
test with Morgan

Narsil:
Can we please keep the same CI for every backend. It's much better to keep it that way so we don't maintain 3 separate CIs.

mfuntowicz:
@Narsil it's the case here no?

.github/workflows/ci_build.yaml
```diff
@@ -37,7 +37,7 @@ jobs:
       # fail-fast is true by default
       fail-fast: false
       matrix:
-        hardware: ["cuda", "cuda-trtllm", "rocm", "intel-xpu", "intel-cpu"]
+        hardware: ["cuda", "rocm", "intel-xpu", "intel-cpu"]
```
Narsil, line 40: This removes it.

.github/workflows/build_trtllm.yaml
```diff
@@ -0,0 +1,151 @@
+name: Build TensorRT-LLM
+on:
+  workflow_call:
+    inputs:
+      runs-on:
+        type: string
```
Narsil, line 6: Please make CI not read the runner in inputs. This is terribly unsafe (and not needed here).

You're directly reading this as the `runs-on` primitive. We're never doing that.

.github/workflows/ci_build.yaml
mfuntowicz: Would this make more sens moving forward? 👍🏻 👎🏻 https://github.com/huggingface/text-generation-inference/pull/2916/files

Narsil (approved):

PR #2913, opened 2025-01-15
Flash Transformers modeling backend support

Cyrilvallez:
# What does this PR do?

Supersedes https://github.com/huggingface/text-generation-inference/pull/2858 for convenience (CIs) as requested (I cannot just change the origin branch from the old PR)

### DIFF
--- a/server/text_generation_server/models/transformers_flash_causal_lm.py
+++ b/server/text_generation_server/models/transformers_flash_causal_lm.py
@@ -1,0 +1,309 @@
+import math
+import sys
+from typing import Optional, Tuple, Dict, Any
+
+import torch
+from opentelemetry import trace
+from loguru import logger
+from transformers import AutoTokenizer, AutoModelForCausalLM
+
+from text_generation_server.models.flash_causal_lm import (
+    FlashCausalLMBatch,
+    FlashCausalLM,
+)
+from text_generation_server.utils.import_utils import (
+    empty_cache,
+    synchronize,
+    get_free_memory,
+)
+from text_generation_server.adapters import AdapterBatchData
+from text_generation_server.layers.attention import paged_attention, attention, Seqlen
+from text_generation_server.layers.attention.kv_cache import KVScales
+from text_generation_server.models.globals import ATTENTION
+from text_generation_server.models.metadata_kernels import block_tables_to_ragged
+
+
+tracer = trace.get_tracer(__name__)
+
+
+def patch_everywhere(
+    attribute_name: str, patch: Any, module_name_prefix: Optional[str] = None
+):
+    """
+    Finds all occurences of `attribute_name` in the loaded modules and patches them with `patch`.
+
+    Args:
+        attribute_name (`str`):
+            The name of attribute to patch.
+        patch (`Any`):
+            The patch for the attribute.
+        module_name_prefix (`Optional[str]`, defaults to `None`):
+            If set, only module names starting with this prefix will be considered for patching.
+    """
+    # sys.modules may be updated while being iterated over, hence the list copy.
+    for name in list(sys.modules):
+        module = sys.modules[name]
+        if module_name_prefix is not None and not name.startswith(module_name_prefix):
+            continue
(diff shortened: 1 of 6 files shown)


server/text_generation_server/utils/logits_process.py
```diff
@@ -219,7 +218,7 @@ def filter(self, indices):
         return None
 
 
-class HeterogeneousTopPLogitsWarper(LogitsWarper):
+class HeterogeneousTopPLogitsWarper(LogitsProcessor):
```
Narsil, line 221: Why those changes ?

It looks like these are Warpers, not Processors.... No ?

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,563 @@
...
+        kv_cache: List[KVCache],
+        block_tables: torch.Tensor,
+        slots: torch.Tensor,
+        seqlen: Seqlen,
+        max_s: int,
+        lm_head_indices: Optional[torch.Tensor],
+    ):
+        # Transformers does not support None as a default
+        if lm_head_indices is None:
+            lm_head_indices = 0
```
Narsil, line 258: But that's wrong isn´t  it ?

`0` would mean use index `0` which is probably wrong.
How does `transformers` use that value anyway ?

server/text_generation_server/models/transformers_flash_causal_lm.py
```diff
@@ -0,0 +1,563 @@
...
+            block_tables=block_tables,
+            slots=slots,
+            seqlen=seqlen,
+            max_s=max_s,
+            kv_head_mapping=self.kv_head_mapping,
+        ).logits.squeeze(dim=0)
+        return logits
+
+
+    def forward(
```
Narsil, line 278: To save rewriting all of this, can we maybe monkey path `model.forward`.

It's not great, but it still feels better than this (which is duplicating quite complex logic that we're likely to forget to update).


in the `__init__`:  `self.model.forward = monkey_patch_forward` (that does the simple unsqueeze, squeeze).

Wdyt ?

The other option would be to introduce an indirection in `FlashCausalLM` by having an inner method that captures that `model.forward`call and we override it (I'm not a fan either to add such level of indirection, it's just as ugly as the monkey patching in my book, because it's still impossible for the reader to know that it's being modified somewhere else).

server/text_generation_server/utils/logits_process.py
Cyrilvallez: Recent Transformers version deprecated `LogitsWarper` in favor of `LogitsProcessor`. They have the same logic/interface

Cyrilvallez: `LogitsWarper` has been completely removed, thus it needs to change here

server/text_generation_server/models/transformers_flash_causal_lm.py
Cyrilvallez: We slice from `-num_logits_to_keep:`, so 0 means "keep all logits"

server/text_generation_server/models/transformers_flash_causal_lm.py
Cyrilvallez: Yes I thought about it but was not sure you guys would like it 🙃 I'll update with the monkey patch

server/text_generation_server/models/transformers_flash_causal_lm.py
Narsil: Then it's still wrong and we cannot use that.

TGI does dynamic batching, so the amount of logits to keep is just not enough information.


We can keep any amount of indices (or just the last ones, or whatever). Depending of if we need the logprobs, or however sliced the various queries are interlaced.

Maybe let's revert to doing this ourselves manually no ?

Narsil (approved):
LGTM. Thanks for this !

Narsil (approved):

PR #2949, opened 2025-01-24
Fix tool call response to adhere to OpenAI spec

Datta0:
# What does this PR do?

1. [OpenAI tool call arguments](https://arc.net/l/quote/gqmavaml) field is supposed to be string. TGI seems to output a JSON. Fix it.
2. [ToolCalls](https://arc.net/l/quote/otijcfye) in streaming case should be a list of tool calls, TGI outputs singular tool call. So update the struct accordingly.

### DIFF
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -727,7 +727,7 @@ pub(crate) struct ChatCompletionChoice {
 pub struct ToolCallDelta {
     #[schema(example = "assistant")]
     role: String,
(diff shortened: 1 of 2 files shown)


drbh:
cleaning up/closing this PR as these changes have been superseded by #3076, and #3086.

- serialization of non streaming outputs resolved in #3076
- tool calls as a vector #3086 

Thank you!

PR #2957, opened 2025-01-27
fix: Telemetry

Hugoch:
# What does this PR do?

This PR fixes unhandled errors that avoided stop telemetry events to be sent. It also adds a 15min heartbeat not to loose info for long running instances.


ErikKaum (approved):
should be good, thank you 👍

PR #2964, opened 2025-01-29
misc(gha): expose action cache url and runtime as secrets

mfuntowicz:
Avoid leaking token and cache url

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -191,6 +191,7 @@ jobs:
             PLATFORM=${{ env.PLATFORM }}
             build_type=${{ env.BUILD_TYPE }}
             sccache_gha_enabled=on
+          secrets:
             actions_cache_url=${{ env.ACTIONS_CACHE_URL }}
             actions_runtime_token=${{ env.ACTIONS_RUNTIME_TOKEN }}
 
--- a/Dockerfile_trtllm
+++ b/Dockerfile_trtllm
@@ -64,12 +64,12 @@ WORKDIR /usr/src/text-generation-inference
 ARG cuda_arch_list
 ARG build_type
 ARG sccache_gha_enabled
-ARG actions_cache_url
-ARG actions_runtime_token
 
 # Install Rust
 ENV PATH="/root/.cargo/bin:$PATH"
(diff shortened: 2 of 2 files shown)


.github/workflows/build.yaml
```diff
@@ -150,6 +150,15 @@ jobs:
           username: ${{ secrets.AZURE_DOCKER_USERNAME }}
           password: ${{ secrets.AZURE_DOCKER_PASSWORD }}
           registry: db4c2190dd824d1f950f5d1555fbadf0.azurecr.io
+      - name: configure aws credentials
+        id: aws-creds
+        uses: aws-actions/configure-aws-credentials@e3dd6a429d7300a6a4c196c26e071d42e0343502
+        with:
+          role-to-assume: ${{ secrets.AWS_ROLE_GITHUB_BUILDX_CACHE }}
```
Narsil, line 157: WE still had issues with those for release, isn't that why we remvoed it altogether ?

mfuntowicz: cc @glegendre01

glegendre01: i increased the session time, but the run tests is failing now.

Narsil: It's not about session time, it's also that the role couldn't be assumed on anything else than PRs.
Where we need to run the CI for building and pushing also on tags/main and basically any kind of branch.

Narsil: So adding this crippled the release process.

mfuntowicz:
Maybe we should split the two scopes that we are addressing here: 
- Focus this PR on making GHA argument secrets and not ENV
- Open a second PR to address the flakyness of AWS S3 stuff

wdyt @Narsil, @glegendre01 ?

.github/workflows/build.yaml
glegendre01: we can allow all branches if you need ?

glegendre01: it's currently filtering on theses : refs/heads/main + refs/tags/v*

Ryans-ui (approved):
1

PR #2982, opened 2025-01-31
Add 'json_schema' alias to GrammarType.Json

aW3st:
Adds a serde alias to `GrammarType.Json` such that `{"type": "json_schema", "value": json_schema}` deserializes correctly.

This brings the OpenAI-compatible API closer to [OpenAI's official spec](https://platform.openai.com/docs/api-reference/chat/create#chat-create-response_format) and makes the server more interoperable with other OpenAI-compatible servers.

Eventually it would be nice to be able to use `beta.chat.completions.parse` method on OpenAI's python client. 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

@OlivierDehaene OR @Narsil

### DIFF
--- a/integration-tests/models/test_grammar_response_format_llama.py
+++ b/integration-tests/models/test_grammar_response_format_llama.py
@@ -29,26 +29,55 @@ async def test_grammar_response_format_llama_json(llama_grammar, response_snapsh
         unit: str
         temperature: List[int]
 
+    json_payload={
+        "model": "tgi",
+        "messages": [
+            {
+                "role": "system",
+                "content": f"Respond to the users questions and answer them in the following format: {Weather.schema()}",
+            },
+            {
+                "role": "user",
+                "content": "What's the weather like the next 3 days in San Francisco, CA?",
+            },
+        ],
+        "seed": 42,
+        "max_tokens": 500,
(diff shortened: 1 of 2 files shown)


Narsil:
Hi @aW3st,

Thanks a lot for the PR. We're unlikely to adhere to anything `beta` from OpenAI, or even everything that OpenAI supports.
The OpenAI adaptation layer exists for simple, convenient ubiquitous tool support from client libraries (most of which have a OpenAI compatibility layer with various degrees of support).

We intend to support those very common use cases and be as easy as possible to be consumed. However, the intent of TGI is to support the many various different open source models, which sometimes operate very differently than OpenAI's. Therefore adhering to exact OpenAI spec can become quite difficult, especially for the most advanced use cases.

That being said `json_schema` should probably be supported. However I don't think json_schema is equivalent to the current `json`, we should probably upgrade `outlines` which now has support for both.

Pinging @drbh in case I said anything wrong here.

drbh:
closing as https://github.com/huggingface/text-generation-inference/pull/3046 was merge. 

Thank you @aW3st for the changes! 🙏

*note: please see https://github.com/huggingface/text-generation-inference/pull/3046#issuecomment-2729926714 for example usage

PR #3003, opened 2025-02-10
change ChatCompletionChunk to align with "OpenAI Chat Completions str…

sywangyi:
…eaming API"

When stream_options: {"include_usage": true} is included, choices is None only for the last chunk, and usage is always None except for the last chunk.

see https://cookbook.openai.com/examples/how_to_stream_completions#4-how-to-get-token-usage-data-for-streamed-chat-completion-response
@OlivierDehaene OR @Narsil

### DIFF
--- a/router/src/lib.rs
+++ b/router/src/lib.rs
@@ -790,12 +790,15 @@ impl ChatCompletionChunk {
             created,
             model,
             system_fingerprint,
-            choices: vec![ChatCompletionChoice {
-                index: 0,
-                delta,
-                logprobs,
-                finish_reason,
-            }],
+            choices: match usage {
+                None => vec![ChatCompletionChoice {
+                    index: 0,
+                    delta,
+                    logprobs,
+                    finish_reason,
+                }],
+                _ => vec![],
+            },
             usage,
         }
     }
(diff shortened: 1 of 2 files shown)


sywangyi:
details please refer discussion in https://github.com/vllm-project/vllm/pull/12394

Narsil:
What's broken otherwise ? 

Our baseline is if it breaks the OpenAI client we *try* to fix it (depending on how much it's tying to  internals it's sometimes hard to be ISO compatible), and otherwise we try to minimize the amount of bits we send (that are not useful) and therefore we don't send what's not needed.

sywangyi:
https://community.openai.com/t/usage-stats-now-available-when-using-streaming-with-the-chat-completions-api-or-completions-api/738156

Narsil:
Closing this as TGI doesn't need to output None all the time. The OpenAI client will put them back, and adding unecessary values (and there's already some that are still there) is just wasted bandwidth.

Narsil:
This has been merged in a different form in the follow-up PR: https://github.com/huggingface/text-generation-inference/pull/3007

The core idea is the same, but lots of testing code was modified to make the tests more consistent a checking for the actual behavior of this flag (plus simplifying some internal structures).

PR #3021, opened 2025-02-13
Optional content field in Chat Completions Request

saileshd1402:
Currently, TGI is validating that "content" field always exists in chat completions request. But in case of an "assistant" message type in OpenAI [messages](https://platform.openai.com/docs/api-reference/chat/create#chat-create-messages) struct, content can be empty.

According to OpenAI spec: "The contents of the assistant message. Required unless tool_calls or function_call is specified."

Example:
```
{
    "model": "llama-3-1-8b",
    "stream": false,
    "messages": [
        {
            "role": "assistant",
            "tool_calls": [
                {
                    "id": "0",
                    "function": {
                        "arguments": "{\"location\":\"Boston\",\"unit\":\"celsius\"}",
                        "name": "get_current_weather"
                    },
                    "type": "function"
                }
            ]
        }
    ]
}
```
Error returned by TGI:
```
Failed to deserialize the JSON body into the target type: messages[0]: missing field `content` at line 17 column 9
```

This PR fixes that by making content field optional and adding `tool_calls` to the `Message` struct (used in ChatRequest)

### DIFF
--- a/router/src/infer/chat_template.rs
+++ b/router/src/infer/chat_template.rs
@@ -67,7 +67,9 @@ impl ChatTemplate {
                     format!("\n---\n{}", tool_prompt)
                 };
                 if let Some(last_message) = messages.last_mut() {
-                    last_message.content.push(MessageChunk::Text { text });
+                    if let Some(content) = last_message.content.as_mut() {
+                        content.push(MessageChunk::Text { text });
+                    }
                 }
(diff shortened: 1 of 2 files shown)


Narsil:
We fixed this here: https://github.com/huggingface/text-generation-inference/pull/3036

Thanks a lot for your PR here !

PR #3022, opened 2025-02-14
Update the llamacpp backend

angt:
This PR:

 - Bumps `llama.cpp` to the latest version.
 - Makes `--model-gguf` optional (defaulting to Q4 quantization for now).
 - Enables mmap, flash attention, and KQV offloading by default, while introducing `--disable-xxx` flags for customization.


backends/llamacpp/requirements.txt
```diff
@@ -1,3 +1,4 @@
 transformers==4.48.2
 huggingface-hub==0.28.1
 hf-transfer==0.1.9
+torch==2.6.0
```
mfuntowicz, line 4: Naive Q: Do we really need `torch` package? It will certainly significantly increase the overall size of the container, so better asking 😅

backends/llamacpp/make-gguf.sh
```diff
@@ -0,0 +1,33 @@
...
+	return 1
+}
+
+GGUF="$1"
+GGUF_DIR=$(dirname "$GGUF")
+GGUF_TMP="model.src/tmp.gguf"
+MODEL_ID="$2"
+MODEL_DIR="model.src/$2"
+REV="${3-main}"
+
```
mfuntowicz, line 14:
```suggestion
HF_HUB_ENABLE_HF_TRANSFER=1
```

backends/llamacpp/src/main.rs
```diff
@@ -205,24 +206,51 @@ async fn main() -> Result<(), RouterError> {
...
+                .arg(&args.revision)
+                .spawn()?
+                .wait()
+                .await?;
+
+            if !status.success() {
+                error!("Failed to generate GGUF");
+            }
+            model_gguf
+        }
```
mfuntowicz, line 236: Do we want this to be handled by the server? Or should it be handled by an entrypoint in the container?

mfuntowicz (approved):

backends/llamacpp/requirements.txt
angt: Yep required:
```
  File "/usr/bin/convert_hf_to_gguf.py", line 22, in <module>
    import torch
ModuleNotFoundError: No module named 'torch'
gguf_init_from_file: failed to open GGUF file 'models/model.gguf'
```

backends/llamacpp/make-gguf.sh
angt: It is derived from the env of the docker

backends/llamacpp/src/main.rs
angt: I initially implemented it in an entrypoint, but I had to manage arguments and environment variables and mix it with users choices or doing something like the launcher...

Hugoch, line 236:
```suggestion
let model_gguf = if let Some(model_gguf) = args.model_gguf {
    model_gguf
} else {
    let make_gguf = std::env::var("MAKE_GGUF").map_err(|e| {
        error!("No GGUF model given and environment variable MAKE_GGUF is missing.");
        RouterError::VarError(e)
    })?;

    let model_gguf = "models/model.gguf".to_string();

    let status = Command::new(make_gguf)
        .arg(&model_gguf)
        .arg(&args.model_id)
        .arg(&args.revision)
        .spawn()?
        .wait()
        .await?;

    if !status.success() {
        error!("Failed to generate GGUF");
    }
    model_gguf
};
```

docs/source/backends/llamacpp.md
```diff
@@ -54,6 +54,10 @@ cd ~/models
 curl -LOJ "https://huggingface.co/Qwen/Qwen2.5-3B-Instruct-GGUF/resolve/main/qwen2.5-3b-instruct-q4_0.gguf?download=true"
 ```
```
Hugoch, line 56: You probably can remove the model preparation step as you now auto-prepare the model and add a `## Using a custom GGUF` section after the `## Run Docker image`? Idea is to have the least steps in the doc for a first "quick try". wdyt?

Narsil:
For the CI, you need to push on the main repo instead of a fork (permissions all that)

docs/source/backends/llamacpp.md
```diff
@@ -75,13 +64,31 @@ docker run \
...
     --n-gpu-layers 99
+    --model-id "Qwen/Qwen2.5-3B-Instruct"
+```
+
+## Using a custom GGUF
+
+GGUF files are optional as they will be automatically generated at
+startup if not already present in the `models` directory. However, if
+the default GGUF generation is not suitable for your use case, you can
+provide your own GGUF file with `--model-gguf`, for example:
```
Narsil, line 78: Is that a local file reference that users need to mount themselves ?
How does this work with file splitting : https://huggingface.co/unsloth/DeepSeek-R1-GGUF/tree/main ?
It's saying we need both, why ? 

Seems like we could potentially keep feeding a Hub link no ? `unsloth/DeepSeek-R1-GGUF/resolve/main/DeepSeek-R1-Q2_K/` (File splitting makes this a mess, and potentially various models don't have the same layout, I'm just thinking how to deploy easily).

Narsil: Also in the unspecfied gguf file case, are we saving the model in the mounted volume (so that subsequent startups are faster ? )

docs/source/backends/llamacpp.md
```diff
@@ -101,10 +108,10 @@ The table below summarizes key options:
 | `--split-mode`                      | Split the model across multiple GPUs                                   |
 | `--defrag-threshold`                | Defragment the KV cache if holes/size > threshold                      |
 | `--numa`                            | Enable NUMA optimizations                                              |
-| `--use-mmap`                        | Use memory mapping for the model                                       |
+| `--disable-mmap`                    | Disable memory mapping for the model                                   |
```
Narsil, line 111: Your call on this, but I like this kind of stuff to be automated and checked by CI.

Narsil (dismissed):

Dockerfile_llamacpp
```diff
@@ -48,15 +47,17 @@ FROM deps AS builder
...
 RUN cargo build \
-    --profile release-opt \
+    --profile release \
     --package text-generation-router-llamacpp --frozen
 
 FROM nvidia/cuda:12.8.0-cudnn-runtime-ubuntu24.04
+WORKDIR /app
 
+ENV DEBIAN_FRONTEND=noninteractive
 RUN apt update && apt install -y \
```
fgbelidji, line 61: `RUN apt update && apt upgrade -y && apt install -y \` otherwise CVE are flagged as critical in Amazon Inspector (for vulnerabilities scans)

angt: Done

docs/source/backends/llamacpp.md
angt: yes, if you omit the `--model-gguf` option, the generated gguf will be stored in `/app/models` and will be used the next time.

fgbelidji (approved):

docs/source/backends/llamacpp.md
angt: Totally agree, I will look into automating it in a future PR.

mfuntowicz (approved):

PR #3027, opened 2025-02-17
feat: Add the parsing of HF_HUB_USER_AGENT_ORIGIN environment variable for telemetry

Hugoch:
Add the parsing of HF_HUB_USER_AGENT_ORIGIN environment variable for telemetry to add info about the environment running TGI. That is useful to track usage in case of collaborations for example.

Synced with https://github.com/huggingface/huggingface_hub/pull/2869


alvarobartt (dismissed):
LGTM, thanks! You probably need to pull from `main` to avoid the `trufflehog` issue, already fixed in `main`.

Narsil (approved):

PR #3033, opened 2025-02-18
Add Neuron backend

dacorvo:
# What does this PR do?

This adds the neuron backend that was previously maintained in the [optimum-neuron](https://github.com/huggingface/optimum-neuron) repository.

This backend is built on top of the AWS Neuron SDK, and comprises:

- the legacy v2 TGI launcher and router,
- a neuron specific inference server for text-generation.

## Documentation

A dedicated [documentation page](https://moon-ci-docs.huggingface.co/docs/text-generation-inference/pr_3033/en/backends/neuron) has been added in the backends subsection.

## Tests

The backend comes with some dedicated tests:
- neuron server tests (using only the server python package),
- integration tests (using docker images).

Both set of tests require some models to be pre-exported and cached to test:
- deploying pre-exported neuron models from the hub directly,
- deploying vanilla models using cached neuron graphs.

The server tests are not run for the moment, only the integration tests.

Since these tests are very specific to neuron, they can only be activated by specifying the new `--neuron` python option.
Conversely, as soon as the `--neuron` option is set, all tests that do not have the `neuron` marker are disabled.

The neuron integraiton tests use also specific fixtures that have been added as local plugins:
- fixtures.neuron.model (takes care of preexporting the models and just fetches them if it has been done before),
- fixtures.neuron.service (custom service loop that uses the specific neuron launch env variables).

## Next steps

- use python 3.11 and align on the main Dockerfile,
- add a custom launcher that only exposes the relevant parameters and sets default values,
- add a new router for servers that have a static batch size.


dacorvo:
The neuron tests are correctly skipped by the other workflows. See for instance the CUDE integration tests:

image

Narsil (approved):
LGTM!

PR #3036, opened 2025-02-19
Improve tool call message processing

drbh:
this PR allows incoming chat requests to specify `tool_calls` and no `content`, which allows tools to be passed from a tool response call into the messages of a subsequent request

for example if the following set of messages is sent without these changes an error is thrown attempting to deserialize the message `missing field 'content'`  

```jsonc
[
    {"role": "user", "content": "What's the weather like in Paris today?"},
    {
        "content": "",
        "role": "assistant",
        "tool_calls": [
            {
                "id": "0",
                "function": {
                    "arguments": '{"longitude": 2.2945, "latitude": 48.8567}',
                    "name": "get_weather",
                    "description": None,
                },
                "type": "function",
            }
        ],
    },
    {"role": "tool", "tool_call_id": "0", "content": "6.7"},
]
```

response
```txt
I'm an AI and do not have access to real-time data. However, based on location information (Paris) I can provide general information. \n\nThe temperature in Paris varies widely throughout the year. In the summer (June to August), the average high temperature is around 23°C (73°F), while in the winter (December to February), the average low temperature is around -1°C (30°F). \n\nTo get the current weather in Paris, I recommend checking a weather website or
```

this PR allows messages without tools to be sent without error.

notes:
- tools are included in the template by serializing the calls as a JSON string and replacing the content
- thank you @saileshd1402 for a opening this draft PR https://github.com/huggingface/text-generation-inference/pull/3021 all of your changes are included in this PR!


saileshd1402:
Thank you for this change! I'll test it out and update here

saileshd1402:
I've tested this change, it is working well!

router/src/lib.rs
```diff
@@ -1221,9 +1223,15 @@ pub struct TextMessage {
 
 impl From<Message> for TextMessage {
     fn from(value: Message) -> Self {
+        let content = value
```
alvarobartt, line 1226: Maybe we could add an additional check here to prevent this type of messages from anything other than the `assistant`, as well as if both `content` and `tool_calls` are provided, WDYT? Maybe even a custom new error as you suggested recently.

router/src/server.rs
```diff
@@ -111,8 +111,9 @@ request_body = CompatGenerateRequest,
 responses(
 (status = 200, description = "Generated Text",
 content(
-("application/json" = Vec<GenerateResponse>),
-("text/event-stream" = StreamResponse),
+(Vec<GenerateResponse> = "application/json"),
+(Vec<GenerateResponse> = "application/json"),
```
alvarobartt, line 115: Duplicated here!

```suggestion
```

router/src/lib.rs
```diff
@@ -1176,15 +1176,17 @@ pub enum MessageChunk {
...
     #[schema(example = "user")]
     role: String,
     #[schema(example = "My name is David and I")]
-    pub content: MessageContent,
+    pub content: Option<MessageContent>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[schema(example = "\"David\"")]
     name: Option<String>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    tool_calls: Option<Vec<ToolCall>>,
```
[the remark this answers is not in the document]
Narsil: Untagged enum are not great for the error message so we should probably implement our own deserialize method instead.  (Since there doesn't seem to be a `type` field equivalent serde could use to know which enum to deserialize)

integration-tests/models/__snapshots__/test_tools_llama/test_flash_llama_tool_reply_response.json
```diff
@@ -0,0 +1,26 @@
...
+  "choices": [
+    {
+      "finish_reason": "stop",
+      "index": 0,
+      "logprobs": null,
+      "message": {
+        "content": "I can't access real-time data, but I can provide you with current conditions and forecast for Paris, France:\n\nThe current conditions in Paris are mostly cloudy with a temperature of 6.7°C (44.1°F). \n\nPlease note that the actual weather may differ from this information, and I recommend checking the forecast on a reliable weather website for the most up-to-date information.",
+        "name": null,
+        "role": "assistant",
+        "tool_calls": null
```
Narsil, line 11: Shouldn't we skip that?

backends/v3/Cargo.toml
```diff
@@ -16,7 +16,7 @@ path = "src/main.rs"
 [dependencies]
 async-trait = "0.1.74"
 async-stream = "0.3.5"
-axum = { version = "0.7", features = ["json"] }
+axum = { version = "0.8", features = ["json"] }
```
Narsil, line 19: Why do we need upgrades ?

integration-tests/models/test_tools_llama.py
```diff
@@ -468,3 +468,42 @@ async def test_flash_llama_grammar_tools_sea_creatures_stream_function_object(
...
+                        "function": {
+                            "arguments": '{"longitude": 2.2945, "latitude": 48.8567}',
+                            "name": "get_weather",
+                            "description": None,
+                        },
+                        "type": "function",
+                    }
+                ],
+            },
+            {"role": "tool", "tool_call_id": "0", "content": "6.7"},
```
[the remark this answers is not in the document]
Narsil: Relevant internal discussion : https://huggingface.slack.com/archives/C06JKEMK6BZ/p1739982025532149

backends/v3/Cargo.toml
drbh: only needed to generate valid openapi docs for the previous approach of `Option<MessageContent>` this is no longer needed with improved typing (using an enum instead of two optionals) in the latest commits

integration-tests/models/__snapshots__/test_tools_llama/test_flash_llama_tool_reply_response.json
drbh: name and tool_calls are actually skipped in the response, but the Python client library adds None when deserializing the response.

the actually response payload is

```jsonc
{
  "object": "chat.completion",
  "id": "",
  "created": 1740011163,
  "model": "meta-llama/Llama-3.1-8B-Instruct",
  "system_fingerprint": "3.1.1-dev0-native",
  "choices": [
    {
      "index": 0,
      "message": {
        "role": "assistant",
        "content": "I can't access real-time data, but I can provide you with current conditions and forecast for Paris, France:\n\nThe current conditions in Paris are mostly cloudy with a temperature of 6.7\u00b0C (44.1\u00b0F). \n\nPlease note that the actual weather may differ from this information, and I recommend checking the forecast on a reliable weather website for the most up-to-date information."
      },
      "logprobs": null,
      "finish_reason": "stop"
    }
  ],
  "usage": {
    "prompt_tokens": 103,
    "completion_tokens": 79,
    "total_tokens": 182
  }
}
```

integration-tests/models/test_tools_llama.py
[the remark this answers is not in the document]
drbh: updated to include `tool_call_id` in the chat template as some models will expect this template input in the latest commits

router/src/lib.rs
drbh: good point! I've updated the PR to prefer a none optional enum that ensure either a content or tools are provided - looking into how errors may be improved now

router/src/server.rs
drbh: thanks, avoided these issues/changes, by avoiding bumping the utopia version

PR #3061, opened 2025-02-26
feat: add support for HF_HUB_USER_AGENT_ORIGIN to add user-agent Origin field in Hub requests.

Hugoch:
# What does this PR do?

Adds support for HF_HUB_USER_AGENT_ORIGIN to add user-agent Origin field in Hub requests.
Useful to track collabs.

### DIFF
--- a/Dockerfile.neuron
+++ b/Dockerfile.neuron
@@ -18,7 +18,8 @@ RUN apt-get update -y \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get clean
 
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.84.0 --profile minimal -y
+COPY rust-toolchain.toml rust-toolchain.toml
+RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain none
 ENV PATH="/root/.cargo/bin:${PATH}"
 RUN cargo install cargo-chef --locked
 
--- a/Dockerfile.neuron
+++ b/Dockerfile.neuron
@@ -18,7 +18,7 @@ RUN apt-get update -y \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get clean
 
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.80.1 --profile minimal -y
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.84.0 --profile minimal -y
 ENV PATH="/root/.cargo/bin:${PATH}"
 RUN cargo install cargo-chef --locked
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1007,9 +1007,9 @@ dependencies = [
 
 [[package]]
 name = "cxx"
-version = "1.0.140"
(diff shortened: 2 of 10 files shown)


Dockerfile.neuron
```diff
@@ -18,7 +18,7 @@ RUN apt-get update -y \
     && rm -rf /var/lib/apt/lists/* \
     && apt-get clean
 
-RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.80.1 --profile minimal -y
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- --default-toolchain 1.84.0 --profile minimal -y
```
angt, line 21: You can use `rust-toolchain.toml`:
```
COPY rust-toolchain.toml rust-toolchain.toml
RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain none
```

Narsil: We can also not pipe anything into `sh` while we're at it, but there are unfortunately a lot of these occurrences.

Narsil (dismissed):
LGTM

Dockerfile.neuron
Hugoch: I'm not sure of what would be the best way to avoid the pipe. `rustup-init` checks the architecture to download the right installer which we need to do. We could copy `rustup-init` in the repo and add a check for `rust-init` signature for each architecture but needs to be updated for each update. Do you have other ideas in mind?

Narsil (approved):

Dockerfile.neuron
dacorvo: > You can use `rust-toolchain.toml`:
> 
> ```
> COPY rust-toolchain.toml rust-toolchain.toml
> RUN curl -sSf https://sh.rustup.rs | sh -s -- -y --no-modify-path --default-toolchain none
> ```

This command does not install any toolchain, which was precisely the whole point of the previous command. Am i missing something ?

Hugoch:
@dacorvo Yes, it builds on https://github.com/huggingface/text-generation-inference/actions/runs/13627605158/job/38160431550

dacorvo:
> @dacorvo Yes, it builds on https://github.com/huggingface/text-generation-inference/actions/runs/13627605158/job/38160431550 Seems that `main` now refers to 1.85.0 which fails

The issue is not related to `1.85.01: the build you are mentioning was before the review.
Here is the failing build with the changes introduced after the review: https://github.com/huggingface/text-generation-inference/actions/runs/13657446044/job/38180115060

Dockerfile.neuron
angt: Yes but the correct toolchain (from `rust-toolchain.toml`) will be installed with `cargo-chef`.
At least it works here: https://github.com/huggingface/text-generation-inference/blob/491ed9e11d23debe366a45783dd805825210bfe6/Dockerfile_llamacpp#L38-L41

dacorvo: I still don't understand how it could work: the code snippet in llamacpp is **different** than the one you suggested above.
This one installs explicitly the 1.85.0 toolchain, instead of specifying `none`, which doesn't install any toolchain.

angt: It was working, the correct toolchain was automatically installed at this step:
```
RUN cargo install cargo-chef --locked 
```
But I confirm it is not the case anymore.

PR #3065, opened 2025-02-27
Expose the real-time internal state of the batcher through SSE

mfuntowicz:


backends/v3/src/backend.rs
```diff
@@ -112,6 +136,10 @@ impl Backend for BackendV3 {
         .is_ok()
     }
 
+    fn events(&self) -> BroadcastReceiver<EngineState> {
+        self.state_events.0.subscribe()
```
co42, line 140: You should send the current state as soon as a new connection is open (new subscribe)

router/src/server.rs
```diff
@@ -1501,6 +1503,28 @@ async fn metrics(prom_handle: Extension<PrometheusHandle>) -> String {
...
+#[utoipa::path(get, tag = "Text Generation Inference", path = "/state")]
+#[instrument(skip_all)]
+async fn state(
+    Extension(infer): Extension<Infer>,
+) -> Result<Sse<impl Stream<Item = Result<Event, axum::Error>>>, StatusCode> {
+    if cfg!(feature = "engine-state") {
+        let stream = infer.events();
+        let sse =
+            Sse::new(BroadcastStream::from(stream).map(|state| {
+                Event::default().json_data(state.map_err(|err| axum::Error::new(err))?)
```
co42, line 1515: Name of the event ?

backends/v3/src/queue.rs
```diff
@@ -138,8 +147,29 @@ async fn queue_task(
...
+                    }
+
+                    // Send new state to the channel for broadcasting
+                    if let Err(err) = queue_events.send(*engine_state.read().await) {
+                        tracing::warn!(
+                            "Failed to send BatchEvent::QueueChanged({}): {err}",
+                            num_queued_tokens + entry_num_tokens
+                        )
+                    }
+                }
```
co42, line 172:
```suggestion
                engine_state.modify(|state| state.in_queue += entry_num_tokens);
```

Function modify then send the new state in the SSE

PR #3079, opened 2025-03-07
wip: comment out prepend full_text

### LINKED ISSUE #1416: Madlad400 support
### Model description

T5 based model for Machine translation.

### Open source status

- [x] The model implementation is available
- [X] The model weights are available

### Provide useful links for the implementation

https://huggingface.co/jbochi/madlad400-3b-mt#training-details

### System Info
text-generation-inference 1.3.0
python 3.10

### Information
 - [X] Docker

### Tasks
 - [X] An officially supported command

### Reproduction
The model loaded proporly but I am getting incorrect results. I want to know if the tgi 1.3.0 supports Madlad400? Thanks a lot.

### Expected behavior
Madlad400 support

jrc2139:
# What does this PR do?

This is an investigation as to why using `"return_full_text": true` as a parameter in hitting the `/generate` endpoint produces a valid translation while running madlad400

Background: https://github.com/huggingface/text-generation-inference/issues/1416#issuecomment-2705165023

Fixes #1416 

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -295,10 +295,10 @@ pub(crate) async fn generate_internal(
     );
 
     let compute_characters = req.inputs.chars().count();
-    let mut add_prompt = None;
-    if req.parameters.return_full_text.unwrap_or(false) {
-        add_prompt = Some(req.inputs.clone());
-    }
+    //let mut add_prompt = None;
+    //if req.parameters.return_full_text.unwrap_or(false) {
+    //    add_prompt = Some(req.inputs.clone());
+    //}
 
     let details: bool = req.parameters.details || req.parameters.decoder_input_details;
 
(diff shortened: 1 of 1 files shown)


Narsil:
`return_full_text` is a legacy option linked to `transformers.pipelines` initial implementation (something like 4+ years ago).
We had API dependencies on that behavior and therefore implemented here, it can mostly be disregarded nowadays.

What you did here is effectively deactivate its job. Which is really simply to decode all the decoder tokens, instead of only the new ones.

I was under the impression that T5 was a encoder/decoder model so I'm surprised to see a decoder-only model here.

I'm pretty sure the "bug" will simply end up being a tokenizer issue, where possibly you're hitting a very old T5 non flash version that doesn't support some flags used in those tokenizers.

Thanks for the fix for your use case, I'll leave it up for others to see, but if we're going to fix it, we need to find the root cause and fix that instead.

josh-at-straker:
Any movement on this one? Would be excellent to fully support MadLad as a translation T5 model.

PR #3091, opened 2025-03-10
Release of Gaudi Backend for TGI

baptistecolle:
What does this PR do?

This PR finalizes the integration of Gaudi into TGI.

Key Updates:
	•	Introduces a new Docker image for Gaudi.
	•	Adds documentation on using TGI with Gaudi.
	•	Deprecates tgi-gaudi. This is also linked to https://github.com/huggingface/tgi-gaudi/pull/287
	•	Migrates the examples section from tgi-gaudi, which includes benchmarking TGI Gaudi with continuous batching.

Sanity Check:
	•	Gaudi models have been tested with the new backend. However, due to the variety of configurations, not all models could be verified.
	•	FP8 and VLM support have also been tested.

Next Steps:
	•	Incorporate the latest changes from the tgi-gaudi fork into this backend.


docs/source/backends/gaudi.mdx
```diff
@@ -0,0 +1,303 @@
...
+| Mixtral-8x7B          | mistralai/Mixtral-8x7B-Instruct-v0.1   | ✔           | ✔          | ✔           | ✔          |
+| Mistral-7B            | mistralai/Mistral-7B-Instruct-v0.3     | ✔           | ✔          | ✔           | ✔          |
+| Falcon-180B           | tiiuae/falcon-180B-chat                |             | ✔          |             | ✔          |
+| Qwen2-72B             | Qwen/Qwen2-72B-Instruct                |             | ✔          |             | ✔          |
+| Starcoder2-3b         | bigcode/starcoder2-3b                  | ✔           | ✔          | ✔           |            |
+| Starcoder2-15b        | bigcode/starcoder2-15b                 | ✔           | ✔          | ✔           |            |
+| Starcoder             | bigcode/starcoder                      | ✔           | ✔          | ✔           | ✔          |
+| Gemma-7b              | google/gemma-7b-it                     | ✔           | ✔          | ✔           | ✔          |
+| Llava-v1.6-Mistral-7B | llava-hf/llava-v1.6-mistral-7b-hf      | ✔           | ✔          | ✔           | ✔          |
```
baptistecolle, line 66: This is copied from tgi-gaudi fork

@regisss QQ, we support actually more models than that no? As the modeling is done in optimum-habana, so for example we support phi models. Should we add a full list of the models we support? And maybe add a test suite to check all those models for TGI

regisss: I think this table only contains the models and configurations that have been tested and validated by Intel. So I guess more models would work yes, but I don't know if we want to extend this table without testing them.
WDYT @yao-matrix @kaixuanliu?

docs/source/backends/gaudi.mdx
```diff
@@ -0,0 +1,303 @@
...
+
+#### BF16 Precision (Default)
+By default, all models run with BF16 precision on Gaudi hardware.
+
+#### FP8 Precision
+TGI-Gaudi supports FP8 precision inference with [Intel Neural Compressor (INC)](https://docs.habana.ai/en/latest/PyTorch/Inference_on_PyTorch/Inference_Using_FP8.html).
+
+To run FP8 Inference:
+
+1. Measure statistics using [Optimum Habana measurement script](https://github.com/huggingface/optimum-habana/tree/main/examples/text-generation#running-with-fp8)
```
baptistecolle, line 115: @regisss Quick question: We could actually do all of this in TGI at startup no? so add a `--fp8` option to TGI and then do quantization before starting the server. We could do that by calling the quantization scripts from optimum-habana in TGI. That would be more user friendly in my opinion that having to manually do the measurement in optimum-habana and then go back to TGI. Is this correct or am I missing something?

backends/gaudi/examples/benchmark/README.md
```diff
@@ -0,0 +1,39 @@
+# TGI-Gaudi example
```
Narsil, line 1: You're welcome to add something in the main doc too (/docs)

backends/gaudi/examples/docker_commands/docker_commands.md
```diff
@@ -0,0 +1,283 @@
...
+volume=$PWD/data   # share a volume with the Docker container to avoid downloading weights every run
+
+docker run -p 8080:80 \
+   --runtime=habana \
+   --cap-add=sys_nice \
+   --ipc=host \
+   -v $volume:/data \
+    -e PREFILL_BATCH_BUCKET_SIZE=1 \
+    -e BATCH_BUCKET_SIZE=1 \
+   ghcr.io/huggingface/text-generation-inference:latest-gaudi \
```
Narsil, line 120: Please never recommend `latest`. It causes way too much confusion to downstream users. Use the actual latest version `3.1.1.` We always update every instance for releases.

Forcing the version number helps tracking down issues, usually users are just running old version and simply upgrading solves their issue.

backends/gaudi/examples/docker_commands/docker_commands.md
```diff
@@ -0,0 +1,283 @@
...
+   --runtime=habana \
+   --cap-add=sys_nice \
+   --ipc=host \
+   -v $volume:/data \
+   -e HF_TOKEN=$hf_token \
+   -e MAX_TOTAL_TOKENS=2048 \
+   -e BATCH_BUCKET_SIZE=256 \
+   -e PREFILL_BATCH_BUCKET_SIZE=4 \
+   -e PAD_SEQUENCE_TO_MULTIPLE_OF=64 \
+   ghcr.io/huggingface/text-generation-inference:latest-gaudi \
```
Narsil, line 99: No latest.

backends/gaudi/examples/docker_commands/docker_commands.md
```diff
@@ -0,0 +1,283 @@
...
+   --runtime=habana \
+   --cap-add=sys_nice \
+   --ipc=host \
+   -v $volume:/data \
+   -e HF_TOKEN=$hf_token \
+   -e MAX_TOTAL_TOKENS=2048 \
+   -e PREFILL_BATCH_BUCKET_SIZE=2 \
+   -e BATCH_BUCKET_SIZE=32 \
+   -e PAD_SEQUENCE_TO_MULTIPLE_OF=256 \
+   ghcr.io/huggingface/text-generation-inference:latest-gaudi \
```
Narsil, line 75: No latest.

backends/gaudi/examples/docker_commands/docker_commands.md
```diff
@@ -0,0 +1,283 @@
...
+   -e PAD_SEQUENCE_TO_MULTIPLE_OF=64 \
+   ghcr.io/huggingface/text-generation-inference:latest-gaudi \
+   --model-id $model \
+   --sharded true --num-shard 8 \
+   --max-input-tokens 1024 --max-total-tokens 2048 \
+   --max-batch-prefill-tokens 4096 --max-batch-size 256 \
+   --max-waiting-tokens 7 --waiting-served-ratio 1.2 --max-concurrent-requests 512
+```
+
+### Llava-v1.6-Mistral-7B on 1 card (BF16)
```
Narsil, line 107: It might be early for that, but in nvidia backend, selecting every flag automatically is a huge bonus for downstream users (so they don't have to do guesswork).

Also having a sane standard default values goes a long way ("not too horrible performance" for most deployments > "perfect but really hard to nail" on most deployments).

I don't know exactly performance profiles for this hardware, but on nvidia there's usually only a handful of settings that are really relevant

backends/gaudi/examples/benchmark/tgi_client.py
```diff
@@ -0,0 +1,88 @@
...
+
+def except_hook(args):
+    print(f"Thread failed with error: {args.exc_value}")
+    os._exit(1)
+
+
+threading.excepthook = except_hook
+
+
+class TgiClient:
```
Narsil, line 21: We already have `text_generation.Client` (deprecated), `huggingface_hub.InferenceClient` (current) and `k6` benchmark loads (really good just missing TTFT) and https://github.com/huggingface/inference-benchmarker/ which is a custom tool that does both (not perfect as any benchmark tool, but we can make PRs if desired).

It would be nice to not accumulate tools more than there already are.

I can already predict that this tool is going to fail under at least certain circumstances. Python is really bad a doing performance measurements. GIL lock will kick in, and it's overall slow enough that you can measure Python's own slowness rather than the underlying system being measured (for instance HTTP negociation is rather slow compared to lower level code).

You're spawning a thread per sample which abysmally bad. Using a ThreadPool at least or even an Async runner should help alleviate some issues.
 That won't fix python being slow though.

Narsil (approved):
LGTM.

Left a few remarks, could be for follow-ups

backends/gaudi/examples/docker_commands/docker_commands.md
baptistecolle: Yes indeed, that is something I want to do! I believe this would be much better from a user stand point. I will probably ask you also for some pointer on the best practices to do that 😅

backends/gaudi/examples/benchmark/README.md
baptistecolle: No relevant anymore as I removed the custom benchmarking scripts for Gaudi

backends/gaudi/examples/benchmark/tgi_client.py
baptistecolle: For reference, I also updated the benchmarking section to mention https://github.com/huggingface/inference-benchmarker as the preferred tool for benchmarking Gaudi hardware

docs/source/backends/gaudi.mdx
yao-matrix: @regisss I think not now, we can adopt the way that "validate one, add one" strategy. We are refactoring the HPU backend to leverage paged kv-cache(OH doesn't support, so we need change to the way that using TGI's modeling and integrate hpu op; rather than directly using OH's modeling) to align performance w/ vllm. When the PR is ready, we'll review w/ you, then we can make plan to validate/enable all TGI supported models step by step, and update README accordingly. WDYT

baptistecolle: Ok thanks for answering then we can keep it like that for the moment. I am also looking into way to add more testing to the backend so that maybe we could validate config in a more automated way. This would allow to extend the model support on TGI as those config would be validated via some integration tests directly

docs/source/backends/gaudi.mdx
baptistecolle: For reference: I discussed that internally with Regis and it is possible, so this will be left to a futur PR as a way to improve the user experience for Gaudi users on TGI

regisss (approved):
LGTM, very nice!

PR #3136, opened 2025-03-24
Improve inference success message

co42:
# What does this PR do?

Makes the message useful without needing the spans

### DIFF
--- a/router/src/server.rs
+++ b/router/src/server.rs
@@ -421,7 +421,12 @@ pub(crate) async fn generate_internal(
     }
 
     tracing::debug!("Output: {}", output_text);
-    tracing::info!("Success");
+    tracing::info!(
+        "Success. total_time={}s inference_time={}s tokens={}",
+        total_time.as_secs_f32(),
+        inference_time.as_secs_f32(),
+        response.generated_text.generated_tokens,
+    );
 
     let response = GenerateResponse {
         generated_text: output_text,


beurkinger (approved):
👍

Narsil:
This information is already in the trace (and therefore in the log message) Why do you want things in the log message itself ?

beurkinger:
@Narsil I'm not seeing the generated_text / generated_tokens fiels in the logs output on my side

co42:
@Narsil We are trying to have more concise logs in Endpoints but still have useful information, right now either we just display the message "Success" without anything else, or everything and there are 8 lines per log.

I would be nice to have a message with useful information and if you want to have more look at the extra fields

Narsil:
> I would be nice to have a message with useful information and if you want to have more look at the extra fields

Yes, but who gets to chose what merits a place in this log ? 

`Success` is mostly enough if you're just looking for errors which is a valid case imho. The core thing, is that adding metrics in the log message, which are already included elsewhere is really bad imho, those metrics are already there, and in parsable format by existing log readers. Adding them again in the message just adds noise.

The truth is that what you want to see depends on what you are doing, there is no "good log". We have no issues looking at everything we want to see in production using log readers in the current format.

I'm pushing back on this because I don't want to have endless discussion about what's a good log message or not, not this one, nor any other one, at least not based on "verbosity" level of a given tool or how pretty it looks in a certain tool. There is just no correct answer.

What should drive the discussion, is how can you identify issues in production, can you understand what's going on in the system and can you solve issues when they occur at runtime. Currently I don't think this change helps any of that (because the information is already there)

PR #3142, opened 2025-03-28
Gaudi: Add Integration Test for Gaudi Backend

baptistecolle:
# What does this PR do?

This PR includes the following changes:
- **Adds an integration test suite for the Gaudi backend:**
This suite tests various model architectures to ensure new models work correctly with TGI. It’s based on the [Optimum TPU integration tests](https://github.com/huggingface/optimum-tpu/tree/main/text-generation-inference/integration-tests).
- **Updates the documentation:**
The docs now reflect the models we currently support on Gaudi. Previously, only a limited subset was listed, even though many more are actually compatible.
- **Removes the old, broken test suite:**
The outdated test suite that never worked properly with tgi-gaudi has been deleted.


regisss (approved):
LGTM!

Is it planned to add a workflow that automatically trigger tests at some point?

baptistecolle:
> LGTM!
> 
> Is it planned to add a workflow that automatically trigger tests at some point?

Yes but only a subset as the test take a whiel to run

PR #3160, opened 2025-04-10
Gaudi: add CI

baptistecolle:
# What does this PR do?

This PR adds CI support for the Gaudi backend. It includes an integration test that starts the model "meta-llama/Llama-3.1-8B-Instruct", performs a few requests, and verifies that the outputs match the expected results.

Additional models are also supported, but running tests for all of them is quite slow, so they are not included in the CI by default. However, instructions on how to run the integration tests for all supported models have been added to the Gaudi backend README.


baptistecolle:
I’ll wait for the Gaudi integration test CI to pass before merging anything:
https://github.com/huggingface/text-generation-inference/actions/runs/14591230970/job/40927197928?pr=3160

The previous run was green, which gives me confidence in the current changes:
https://github.com/huggingface/text-generation-inference/actions/runs/14384130453/job/40336095297

Unfortunately, it can take days to get assigned a Gaudi1 runner 😭, so I figured I could start iterating on your reviews in the meantime rather than wait for the CI to finish before requesting feedback. In any case, I’ll only merge once the Gaudi integration test passes.

regisss (commented):
LGTM!

We should soon have access to Gaudi2 and Gaudi3 ephemeral runners on demand, which will makes things much easier than waiting for a DL1 instance. I suggest we wait for this to be available to update and merge this PR.

baptistecolle:
Ok, I will wait for the new runners before adding Gaudi to the CI, as indeed the DL1 runners are super unreliable

Narsil (dismissed):
LGTM

.github/workflows/build.yaml
```diff
@@ -129,9 +129,9 @@ jobs:
                 export label_extension="-gaudi"
                 export docker_volume="/mnt/cache"
                 export docker_devices=""
-                export runs_on="ubuntu-latest"
+                export runs_on="itac-bm-emr-gaudi3-dell-1gaudi"
```
regisss, line 132: All tests are going to pass with 1 device only? Big (i.e. 70B+ parameters) models are not tested?

regisss:
@baptistecolle A couple of  questions:

- It's not possible to select a specific runner for each test config right?
- If I want to add a new model to test, I just need to add a new test config in `test_gaudi_generate.py`?

.github/workflows/build.yaml
baptistecolle: Indeed, I disable big models for testing and only did a small model for faster iteration. I just activated back a mutli-card test and the test is broken 😬. There seems to be a regression between the original PR and the latest TGI backend so I am looking into it 👀

baptistecolle:
> @baptistecolle A couple of questions:
> 
> * It's not possible to select a specific runner for each test config right?
> * If I want to add a new model to test, I just need to add a new test config in `test_gaudi_generate.py`?

1. No it is not i think this would require some rework of the build workflow which is global for all the hardwares. The best alternative would be to use a runner with 8 card and then set HABANA_VISIBLE_DEVICES=1
2. Yes, that's correct

.github/workflows/build.yaml
```diff
@@ -129,9 +129,9 @@ jobs:
                 export label_extension="-gaudi"
                 export docker_volume="/mnt/cache"
                 export docker_devices=""
-                export runs_on="ubuntu-latest"
+                export runs_on="itac-bm-emr-gaudi3-dell-8gaudi"
                 export platform=""
-                export extra_pytest=""
+                export extra_pytest="--gaudi"
```
regisss, line 134: This will run the models with `"run_by_default": True` in PRs right?
If yes, I think we should change the runner above from `itac-bm-emr-gaudi3-dell-8gaudi` to `itac-bm-emr-gaudi3-dell-2gaudi` so that we test Llama 8b on a single device and on 2 devices.

integration-tests/gaudi/test_gaudi_generate.py
```diff
@@ -1,10 +1,16 @@
...
+
+def pytest_configure(config):
+    config.addinivalue_line(
+        "markers", "gaudi_all_models: mark test to run with all models"
+    )
+
+
+# The "args" values in TEST_CONFIGS are not optimized for speed but only check that the inference is working for the different models architectures.
 TEST_CONFIGS = {
     "meta-llama/Llama-3.1-8B-Instruct-shared": {
```
regisss, line 15:
```suggestion
    "meta-llama/Llama-3.1-8B-Instruct-sharded": {
```

integration-tests/gaudi/test_gaudi_generate.py
```diff
@@ -25,6 +31,7 @@
             "--max-batch-prefill-tokens",
             "2048",
         ],
+        "run_by_default": True,
```
regisss, line 34: I cannot add the comment at line 24 so adding it here. Related to my other comment, let's test this model on two devices only (so 2 shards). This allows us to use a smaller runner and should be a bit faster.

regisss (commented):
I just left a couple of comments regarding the runner size.
I'll add more tests later anyway for Llama4 and R1 on 8 devices.

Do you know if there is a nightly CI in TGI?

baptistecolle:
I updated the CI to a smaller runner so that I do testing on a 1 Gaudi card and a test on 2 Gaudi cards to test the model sharding logic

regisss (approved):
LGTM!

PR #3186, opened 2025-04-22
feat: support logit bias in chat request

drbh:
This PR adds support for the previously unused logit_bias parameter in a chat request.

Example request without logit_bias using `Qwen/Qwen2-VL-2B-Instruct`
```bash
curl http://localhost:3000/v1/chat/completions -X POST \
-d '{
    "model": "tgi",
    "seed": 42,
    "max_tokens": 10,
    "messages": [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "say Hello"
                }
            ]
        }
    ]
}'
```

response
```json
{
    "object": "chat.completion",
    "id": "",
    "created": 1745338432,
    "model": "Qwen/Qwen2-VL-2B-Instruct",
    "system_fingerprint": "3.2.3-dev0-native",
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "Hello! How can I help you today?"
            },
            "logprobs": null,
            "finish_reason": "length"
        }
    ],
    "usage": {
        "prompt_tokens": 21,
        "completion_tokens": 10,
        "total_tokens": 31
    }
}
```

with logit_bias specified (specifically `Hello` with a large negative bias to avoid generating it

```bash
curl http://localhost:3000/v1/chat/completions -X POST \
-d '{
    "model": "tgi",
    "seed": 42,
    "max_tokens": 10,
    "logit_bias": {
        "9707": -100
    },
    "messages": [
        {
            "role": "user",
            "content": [
                {
                    "type": "text",
                    "text": "say Hello"
                }
            ]
        }
    ]
}'
```

it returns a different response; which happens to be in nice greeting in Spanish

```json
{
    "object": "chat.completion",
    "id": "",
    "created": 1745338592,
    "model": "Qwen/Qwen2-VL-2B-Instruct",
    "system_fingerprint": "3.2.3-dev0-native",
    "choices": [
        {
            "index": 0,
            "message": {
                "role": "assistant",
                "content": "¡Hola! ¿Cómo puedo ayudarte?"
            },
            "logprobs": null,
            "finish_reason": "length"
        }
    ],
    "usage": {
        "prompt_tokens": 21,
        "completion_tokens": 10,
        "total_tokens": 31
    }
}
```

> [!IMPORTANT]  
> This PR contains breaking changes as the `logit_bias` type has changed from a list to a map

### DIFF
--- a/server/text_generation_server/models/vlm_causal_lm.py
+++ b/server/text_generation_server/models/vlm_causal_lm.py
@@ -418,7 +418,7 @@ class VlmCausalLM(FlashCausalLM):
             logger.warning(
                 f"Tokenizer vocab size {self.tokenizer.vocab_size} does not match model vocab size {self.config.vocab_size}. Updating tokenizer vocab size."
             )
-            # TODO: HUGE HACK! This is a workaround to update the vocab size 
+            # TODO: HUGE HACK! This is a workaround to update the vocab size
             # in the tokenizer. When the tokenizer is updated within the model
             # the vocab size is not updated in the tokenizer.
             self.tokenizer._vocab_size = self.config.vocab_size


drbh:
pinging @hanouticelina regarding changes to the hub library https://github.com/huggingface/huggingface_hub/pull/2724 once this is merged

server/text_generation_server/utils/tokens.py
```diff
@@ -248,9 +260,14 @@ def __init__(
...
         grammar_types: List[int],
         fsm_grammar_states=List[int],
+        logit_biases: List[Optional[dict]] = None,
     ):
         warpers = []
 
+        # Initialize with empty logit biases if none provided
+        if logit_biases is None:
+            logit_biases = [None] * len(do_sample)
+
```
Narsil, line 270: Can we do like other arguments, and just send everything initialized instead of `None`

server/text_generation_server/utils/tokens.py
```diff
@@ -57,7 +60,13 @@ def __init__(
             if grammar != ""
             else None
         )
+        self.logit_bias_processor = (
+            LogitBiasProcessor(logit_bias, tokenizer, device)
+            if logit_bias is not None and len(logit_bias) > 0
+            else None
+        )
         self.tokenizer = tokenizer
+        self.logit_bias = logit_bias
```
Narsil, line 69: This is not necessary, once we have the processor we should let go of the other object (Think of this as logit_bias taking ownership)

server/text_generation_server/utils/logits_process.py
```diff
@@ -623,3 +623,112 @@ def filter(self, indices):
...
+                    token_id = tokens[0] if tokens else -1  # Use -1 for not found
+
+                self.token_id_mapping[token_str] = token_id
+
+            token_id = self.token_id_mapping[token_str]
+
+            # Apply bias if token ID is valid
+            if 0 <= token_id < scores.size(-1):
+                scores[:, token_id] += bias_value
+
```
Narsil, line 664: This is too slow implementation, the logit_bias must be a tensor precalculated, that we just need to add to the scores.

server/text_generation_server/utils/logits_process.py
```diff
@@ -623,3 +623,112 @@ def filter(self, indices):
...
+        self.tokenizer = tokenizer
+        self.logit_biases = logit_biases or {}
+
+        # Pre-compute token IDs for each token string
+        self.token_id_mapping = {}
+
+    def __call__(self, input_ids: torch.Tensor, scores: torch.Tensor) -> torch.Tensor:
+        # If no logit biases, return scores unchanged
+        if not self.logit_biases:
+            return scores
```
Narsil, line 643: We should never be in that case since we're only adding the processor when it's not empty, I'd happily switch to an assert here.

server/text_generation_server/utils/logits_process.py
```diff
@@ -623,3 +623,112 @@ def filter(self, indices):
...
+
+
+class LogitBiasProcessor:
+    """Process logits with logit biases."""
+
+    def __init__(
+        self, logit_biases: Optional[dict], tokenizer: PreTrainedTokenizerBase
+    ):
+        self.tokenizer = tokenizer
+        self.logit_biases = logit_biases or {}
```
Narsil, line 635:
```suggestion
        self.logit_biases = logit_biases
```

server/text_generation_server/utils/logits_process.py
```diff
@@ -623,3 +623,112 @@ def filter(self, indices):
...
+
+        # Pre-compute token IDs for each token string
+        self.token_id_mapping = {}
```
Narsil, line 638: Where is the pre-computing ?

server/text_generation_server/utils/logits_process.py
```diff
@@ -623,3 +623,112 @@ def filter(self, indices):
...
+
+                    self.token_id_mapping[token_str] = token_id
+
+                token_id = self.token_id_mapping[token_str]
+
+                # Apply bias if token ID is valid
+                if 0 <= token_id < scores.size(-1):
+                    scores[i, token_id] += bias_value
+
+        return scores
```
Narsil, line 727: Same no forloop, single tensor addition should be doable.

server/text_generation_server/utils/tokens.py
```diff
@@ -125,6 +136,7 @@ def from_pb(
             tokenizer=tokenizer,
             grammar=pb.grammar,
             grammar_type=pb.grammar_type,
+            logit_bias=dict(pb.logit_bias) if pb.logit_bias else None,
```
Narsil, line 139: Why is `pb.logit_bias` not possible ? It would maintain consistency better.

server/text_generation_server/utils/tokens.py
```diff
@@ -500,6 +530,9 @@ def from_pb(
             fsm_grammar_states=(
                 fsm_grammar_states if fsm_grammar_states else [0] * len(pb)
             ),
+            logit_biases=[
+                dict(pb_.logit_bias) if pb_.logit_bias else None for pb_ in pb
```
Narsil, line 534: Same here.

server/text_generation_server/utils/logits_process.py
```diff
@@ -623,3 +623,112 @@ def filter(self, indices):
...
+                if token_str not in self.token_id_mapping:
+                    if token_str.isdigit():
+                        # If the token string is already a numeric ID
+                        token_id = int(token_str)
+                    else:
+                        # Otherwise, use the tokenizer to get the ID
+                        tokens = self.tokenizer.encode(
+                            token_str, add_special_tokens=False
+                        )
+                        token_id = tokens[0] if tokens else -1  # Use -1 for not found
```
Narsil, line 717: We should do the sanitation much earlier, way up in the rust code, we also have the tokenizer there, and we can reject requests that contain invalid logit_bias early without having to encode or fail here.

server/text_generation_server/utils/tokens.py
drbh: yep that makes sense, updated in latest commit. Thanks!

server/text_generation_server/utils/tokens.py
drbh: oo yea removed 🙏

server/text_generation_server/utils/logits_process.py
drbh: updated along with a much better `LogitBiasProcessor`. The bias tensor is now created in `init` and simply added via `add_` in `__call__`. Thanks 🙏

server/text_generation_server/utils/logits_process.py
drbh: good point, updated along with logit bias processor changes.

server/text_generation_server/utils/logits_process.py
drbh: updated along with logit bias processor changes

server/text_generation_server/utils/logits_process.py
drbh: updated along with logit bias processor changes

server/text_generation_server/utils/logits_process.py
drbh: updated along with logit bias processor changes

server/text_generation_server/utils/tokens.py
drbh: oh good catch, simplified in the latest commit

server/text_generation_server/utils/logits_process.py
drbh: great point, I've removed the extra checking logic from the python side and reject in `validate` if any values are invalid (token id not in vocab range)

server/text_generation_server/utils/logits_process.py
```diff
@@ -623,3 +623,94 @@ def filter(self, indices):
...
+            # Create a tensor and directly copy bias values at the corresponding indices
+            self.bias_matrix[i].index_put_((token_ids,), bias_values, accumulate=True)
+
+    def __call__(self, input_ids: torch.Tensor, scores: torch.Tensor) -> torch.Tensor:
+        # Apply bias matrix as a broadcasted addition
+        if self.bias_matrix.shape[1] != scores.shape[1]:
+            # Pad the bias matrix to match the scores if it's smaller
+            self.bias_matrix = torch.nn.functional.pad(
+                self.bias_matrix, (0, scores.shape[1] - self.bias_matrix.shape[1])
+            )
```
Narsil, line 705: Why do we need the padding, I'm surprised here.

It seems to me that self.bias_matrix is (BS, VOCAB), while scores is (SEQ_LENTHS, VOCAB).

In the most common scenario (decode), it's easy, as BS == SEQ_LENGTHS.

But in prefill, and mixed prefill + decode, by using pad, you're effectively spilling the bias_matrix onto other users, no ?
It seems to me we're needing cu_seqlengths (or whatever we currently have) in order to expand (probably using https://pytorch.org/docs/stable/generated/torch.repeat_interleave.html#torch.repeat_interleave ).

Again ideally we do this at init time, not at call time (so the __call__ function is literally just an add).

Narsil (commented):
Looks much better, I think I found a flaw in the actual computation, but other than that it looks great.

router/src/validation.rs
```diff
@@ -387,6 +402,35 @@ impl Validation {
...
+                }
+
+                // Transform into the required format
+                Some(
+                    bias.iter()
+                        .map(|(k, v)| (k.parse::<u32>().unwrap(), *v as f32))
+                        .collect(),
+                )
+            }
+            _ => None,
```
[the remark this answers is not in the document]
drbh: great points and totally agree that a map is cleaner here. Updated in the latest changes for a more simple solution with map

router/src/validation.rs
```diff
@@ -387,6 +402,35 @@ impl Validation {
...
+                    }
+                }
+
+                // Transform into the required format
+                Some(
+                    bias.iter()
+                        .map(|(k, v)| (k.parse::<u32>().unwrap(), *v as f32))
+                        .collect(),
+                )
+            }
```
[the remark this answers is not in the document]
drbh: The intention is to accept token ids as string and not the token value. 

I believe that the early changes included some token encoding decoding that is now remove which is an improvement.

Additionally I've updated the logic to be more simple and avoid the unwrap/extra loop in the latest changes

server/text_generation_server/utils/logits_process.py
drbh: I was running into a strange error that appears to be a bug with the configuration of Qwen2 2B (used in the test) where the vocab size returned on `Qwen2TokenizerFast` is not the correct size (151643 instead of 151936) https://huggingface.co/Qwen/Qwen2-VL-2B-Instruct/blob/main/config.json. 

The correct vocab size is set downstream within the custom modeling code but this is not accessible by the logit_processor so I've added a hacky patch that sets `_vocab_size` on the tokenizer if the vocab_size post loading does not match the tokenizer.vocab_size.

This solution as it feels hacky and obtuse yet reliably resolves the issue.. Any ideas on a cleaner approach? 

Aside from this I've removed the padding step and now the forward is simply an `add_`

PR #3197, opened 2025-04-27
Set `uv` UV_PYTHON_INSTALL_DIR explicitly

### LINKED ISSUE #3082: Running container rootless does not work anymore
### System Info

Any environment, as long as you the TGI container with user != 0.

### Information

- [x] Docker
- [ ] The CLI directly

### Tasks

- [x] An officially supported command
- [ ] My own modifications

## Symptom
I cannot run the current TGI image with user different than root

```
docker run --platform linux/amd64 -it --user 1000 ghcr.io/huggingface/text-generation-inference:3.1.1
```
fails immediately with the following error message
```
unable to refresh ld cache, not a big deal in most cases
/usr/local/bin/text-generation-launcher: error while loading shared libraries: libpython3.11.so.1.0: cannot open shared object file: No such file or directory
```

## Error analysis
- The python library which cannot be found is installed in `/root/.local/share/uv/python/cpython-3.11.11-linux-x86_64-gnu/lib/`
- The directory `/root/` is not readable for non-root users.
- I suspect that the problem has been introduced with #2919 because release v3.0.1 still works.

### Expected behavior

I expect to be able to run TGI with non-root users, e.g. on Kubernetes. This was possible previously.

sebastianliebscher:
# What does this PR do?

In recent commits, `uv` was introduced. When installing Python via `uv python install` it [defaults to `~/.local/share/uv/python`](https://docs.astral.sh/uv/reference/cli/#uv-python-install--install-dir). Since the user context in the base stage of the Dockerfile is root, the default path resolves to `/root/.local/share/uv/python`. Unfortunately, this directory is not accessible to non-root users in unprivileged containers, e.g.

```bash
docker run --rm --user 1000 -it ghcr.io/huggingface/text-generation-inference
```

This PR fixes this by explicitly setting the `UV_PYTHON_INSTALL_DIR` variable to `/usr/share/uv` which is accessible for non-root users.

Fixes #3082

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -165,8 +165,7 @@ RUN apt-get update && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-ins
         git \
         && rm -rf /var/lib/apt/lists/*
 
-# RUN curl -LsSf https://astral.sh/uv/install.sh | sh
-# ENV PATH="$PATH:/root/.local/bin"
+ENV UV_PYTHON_INSTALL_DIR="/usr/share/uv"
 COPY --from=ghcr.io/astral-sh/uv:0.5.31 /uv /uvx /bin/
 # Install flash-attention dependencies
 # RUN pip install einops --no-cache-dir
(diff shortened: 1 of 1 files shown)


sebastianliebscher:
@Narsil any chance this will get a review? Let me know if there's anything I'm missing.

PR #3204, opened 2025-05-01
Skip `{% generation %}` and `{% endgeneration %}` template handling

alvarobartt:
# What does this PR do?

This PR is just a tentative fix for both `{% generation %}` and `{% endgeneration %}` custom syntax introduced within some chat templates for [returning a mask of the assistant generated tokens](https://github.com/huggingface/transformers/blob/7a3e208892c06a5e278144eaf38c8599a42f53e7/src/transformers/processing_utils.py#L382-L385) which is only useful during training and should indeed be ignored during inference, affecting models as e.g. [`microsoft/Phi-4-reasoning-plus`](https://huggingface.co/microsoft/Phi-4-reasoning-plus).

Thanks to @Rocketknight1 for the explanation!

## Before submitting
- [X] Did you read the [contributor guideline](https://github.com/huggingface/transformers/blob/main/CONTRIBUTING.md#start-contributing-pull-requests), Pull Request section?
- [ ] Was this discussed/approved via a GitHub issue or the [forum](https://discuss.huggingface.co/)? Please add a link to it if that's the case.
- [X] Did you make sure to update the documentation with your changes? Here are the [documentation guidelines](https://github.com/huggingface/transformers/tree/main/docs), and [here are tips on formatting docstrings](https://github.com/huggingface/transformers/tree/main/docs#writing-source-documentation).

## Who can review?

@Narsil

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 target
 router/tokenizer.json
 *__pycache__*
-.DS_Store
 
 backends/v2/src/client/pb
 backends/v3/src/client/pb
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,7 @@
 target
 router/tokenizer.json
 *__pycache__*
+.DS_Store
 
 backends/v2/src/client/pb
 backends/v3/src/client/pb
--- a/router/src/infer/chat_template.rs
+++ b/router/src/infer/chat_template.rs
@@ -47,9 +47,8 @@ impl ChatTemplate {
         let mutated_template = mutated_template.replace("[::-1]", "|reverse");
         // TODO: replace with a better solution
         // Hack to remove the {% generation %} and {% endgeneration %} statements from
-        // Phi4 Reasoning chat templates, as those are required when generating a mask
-        // for the assistant generated tokens, not natively handled yet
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -2,6 +2,7 @@
 target
 router/tokenizer.json
 *__pycache__*
+.DS_Store
```
Narsil, line 5: Can you remove this ? This belong is your own global gitignore.

Narsil (dismissed):
LGTM

Rocketknight1:
This will work fine in a lot of cases, but will fail to catch syntax like `{%- generation %}`! Because the tags can also include whitespace-stripping syntax like `{-`, it can be dangerous to just delete them instead of adding a null Jinja handler for `generate`.

alvarobartt:
> This will work fine in a lot of cases, but will fail to catch syntax like `{%- generation %}`! Because the tags can also include whitespace-stripping syntax like `{-`, it can be dangerous to just delete them instead of adding a null Jinja handler for `generation`.

Thanks! I'll make sure to update it to handle that properly rather than a raw replacement 👍

PR #3230, opened 2025-05-16
fix: count gpu uuids if NVIDIA_VISIBLE_DEVICES env set to all

drbh:
This PR simply counts the number of gpu uuids in the case that the `NVIDIA_VISIBLE_DEVICES` is set to `"all"`


Narsil (approved):
LGTM.

PR #3256, opened 2025-06-05
Perf opt

sywangyi:
performance optimization
we test llama 70b, 8b, non-TP, TP2, TP8
TPOT improvement geomean 7%
TTFT  improvement geomean 7%


sywangyi:
@yuanwu2017 @regisss please help review

regisss:
Cc @Narsil

PR #3258, opened 2025-06-06
fix multi-modality apply template issue

sywangyi:
# What does this PR do?

Fixes # (issue)

## Before submitting
      Pull Request section?
      to it if that's the case.

## Who can review?


Narsil:
This is most likely a wrong fix. We cannot just skip using templates on some models because the answer is not to our liking.

We need to fix the template instead, no ? But overriding like this is most likely purely wrong.

sywangyi:
VLM template input like,
conversation = [
    {

      "role": "user",
      "content": [
          {"type": "text", "text": "What is shown in this image?"},
          {"type": "image"},
        ],
    },
]

 which is different with llm. 
 {"role": "user", "content": "Who are you?"},

so we may need reconstruct the msg for vlm before apply chat template. also, there is model that does not have chat template. we need to handle it as well.

PR #3264, opened 2025-06-13
[gaudi] HuggingFaceM4/idefics2-8b issue fix

sywangyi:
batch.prefill_cache_indices is reset in generate_token instead of forward, so that position_id could be updated correctly

@regisss please help review and merge it, thanks


regisss (approved):
LGTM

PR #3287, opened 2025-07-07
Deepseek v2 mla

sywangyi:
add ep to unquantized moe, which will benefit qwen3 moe bf16 and deepseek v2 bf16, greatly reduce model loading time
Qwen/Qwen3-30B-A3B load weight 925s->50s
also add mla to deepseek_v2.


sywangyi:
@regisss @yuanwu2017 please help review and merge it. thanks

regisss (approved):
LGTM

PR #3346, opened 2025-12-18
 fix(num_devices): fix num_shard/num device auto compute when NVIDIA_VISIBLE_DEVICES == "all" or "void"

oOraph:
If NVIDIA_VISIBLE_DEVICES == "all" then auto computed num shards was always 1
NVIDIA_VISIBLE_DEVICES == "void" is a new value coming from nvidia container runtime > 1.18


drbh (dismissed):
LGTM

julien-c:
nice fix

